PR #81, opened 2025-11-09
prevents flickering with Ink's Static to render only the last line dynamic

DenizOkcu:
## Description

This fixes the flickering for streaming LLM responses

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

### Automated Tests

### Manual Testing

- [ ] Tested with Ollama
- [ ] Tested with OpenRouter

## Checklist

- [x] Self-review completed

### DIFF
--- a/source/app.tsx
+++ b/source/app.tsx
@@ -1,4 +1,4 @@
-import {Box, Text, useApp} from 'ink';
+import {Box, Text, useApp, Static} from 'ink';
 import WelcomeMessage from '@/components/welcome-message';
 import React from 'react';
 import {getThemeColors} from '@/config/themes';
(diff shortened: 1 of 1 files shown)


source/app.tsx
```diff
@@ -313,17 +313,33 @@ export default function App() {
...
+										<Box marginBottom={1}>
+											<Text color={themeContextValue.colors.primary} bold>
+												{appState.currentModel}:
+											</Text>
+										</Box>
+										{/* Use Static for completed lines to prevent re-renders */}
+										{completedLines.length > 0 && (
+											<Static items={completedLines}>
+												{(line, index) => (
+													<Text key={`streaming-line-${index}`}>{line}</Text>
```
spinualexandru, line 334: Using index in the key is generally a non-issue like here if it's a static link. If we change this to a dynamic list/filterable would be good to use something like useId() from react;

PR #292, opened 2026-01-08
[Feature] Support for project-level MCP configuration files (.mcp.json)

Avtrkrb:
# [Feature] Support for project-level MCP configuration files (.mcp.json)

## Description

This PR implements hierarchical configuration loading for MCP (Model Context Protocol) servers, enabling team collaboration through project-level configuration files. The implementation adds support for Claude Code's object-based MCP server format while maintaining full backward compatibility with existing configurations. Additionally, this PR adds CLI options for version and help information.

### Key Features Added:

1. **Project-Level Configuration Support**:
   - Added support for `.mcp.json`, `mcp.json`, `.nanocoder/mcp.json`, `.claude/mcp.json`, and `.nanocoder/mcp.local.json` files
   - Project-level configs take precedence over global configs
   - Enables team collaboration by allowing MCP server configs to be committed to version control

2. **Claude Code Format Compatibility**:
   - Added support for Claude Code's object-based format: `{ "serverName": { ...serverConfig } }`
   - Maintains full compatibility with traditional array format: `[ {...serverConfig} ]`
   - Both formats work in all supported configuration locations

3. **Hierarchical Configuration Loading**:
   - Configuration files are loaded with proper precedence (project overrides global)
   - Security validation for sensitive data in project-level configs
   - Environment variable substitution works across all configuration levels

4. **Enhanced MCP Command Display**:
   - `/mcp` command now shows configuration source levels (e.g., [project], [global])
   - Users can see where each MCP server configuration originates

5. **Security Enhancements**:
   - Validation to detect hardcoded credentials in project-level config files
   - Automatic `.gitignore` generation for sensitive config files
   - Warnings for security-sensitive configurations

6. **CLI Version and Help Options**:
   - Added `--version`/`-v` flag to display version information
   - Added `--help`/`-h` flag to show usage information and available options
   - Added comprehensive documentation for CLI options in README.md
   - Added integration tests for CLI version/help functionality

7. **New Quit Command and UI Improvements**:
   - Added `/quit` command as an alternative to `/exit` for exiting the application
   - Updated `/help` command display to show "Help" instead of "/help" as the title
   - Updated `/mcp` command display to show "Model Context Protocol Servers" instead of "/mcp" as the title
   - Enhanced `/mcp` command UI with improved color scheme using theme's tool color
   - Added graceful exit handling with goodbye message for the quit command

## Type of Change

- [x] New feature
- [x] Bug fix (global providers not loaded when project config exists without providers)
[...]

PR #300, opened 2026-01-11
fix: correct log directory paths and enable file logging in production

will-lamerton:
## Summary
- Update log directory paths to follow OS conventions:
  - macOS: `~/Library/Logs/nanocoder` (was `~/Library/Preferences/nanocoder/logs`)
  - Linux: `~/.local/state/nanocoder/logs` (respects `XDG_STATE_HOME`)
  - Windows: `%LOCALAPPDATA%/nanocoder/logs`
- Enable file logging at 'info' level in production (was 'silent')
- Keep fallback logger silent to avoid console spam during async load
- Consolidate log level defaults to reduce duplication
- Fix pre-existing test bugs (wrong env var restoration)

## Test plan
- [x] Run `pnpm run test:ava source/utils/logging/*.spec.ts` - all 350 tests pass
- [x] Build succeeds
- [x] Verify logs appear in correct OS-specific directory
- [x] Verify no console spam in production mode

### DIFF
--- a/source/utils/logging/config.spec.ts
+++ b/source/utils/logging/config.spec.ts
@@ -39,7 +39,23 @@ test('getDefaultLogDirectory returns valid directory path', t => {
 
 	t.is(typeof directory, 'string', 'Should return string');
 	t.true(directory.length > 0, 'Should not be empty');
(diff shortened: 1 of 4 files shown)


will-lamerton:
To merge after #292

PR #301, opened 2026-01-11
Add /nanocoder-shape command for branding font selection

will-lamerton:
## Summary

- Add new `/nanocoder-shape` command that allows users to choose different font styles for the Nanocoder welcome banner
- Interactive selector with real-time preview as you navigate through options
- 12 font styles available: tiny (default), block, simple, simpleBlock, slick, grid, pallet, shade, 3d, simple3d, chrome, huge
- Responsive preview: shows "NC" on narrow terminals, "Nanocoder" on wide terminals
- Preference saved and persists across sessions
- Direct argument support: `/nanocoder-shape block`

## Test plan

- [x] Run `/nanocoder-shape` and verify interactive selector appears
- [x] Navigate through options and verify real-time preview updates
- [x] Select a shape and verify it is applied to welcome message
- [x] Restart Nanocoder and verify selected shape persists
- [x] Test `/nanocoder-shape <name>` direct argument
- [x] Test on narrow terminal to verify "NC" preview
- [x] All 700+ tests pass including new tests

### DIFF
--- a/source/services/bash-executor.spec.ts
+++ b/source/services/bash-executor.spec.ts
@@ -3,10 +3,28 @@ import {BashExecutor} from './bash-executor';
 
 console.log(`\nbash-executor.spec.ts`);
 
+// Track executors for cleanup
+const executorsToCleanup: BashExecutor[] = [];
+
 // Helper to create a fresh executor for each test
 function createExecutor(): BashExecutor {
-	return new BashExecutor();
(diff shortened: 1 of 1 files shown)


Avtrkrb (dismissed):
This is so cool ! We could enhance this with gradient style colors that could go hand in hand with the current theme. Probably worth exploring in a future PR.

PR #320, opened 2026-01-21
feat: add /settings command for interactive command menu

### LINKED ISSUE #316: [Feature] general /settings command
use /settings command to combine all commands so people dont have to remember all commands
after /settings is typed:

use direction keys to select one command to run

akramcodez:
### Description
Implements `/settings` command that displays an interactive menu of all available commands, eliminating the need to memorize individual command names.

Closes #316

### Changes
- Added `/settings` command to command registry
- Created `SettingsSelector` component with dynamic command list from `commandRegistry`
- Integrated settings mode into app state and mode handlers
- Added ESC key support for cancellation
- Updated README documentation

### Implementation Details
- Dynamically fetches all commands from `commandRegistry.getAll()`
- Maintains consistent theming and UI patterns
- Executes selected command via existing command flow

### Testing
- [x] `/settings` opens interactive menu
- [x] Arrow keys navigate commands
- [x] Enter executes selected command
- [x] ESC cancels and returns to chat
- [x] All commands work when launched from menu

### Files Changed
- `source/app/components/settings-selector.tsx` (new)
- `source/app/App.tsx`
- `source/app/components/modal-selectors.tsx`
- `source/app/utils/app-util.ts`
- `source/hooks/useAppState.tsx`
- `source/hooks/useModeHandlers.tsx`
- `source/hooks/useAppHandlers.tsx`
- `source/types/app.ts`
- `README.md`


Avtrkrb:
> I wonder if the `/settings` command should be used to reduce the need for other commands that can be grouped under this one. For example, `/theme`, `/title-shape`, `/nanocoder-shape`, `/setup-providers` and `/setup-mcp`. Instead of having seperate commands for all these configuration options, you can access these options under a more generic `/settings` command with an interactive menu that triggers each still command as it does now still.
> 
> I personally think the above would make more sense.

I second this, but differ only in keeping `setup-providers` & `/setup-mcp` commands dedicated & separate as they have multiple sequential steps. For the UI related commands like `/theme`, `/title-shape`, `/nanocoder-shape`, it would make a lot of sense to club them under `/settings`. 

I vouch for having dedicated commands for providers & mcp because what if there is a new mcp or provider that I want to configure in the future ? These commands are multi-step in nature & I would rather prefer to directly configure the provider/mcp rather than having to go through the entire sequence of all of these commands.

Whereas for a UI change like the theme or the title shape, I would do this occasionally or when I get bored of the current shape & would logically want to change other UI related preferences as well.

Just my thoughts...

will-lamerton:
Hey @akramcodez - can you go for this?

> I second this, but differ only in keeping setup-providers & /setup-mcp commands dedicated & separate as they have multiple sequential steps. For the UI related commands like /theme, /title-shape, /nanocoder-shape, it would make a lot of sense to club them under /settings.

> I vouch for having dedicated commands for providers & mcp because what if there is a new mcp or provider that I want to configure in the future ? These commands are multi-step in nature & I would rather prefer to directly configure the provider/mcp rather than having to go through the entire sequence of all of these commands.

> Whereas for a UI change like the theme or the title shape, I would do this occasionally or when I get bored of the current shape & would logically want to change other UI related preferences as well.

> Just my thoughts...

So the /settings command just tidies up some core commands that are not necessarily needed? So theme, title shape, nanocoder branding. These arguably don't need their own commands and can just fall under a global settings. Happy to take your thoughts too?

PR #347, opened 2026-02-08
fix: remove tool call deduplication to prevent "Tool result is missing" errors

cleyesode:
## Summary

This fix addresses the critical bug where duplicate tool calls were being silently removed during parsing, causing a mismatch between tool calls and their results. This resulted in the error message: *"Tool result is missing for tool call [ID]"*, which would cause the agent's turn to end and prevent users from prompting the agent to continue.

## Root Cause

The bug occurred due to a mismatch between the AI SDK's internal step tracking and the tool call parser:

1. **AI SDK's Behavior**: When `generateText()` executes a tool, it creates a step with:
   - `toolCalls`: Array of tool calls requested by the model
   - `toolResults`: Array of results from tool execution
   - **Crucial**: These arrays must have the same length for each step

2. **Parser's Behavior**: The JSON parser was removing duplicate tool calls:
   ```typescript
   function deduplicateToolCalls(toolCalls: ToolCall[]): ToolCall[] {
     const seen = new Set<string>();
     const unique: ToolCall[] = [];
     for (const call of toolCalls) {
       const hash = `${call.function.name}:${JSON.stringify(call.function.arguments)}`;
       if (!seen.has(hash)) {
         seen.add(hash);
         unique.push(call);
       }
       // Else: duplicate, skip it  ← THIS WAS THE BUG
     }
     return unique;
   }
   ```

3. **The Mismatch**:
   - Model requests tool `read_file` with path `/test.txt` twice
   - AI SDK creates step with 2 tool calls and 2 results
   - Parser removes one duplicate → returns 1 tool call
   - AI SDK expects results for 2 calls, finds only 1
   - Result: "Tool result is missing" error

## Changes

1. **Removed `deduplicateToolCalls()` function** from `source/tool-calling/json-parser.ts`
2. **Updated test expectations** to preserve duplicate tool calls instead of deduplicating
3. **Consolidated regex patterns** in `source/tool-calling/json-parser.ts` to prevent duplicate matches:
   - **Before**: 3 overlapping patterns
     ```typescript
     const toolCallPatterns = [
       /\{"name":\s*"([^"]+)",\s*"arguments":\s*(\{[^}]*\})\}/g,
       /\{"name":\s*"([^"]+)",\s*"arguments":\s*(\{[^}]+\})\}/g,
       /\{"name":\s*"([^"]+)",\s*"arguments":\s*"([^"]+)"\}/g,
     ];
     ```
   - **After**: 1 comprehensive pattern
     ```typescript
     const toolCallPatterns = [
       /\{"name":\s*"([^"]+)",\s*"arguments":\s*\{[\s\S]*?\}\}/g,
     ];
     ```
   - **Reason**: The original patterns would match the same inline JSON twice, causing duplicate tool calls even without the deduplication function. This was discovered during testing when the inline JSON test started failing after removing deduplication.

## Testing

All tests pass successfully:
- ✅ **json-parser.spec.ts**: 35/35 tests passed
- ✅ **tool-parser.spec.ts**: 20/20 tests passed
- ✅ **xml-parser.spec.ts**: 15/15 tests passed (unchanged)

## Impact

- **Fixes Critical Bug**: Eliminates "Tool result is missing" errors
- **Improves Reliability**: Agents can now handle repeated operations safely
- **Better LLM Support**: Works with strict providers like Anthropic
- **Aligns with Changelog**: Matches fix mentioned in v1.22.1

This aligns with the v1.22.1 changelog entry about removing aggressive deduplication and completes that fix.

## References

- Changelog Entry: v1.22.1 - "Removed aggressive tool call deduplication"
- Related Code: `source/ai-sdk-client/chat/chat-handler.ts` (AI SDK integration)
- Test Files: `source/tool-calling/*.spec.ts` (comprehensive test coverage)

### DIFF
--- a/source/tool-calling/json-parser.spec.ts
+++ b/source/tool-calling/json-parser.spec.ts
@@ -204,6 +204,7 @@ test('parseJSONToolCalls: parses inline JSON tool calls', t => {
 
 	t.is(calls.length, 1);
 	t.is(calls[0].function.name, 'read_file');
+	t.deepEqual(calls[0].function.arguments, {path: '/test.txt'});
 });
 
 test('parseJSONToolCalls: handles nested objects in arguments', t => {
--- a/source/tool-calling/json-parser.ts
+++ b/source/tool-calling/json-parser.ts
@@ -126,7 +126,7 @@ export function parseJSONToolCalls(content: string): ToolCall[] {
 
 	// Look for embedded tool calls using regex patterns
 	const toolCallPatterns = [
-		/\{"name":\s*"([^"]+)",\s*"arguments":\s*\{[\s\S]*?\}\}/g,
+		/\{"name":\s*"([^"]+)",\s*"arguments":\s*(\{[\s\S]*?\})\}/g,
 	];
 
 	for (const pattern of toolCallPatterns) {
@@ -197,9 +197,7 @@ export function cleanJSONToolCalls(
 	// Remove JSON blocks that were parsed as tool calls (for non-code-block cases)
 	const toolCallPatterns = [
 		/\{\s*\n\s*"name":\s*"([^"]+)",\s*\n\s*"arguments":\s*\{[\s\S]*?\}\s*\n\s*\}/g, // Multiline JSON blocks
-		/\{"name":\s*"([^"]+)",\s*"arguments":\s*(\{[^}]*\})\}/g,
-		/\{"name":\s*"([^"]+)",\s*"arguments":\s*(\{[^}]+\})\}/g,
-		/\{"name":\s*"([^"]+)",\s*"arguments":\s*"([^"]+)"\}/g,
+		/\{"name":\s*"([^"]+)",\s*"arguments":\s*(\{[\s\S]*?\})\}/g, // Consolidated inline pattern
 	];
 
 	for (const pattern of toolCallPatterns) {


cleyesode:
Happy to contribute. Thanks for the excellent project - a truly local coder is long overdue!

Thanks for the thorough review! You're absolutely right - the capture group issue would have caused silent failures for inline tool calls. I've made the following changes:

## Fixes Applied

### 1. Added Arguments Capture Group to Regex Pattern
Changed from:
```typescript
/\{"name":\s*"([^"]+)",\s*"arguments":\s*\{[\s\S]*?\}\}/g
```

To:
```typescript
/\{"name":\s*"([^"]+)",\s*"arguments":\s*(\{[\s\S]*?\})\}/g
```

This ensures both `name` and `argsStr` are captured, fixing the silent failure for inline tool calls.

### 2. Updated `cleanJSONToolCalls` to Use Consolidated Pattern
Replaced the old 3 overlapping patterns with the single consolidated pattern to maintain consistency between parsing and cleaning.

### 3. Enhanced Test Coverage
Added assertion to verify arguments are correctly parsed:
```typescript
t.deepEqual(calls[0].function.arguments, {path: '/test.txt'});
```

This would have caught the capture group issue during testing.

## Testing
All 70 tests pass successfully. The fix ensures:
- Inline tool calls now have correct arguments
- Multiline blocks continue to work (already using `JSON.parse`)
- `cleanJSONToolCalls` is now consistent with parsing logic

Thanks again for catching this - it's a great reminder to always verify capture groups match the expected destructuring pattern!

PR #355, opened 2026-02-10
test: add comprehensive edge case coverage for JSON tool call parsing

cleyesode:
Closes #348. Re-opening with correct edge case commits from the working PR.


cleyesode:
Re-opening with correct edge case commits (without parallel tool execution)

PR #369, opened 2026-02-18
feat: add centralized graceful shutdown system

### LINKED ISSUE #239: feat: Implement graceful shutdown system
## Use Case

Selecting "No, exit" on the security warning banner causes the process to hang indefinitely instead of exiting. This reveals a broader architectural gap: nanocoder lacks a centralized graceful shutdown system.

**Steps to Reproduce:**
1. Run `pnpm build && pnpm start`
2. When the security warning appears, select "No, exit"
3. Observe the process hangs (cursor disappears, no prompt return)
4. Verify with `ps aux | grep "node dist/cli"` - process is still running

### 1. ink's `exit()` doesn't terminate the process

The `handleExit` function in `App.tsx` calls `exit()` from ink's `useApp()` hook, which only unmounts the React component tree. If there are active event listeners or open handles, the Node.js event loop stays alive.

### 2. Logger signal handlers don't exit

In `source/utils/logging/index.ts`, SIGTERM/SIGINT handlers flush logs but never call `process.exit()`.

### 3. No centralized shutdown coordinator

Each service has its own cleanup method, but there's no orchestration:

| Resource | Location | Cleanup Method | Called on Exit? |
|----------|----------|----------------|-----------------|
| Logger | `loggerProvider` | `flush()` → `end()` | No |
| LSP Manager | `lspManagerInstance` | `shutdown()` | No |
| MCP Client | via `ToolManager` | `disconnectMCP()` | No |
| VSCode Server | `serverInstance` | `stop()` | No |
| Health Monitor | `HealthMonitor` | `stop()` | No |

### Create a ShutdownManager

```
source/utils/shutdown/
├── index.ts              # Main export
├── shutdown-manager.ts   # Coordinator singleton
└── types.ts              # Interfaces
```

```typescript
interface CleanupHandler {
    name: string;
    priority: number;  // Lower = runs first
    cleanup: () => Promise<void>;
}

class ShutdownManager {
    private handlers: CleanupHandler[] = [];
    private isShuttingDown = false;

    register(handler: CleanupHandler): void;
    unregister(name: string): void;
    async gracefulShutdown(exitCode?: number): Promise<never>;
}
```

### Cleanup Order (by priority)

1. **Priority 10**: VSCode Server - stop accepting new connections
2. **Priority 20**: MCP Client - disconnect from MCP servers
3. **Priority 30**: LSP Manager - stop language servers
4. **Priority 40**: Health Monitor - clear intervals
5. **Priority 50**: Logger - flush pending logs and close streams
6. **Priority 100**: Call `process.exit()`

### Integration Points

Services register their cleanup on init, signal handlers are centralized in ShutdownManager, and `App.tsx` uses `shutdownManager.gracefulShutdown()` for exit.

## Alternatives Considered

1. **Just add `process.exit(0)` everywhere** - Quick fix but doesn't properly clean up resources
2. **Use ink's render instance with `waitUntilExit()`** - Doesn't solve the cleanup coordination problem

## Tasks

- [ ] Create `source/utils/shutdown/` module with ShutdownManager
- [ ] Register logger cleanup with ShutdownManager
- [ ] Register LSP Manager cleanup with ShutdownManager
- [ ] Register MCP Client cleanup with ShutdownManager
- [ ] Register VSCode Server cleanup with ShutdownManager
- [ ] Register Health Monitor cleanup with ShutdownManager
- [ ] Update `App.tsx` to use ShutdownManager for exit
- [ ] Add tests for graceful shutdown
- [ ] Update `/exit` command to use ShutdownManager

## Additional Context

- **Immediate workaround**: Add `process.exit(0)` after `exit()` in `App.tsx` for the security banner (acceptable since no services need cleanup at that stage)
- **Related**: Duplicate signal handler registration - logging module's handlers fire twice, suggesting module double-loading (separate investigation needed)

## Environment

- OS: macOS (Darwin 25.2.0)
- Node.js: v24.3.0
- nanocoder: v1.19.2

will-lamerton:
## Summary

- Adds a `ShutdownManager` singleton that coordinates priority-ordered cleanup of all services on exit (VSCode server, MCP client, LSP manager, health monitor, logger)
- Replaces scattered `process.exit()` calls with `gracefulShutdown()` across all exit paths (`/exit`, `/quit`, Ctrl+C, signals, security disclaimer, non-interactive mode)
- Includes 5-second timeout safety net, re-entrant guard, and AVA test coverage

## Test plan

- [x] `pnpm run test:all` passes (format, lint, types, AVA tests, knip)
- [x] Manual: Ctrl+C exits cleanly without orphaned child processes
- [x] Manual: `/exit` command renders goodbye message and exits
- [x] Manual: Security disclaimer "No, exit" exits without hanging
- [x] Manual: With MCP servers configured, Ctrl+C terminates MCP child processes (`ps aux`)

Closes #239

### DIFF
--- a/source/components/question-prompt.tsx
+++ b/source/components/question-prompt.tsx
@@ -1,7 +1,7 @@
 import {Box, Text, useInput} from 'ink';
 import SelectInput from 'ink-select-input';
-import TextInput from 'ink-text-input';
 import {useRef, useState} from 'react';
+import TextInput from '@/components/text-input';
 import {useTerminalWidth} from '@/hooks/useTerminalWidth';
 import {useTheme} from '@/hooks/useTheme';
 import type {PendingQuestion} from '@/utils/question-queue';


source/utils/shutdown/shutdown-manager.ts
```diff
@@ -0,0 +1,122 @@
+import {loggerProvider} from '@/utils/logging/logger-provider';
+import type {ShutdownHandler, ShutdownManagerOptions} from './types';
+
+const DEFAULT_TIMEOUT_MS = 5000;
```
Avtrkrb, line 4: @will-lamerton Just a thought, would we need to increase this in the scenario where a user has multiple lsp, mcp servers loaded & waits for all of this to be gracefully shutdown before quitting ?

I have a gitlab mcp server that is sometimes problematic, hence the concern. 5000 is good, but how about we allow users to customize this with a `NANOCODER_DEFAULT_SHUTDOWN_TIMEOUT` environment variable or something similar ?

Not a "review" per say, just discussing...

will-lamerton: A good idea! I added :)

PR #371, opened 2026-02-20
feat: v2 type-safe tool system overhaul with defensive parsing

cleyesode:
Implements a three-tiered defense system for handling chaotic LLM outputs, preventing crashes from non-string responses and enabling robust self-correction.

**Closes #362**

### 1. Universal Type Safety (type-helpers.ts)
- Consolidated type conversion with \`ensureString()\` function
- Eliminated duplicate/confusing functions (removed \`toRequiredString\`, unused \`clone\`)
- Intelligent handling of unknown types: null/undefined → empty string, objects → JSON.stringify, primitives → String()
- Preserves object types in memory (ToolCall.arguments stays as Record<string, unknown>)
- Only converts to string for parsing/display operations

### 2. Response Normalization (response-formatter.ts)
- **Lamerton Fix**: Runtime trust via \`JSON.parse()\` validation instead of aggressive regex
- Validates valid JSON objects (like package.json) as not malformed
- Only flags truly broken patterns (missing name/arguments, null arguments, etc.)
- Comprehensive metadata building for debugging and strategy selection

### 3. Confidence System Inversion
- Shifted from "syntax confidence" to "actionability confidence"
- HIGH: Actual tool calls extracted (execute)
- MEDIUM: Valid JSON/XML but no tool calls (conversational text)
- LOW: Plain text (conversational text)
- MALFORMED: Broken tool patterns (self-correction)

### 4. Hall of Mirrors Defense (tool-parser.ts, json-parser.ts)
- **Line-start anchoring**: \`(?:^|\\n)\` prevents inline mentions from triggering detection
- **Optimistic reordering**: Check valid tool calls before malformed detection
- Prevents infinite loops from quoted error messages

### 5. Ghost Echo Deduplication (conversation-loop.tsx)
- Deduplicates native AI SDK tool calls with content-parsed calls
- Uses \`JSON.stringify(arguments)\` equality checks
- Prevents double-execution from API bleed in open-weights models

### 6. AI SDK Contract Fix (conversation-loop.tsx)
- Generates dummy "aborted" results for valid tools when skipping to fix bad tools
- Satisfies strict 1:1 tool call/result mapping even during self-correction
- Prevents "Tool result is missing" fatal errors

### 7. Tool Processor Modernization (tool-processor.ts)
- Converted to async function with strict \`unknown\` type
- Routes through \`normalizeLLMResponse\` middleware
- XML/Legacy models now get same type-safety guarantees as JSON models

### 8. Tool Refinement (write-file.tsx)
- Schema is \`type: 'string'\` (clear, doesn't confuse weaker models)
- Removed length validation block (runtime safety via \`ensureString()\`)
- Explicitly allows empty strings for intentional file creation (.gitkeep)

## Testing

- Complete test suite rewrite for new confidence system
- Added anchoring and optimistic reordering tests
- Updated all tool-processor tests to use async/await
- Verified malformed detection with runtime trust

## Fixes from PR #362

✅ Validator allows empty strings (intentional file creation)
✅ Schema is \`type: 'string'\` (doesn't advertise all types)

## Impact

Transforms Nanocoder from crash-prone to production-ready:
- Handles any LLM output type without crashes
- Accepts valid JSON objects as conversational text
- Prevents infinite self-correction loops
- Eliminates double-execution from API bleed
- Enables successful self-correction without SDK errors
- Local LLM experience is now stable and production-ready

## Files Modified

- source/hooks/chat-handler/conversation/conversation-loop.tsx
- source/utils/response-formatter.ts
- source/utils/response-formatter.spec.ts
- source/utils/type-helpers.ts
- source/utils/type-helpers.spec.ts
- source/tool-calling/json-parser.ts
- source/tool-calling/xml-parser.ts
- source/tool-calling/tool-parser.ts
- source/ai-sdk-client/chat/tool-processor.ts
- source/ai-sdk-client/chat/tool-processor.spec.ts
- source/ai-sdk-client/chat/chat-handler.ts
- source/tools/write-file.tsx

### DIFF
--- a/source/tools/file-ops/write-file.spec.tsx
+++ b/source/tools/file-ops/write-file.spec.tsx
@@ -860,10 +860,12 @@ test('write_file formatter: handles object content (regression test)', async t =
 	const output = lastFrame();
 
 	t.truthy(output);
-	t.regex(output!, /write_file/);
-	t.regex(output!, /manifest\.json/);
+	// Strip ANSI color codes before regex matching to avoid failures due to syntax highlighting
+	const plainOutput = stripAnsi(output!);
+	t.regex(plainOutput, /write_file/);
+	t.regex(plainOutput, /manifest\.json/);
 	// ensureString converts object to JSON string, so we expect the keys to appear
-	t.regex(output!, /"test_id":1/);
+	t.regex(plainOutput, /"test_id":1/);
 });
 
 // ============================================================================


will-lamerton:
Hey @cleyesode!

Nice work on this :)

Just a few things left before this is good to merge I think:

- `source/ai-sdk-client/chat/tool-processor.spec.ts.md` is still in the tree - looks like a scratch file that snuck in. Just `git rm` it
- `_allToolCalls` in conversation-loop.tsx is assigned but never read (leftover from before the de-dup logic). Needs removing I think.
- `_contentStr` in write-file.tsx - same deal, assigned and unused.

**One small logic fix**

In conversation-loop.tsx, `normalizeLLMResponse(fullContent)` internally runs both parsers to extract tool calls, but then `parseToolCalls(normalized.content)` runs them again right after - and `normalized.toolCalls` is never actually used. So every LLM response gets double-parsed. Easiest fix is probably to just drop the `normalizeLLMResponse` call since you're only using `normalized.content` (which for strings is just `.trim()`), or alternatively use `normalized.toolCalls` directly and drop the second `parseToolCalls()` call.

That's it! Once those are in this is ready to go I think 🚀

cleyesode:
I've applied the requested updates - also updates for a couple items identified in a final field testing sweep:

- write-file.tsx 'newContent.split is not a function' fatal error, ensureString safety wrapper now applied.

- mcp-client.ts  'Expected string, received object' non-fatal error (reattempt working), but in-practice the model tends to bypass mcp via bash command on reattempt. fix:if the tool wants a string but receives an object, ensureString is applied. mcp now runs without need for reattempt.

By the way, how exactly are we defining native tools? By technicality it appears to me the system treats anything in index as native, but in literature I see we generally exclude bash commands from the definition of native tool. Are we also excluding mcp from this definition? I'm probably just misinterpreting, thanks for clarifying.

Alas, I see I've failed testing. I'll resolve new errors when I return home.

PR #383, opened 2026-02-25
fix: safe memory operations to prevent process polyfill crash

cleyesode:
Fixes a runtime crash when `process.memoryUsage()` is called in environments where `process` is polyfilled to null.

### Problem
The application crashed with:
```text
ERROR  Cannot read properties of null (reading 'memoryUsage')
dist/utils/logging/performance.js:89:30
```

This happened because:
* Code called `process.memoryUsage()` and `process.cpuUsage()` without runtime checks
* In certain environments, `process` is polyfilled to null, causing direct calls to fail
* The crash occurred during startup in `startMetrics()` → `addTypedMessage()` → `logError()` → `loadPreferences()` → `useAppState()`

### Solution
Implemented defensive helper functions with runtime checks:
* `getSafeMemory()` - Safe memory usage getter with fallback values
* `getSafeCpuUsage()` - Safe CPU usage getter with fallback values

### Files Modified
* `source/utils/logging/performance.ts`
* `source/utils/logging/request-tracker.ts`
* `source/utils/logging/health-monitor/checks/memory-check.ts`
* `source/utils/logging/health-monitor/core/health-monitor.ts`

### Testing
Added tests in `performance.spec.ts` and `request-tracker.spec.ts` to verify:
* Fallback behavior when `process.memoryUsage()` throws
* Correct fallback values are returned

### DIFF
--- a/source/ai-sdk-client/chat/chat-handler.ts
+++ b/source/ai-sdk-client/chat/chat-handler.ts
@@ -24,6 +24,7 @@ import {
 	startMetrics,
 	withNewCorrelationContext,
 } from '@/utils/logging';
+import {getSafeMemory} from '@/utils/logging/safe-process.js';
 import {convertToModelMessages} from '../converters/message-converter.js';
 import {
 	convertAISDKToolCalls,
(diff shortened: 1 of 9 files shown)


cleyesode:
Fair take - will do.

PR #454, opened 2026-04-13
fix(nix): copy themes.json to fix nix run crash

p0ryae:
## Description

Add `install -D source/config/themes.json $out/lib/nanocoder/source/config/themes.json` to the `installPhase` in the Nix package derivation. The `source/config/themes.json` file was not being copied into the Nix store which caused `nix run` to crash immediately. I could run nanocoder by building it locally, so I figured it was probably a nix flake issue, which it was. the `source/config/themes.json` wasn't being copied and it isn't bundled by the tsc so we need to manually copy it.

## Type of Change

- [x] Bug fix

### Automated Tests

### Manual Testing

- [ ] Tested with Ollama
- [ ] Tested with OpenRouter

## Checklist
- [x] Self-review completed

### DIFF
--- a/nix/packages/default/default.nix
+++ b/nix/packages/default/default.nix
@@ -55,9 +55,8 @@ stdenv.mkDerivation (finalAttrs: {
         cp package.json $out/lib/nanocoder/
         cp -r plugins $out/lib/nanocoder/
 
-        # Copy static assets not bundled by tsc
-        mkdir -p $out/lib/nanocoder/source/config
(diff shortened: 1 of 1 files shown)


will-lamerton:
A great catch. Thank you for this.

p0ryae:
@will-lamerton just as an FYI: I noticed you added an extra copy for `prompts`, You should probably use `install -D` instead of `cp`, since `install -D` does a mkdir and cp together. I think the tests failed because the dir didn't exist. Hope it helps!

PR #493, opened 2026-05-01
fix: parse JSON tool fallbacks for local models

### LINKED ISSUE #489: it run good but
not support create file but read search

zerone0x:
## Description

Fixes #489 by teaching the XML fallback path to also accept structured JSON tool calls from local models like Ollama.

When native tool calling gets disabled after a provider error, some local models still emit JSON such as `tool_calls`, fenced `{ "name": ..., "arguments": ... }` blocks, or `create_file`-style aliases instead of the documented XML format. Before this change, those responses were treated as plain assistant text, so no tool execution happened.

This PR adds a JSON fallback parser that:
- extracts tool calls from `tool_calls` arrays and single JSON call objects
- strips fenced JSON tool blocks from assistant content once parsed
- normalizes common local-model aliases like `create_file` + `file_path`/`contents` to Nanocoder's `write_file` shape
- updates regression tests for both `parseToolCalls()` and `normalizeLLMResponse()`

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

### Automated Tests

### Manual Testing

- [ ] Tested with Ollama
- [ ] Tested with OpenRouter

## Checklist

- [x] Self-review completed

### DIFF
--- a/source/tool-calling/tool-parser.spec.ts
+++ b/source/tool-calling/tool-parser.spec.ts
@@ -141,6 +141,81 @@ test('parseToolCalls: preserves identical XML tool calls (no deduplication)', t 
 	}
 });
 
+test('parseToolCalls: parses JSON tool calls from tool_calls arrays', t => {
+	const content = JSON.stringify({
+		tool_calls: [
+			{
+				function: {
+					name: 'read_file',
+					arguments: {path: '/tmp/test.txt'},
+				},
+			},
+		],
+	});
+
+	const result = parseToolCalls(content);
+
+	t.true(result.success);
+	if (result.success) {
+		t.is(result.toolCalls.length, 1);
+		t.is(result.toolCalls[0].function.name, 'read_file');
+		t.deepEqual(result.toolCalls[0].function.arguments, {
+			path: '/tmp/test.txt',
+		});
+		t.is(result.cleanedContent, '');
+	}
+});
+
+test('parseToolCalls: parses JSON tool calls from fenced code blocks', t => {
+	const content = `I'll create that now.\n\n\
+\`\`\`json
+{
+  "name": "write_file",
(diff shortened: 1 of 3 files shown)


will-lamerton:
Hey, thanks so much for this contribution - and thanks more broadly for the work you've been putting into nanocoder lately. Your previous PRs have all been merged and were great, so this is in no way a reflection on the quality of your contributions overall.

I've spent some time digging into this one against issue #489 and I'd like to walk through where I've landed.

_On issue #489 itself:_

Looking at the attached transcript, the model never emits a tool call in any format - it just declines in plain prose. That's the only thing we can say with confidence from what's been provided. The actual cause could be any number of things: Ollama context window settings, model choice and capability, missing or incorrect configuration on the user's side, mode interactions, something else entirely. We don't have enough information to know, and a JSON parser wouldn't change the outcome shown in the transcript regardless, because there's no JSON for it to parse.

_On the PR itself_

The implementation is thoughtful and I can see the intent - but without a concrete reproduction showing a local model emitting JSON tool calls that we're failing to parse, we'd be adding ~200 lines that handle seven different JSON shapes plus tool-name and argument-key aliasing on speculation. A couple of the branches also worry me a little - the `{tool: "..."}` match in particular would treat any object with a string `tool` field as a tool call, which has false-positive risk on prose containing things like `{tool: "hammer", purpose: "demo"}`.

_Where I've landed_

I'm going to close this PR and issue #489 for now, but please don't read that as a closed door. If you do reproduce a real case where a local model emits JSON that we're not handling, I'd be genuinely glad to revisit - open a fresh issue with the raw output and we can scope something narrow from there.

Thank you again for the contribution and for understanding. Looking forward to your next one.

PR #513, opened 2026-05-11
fix: normalise pnpm v11 index.db for deterministic Nix builds

### LINKED ISSUE #511: Nix flake auto-update disabled until nixpkgs ships pnpm 11
## Summary

The Nix package auto-update workflow (`.github/workflows/update-nix.yml`) is disabled as of v1.26.1 and will stay disabled until [nixpkgs PR #505103 (`pnpm_11: init at 11.0.4`)](https://github.com/NixOS/nixpkgs/pull/505103) merges into `nixpkgs-unstable`.

## Background

In v1.26.0 we migrated the project to pnpm 11 (Node 22, new lockfile format, SQLite-backed v11 store) to track the current pnpm release. The npm and Homebrew distribution paths handled this fine. **The Nix path did not** — and we spent several hours discovering that nixpkgs's pnpm tooling simply hasn't caught up to pnpm 11 yet.

Concretely, the failure cascade was:

1. **nixpkgs ships pnpm 10.33.2 by default.** No `pnpm_11` attribute exists in the current `nixpkgs-unstable` `pkgs.pnpm` set. Verified by reading `pkgs/development/tools/pnpm/default.nix` directly.
2. **pnpm 10 can't read pnpm 11 lockfiles** (`ERR_PNPM_LOCKFILE_CONFIG_MISMATCH` on `patchedDependencies` / `overrides` shape changes).
3. Hand-overriding to pnpm 11 in `flake.nix` (`pkgs.pnpm.override { version = "11.0.9"; ... }`) surfaces a moving target of pnpm-11-specific incompatibilities in nixpkgs's `fetchPnpmDeps` setup hook:
    - pnpm 11's `bin/pnpm.cjs` (Corepack-compat shim) ships without execute bit; the nixpkgs builder symlinks `$out/bin/pnpm → pnpm.cjs`. Workaround: `chmod +x` in postInstall.
    - pnpm 11 rejects `pnpm config set manage-package-manager-versions false` against the global config (only allowed in `pnpm-workspace.yaml` now). nixpkgs's installPhase runs that command unconditionally. Workaround: `builtins.replaceStrings` patch.
    - nixpkgs's fixupPhase hardcodes cleanup of `$storePath/{v3,v10}/{tmp,projects}` — the `v11/` directory pnpm 11 actually writes to is left alone, so stale symlinks pointing at `/build/source/<random>` and timestamped tmp files leak into the output.
    - Even after cleaning `v11/tmp` and `v11/projects`, **pnpm 11's `v11/index.db` (a SQLite database new in pnpm 11) remains non-deterministic** — page layout depends on insertion order and the file has a header change counter that bumps on every write. We saw three distinct `pnpmDeps` hashes across three identical CI runs.

In short: every "fix" was reverse-engineering one piece of what nixpkgs PR #505103 needs to ship properly. The PR has been open and actively iterated since 2026-03-30 — six weeks at the time of writing — and is the canonical solution.

## What's currently in the repo

- `flake.nix` retains the pnpm 11 override scaffolding from our investigation, with comments explaining each workaround. **It does not produce a reproducible build** (the SQLite `index.db` issue is unsolved).
- `.github/workflows/update-nix.yml` is no-op'd: `workflow_dispatch` fails loudly with an explanatory message, `workflow_call` (invoked from `release.yml`) succeeds-with-warning so releases don't fail.
- Existing Nix users who installed at v1.26.0 or earlier are unaffected — those cached store paths still work.

## When to re-enable

Watch [nixpkgs PR #505103](https://github.com/NixOS/nixpkgs/pull/505103) and [nixpkgs issue #514444](https://github.com/NixOS/nixpkgs/issues/514444). Once the PR merges and `pnpm_11` lands in `nixpkgs-unstable`:

1. Replace the override scaffolding in `flake.nix` with a one-line `pkgs.pnpm_11` reference (or just `pkgs.pnpm` if it becomes the default).
2. Remove all our `overrideAttrs` patches, `preFixup` cleanup, etc. — they should all be obsoleted upstream.
3. Restore the original `update-nix.yml` body (preserved as a comment block at the bottom of the file).
4. Trigger the workflow once with the current version to refresh `flake.nix` hashes and verify a clean build, then close this issue.

## References

- nixpkgs PR #505103: https://github.com/NixOS/nixpkgs/pull/505103
- nixpkgs issue #514444: https://github.com/NixOS/nixpkgs/issues/514444
- pnpm 11 release notes (v11 store / SQLite changes): https://pnpm.io/blog/releases/11.0

ragini-pandey:
## Summary

pnpm 11 introduced a SQLite-backed store at `$storePath/v11/index.db`. Two
properties of that file make `fetchPnpmDeps` non-reproducible, so the
fixed-output derivation hash changes on every build:

1. The B-tree page layout depends on the order rows are inserted (which pnpm
   parallelises).
2. Each row's msgpackr-encoded value carries a `checkedAt` timestamp.

On top of that, nixpkgs' built-in `fetchPnpmDeps` fixup only cleans the
`v3/` and `v10/` store layouts, so pnpm 11's `v11/tmp/` (timestamped temp
files) and `v11/projects/` (symlinks into `/build/source/<random>`) leak
into the output.

Three identical CI runs produced three distinct `pnpmDeps` hashes before
this change.

## Fix

A `preFixup` hook on `fetchPnpmDeps` in `flake.nix`:

1. `rm -rf $storePath/v11/{tmp,projects}` — extends nixpkgs' cleanup to the
   v11 layout.
2. Loads pnpm's own `@pnpm/store.index` module (bundled at
   `${pnpm}/libexec/pnpm/node_modules/`) via `NODE_PATH` and normalises
   `v11/index.db` in-place:
   - rewrites every entry's msgpack value with `checkedAt: 0`
   - deletes all rows and re-inserts them in sorted-key order via
     `setRawMany`, so the SQLite B-tree page layout is stable
3. Restores `.github/workflows/update-nix.yml` to its original working
   form (the `workflow_dispatch` / `workflow_call` no-op blocks are
   removed) now that the underlying non-determinism is addressed.

This mirrors the `pnpm-fixup-state-db` helper introduced in
[nixpkgs PR #505103](https://github.com/NixOS/nixpkgs/pull/505103),
inlined here until that PR merges into `nixpkgs-unstable`. When it does,
the `preFixup` block and the surrounding `pnpm.override` / `overrideAttrs`
scaffolding can be ripped out in favour of a one-line `pkgs.pnpm_11`
reference (per the rollback steps in #511).

Closes: #511

## Type of change

- [x] Bug fix

### Automated

- `pnpm test:all` passes.
- Existing CI green (9/10 checks; the 10th is the Nix workflow this PR
  itself re-enables).

### Manual (Nix reproducibility — the actual acceptance test)

This needs to be run by someone with a working Nix install; I do not
have Nix locally, so I have not been able to verify reproducibility
empirically. Suggested procedure:

```bash
git fetch origin pull/513/head:pr-513 && git checkout pr-513

nix build --no-link .#default.pnpmDeps
nix path-info --json .#default.pnpmDeps | jq -r '.[].narHash' > /tmp/h1
nix store delete --ignore-liveness $(nix path-info .#default.pnpmDeps)

nix build --no-link .#default.pnpmDeps
nix path-info --json .#default.pnpmDeps | jq -r '.[].narHash' > /tmp/h2
nix store delete --ignore-liveness $(nix path-info .#default.pnpmDeps)

nix build --no-link .#default.pnpmDeps
nix path-info --json .#default.pnpmDeps | jq -r '.[].narHash' > /tmp/h3

diff /tmp/h1 /tmp/h2 && diff /tmp/h2 /tmp/h3 && echo "REPRODUCIBLE"
```

Or, equivalently, deliberately falsify the `hash = "sha256-..."` value on
`pnpmDeps` and run `nix build .#default.pnpmDeps` three times — the
computed hash Nix prints on each failure should be identical.

End-to-end smoke:

```bash
nix build .#default
./result/bin/nanocoder --version
```

CI side: triggering `.github/workflows/update-nix.yml` via
`workflow_dispatch` on this branch should now succeed (it currently
fails-loud / no-ops on `main`).

## Caveat / known follow-up

If `narHash` still drifts across runs, the most likely remaining sources
are SQLite's header change-counter and freelist, which sorted re-insert
alone does not fully reset. The fix in that case is to append a `VACUUM;`
(via `sqlite3 $storePath/v11/index.db 'VACUUM;'`) after the
`@pnpm/store.index` rewrite. Happy to add that if CI / local reproduction
shows it's needed.

## Checklist

- [x] Self-review completed
- [x] Documentation updated where relevant (inline comments in `flake.nix`
      explain each workaround and the rollback path)
- [x] No breaking changes
- [x] No new runtime logging needed (build-time only)


will-lamerton:
Hi @ragini-pandey - I notice this is still draft is there still work to do or are you there with it for testing? I can if needs be :)

Let me know!

ragini-pandey:
> Hi @ragini-pandey - I notice this is still draft is there still work to do or are you there with it for testing? I can if needs be :)
> 
> Let me know!

Hi @will-lamerton, It's ready for testing
from my side; the only reason it's still in draft is that I don't have
Nix installed locally, so I haven't been able to verify the
`pnpmDeps` `narHash` is actually stable across runs. The CI checks that
ran (9/10 green) don't exercise the Nix path.

Could you give it a spin locally? The acceptance test is just running
`nix build .#default.pnpmDeps` two or three times in a clean store and
confirming the `narHash` matches — there's a copy-pasteable script in
the updated PR description.

If the hash still drifts, the most likely remaining culprit is SQLite's
header change-counter / freelist, in which case appending
`sqlite3 $storePath/v11/index.db 'VACUUM;'` after the
`@pnpm/store.index` rewrite should finish the job. I'll push that as a
follow-up commit the moment we see a failing run.

Flipping to "ready for review" now — please mark it back to draft if
your local repro fails and I'll iterate.

will-lamerton:
Thanks @ragini-pandey for digging into this and proposing a workaround. The diagnosis (v11/index.db non-determinism from B-tree page order + msgpack `checkedAt` timestamps) was on the right track and matches what nixpkgs PR [#505103](https://github.com/NixOS/nixpkgs/pull/505103) ended up doing with its `pnpm-fixup-state-db` helper, which has now landed in nixpkgs-unstable.

Closing in favour of an alternative fix that turned out to be much smaller once #505103 unblocked `pkgs.pnpm_11`. The *residual* drift (after the helper runs) traces back to a 1-character shell bug in upstream `fetch-pnpm-deps/default.nix`: two `export pnpm_config_X val` lines (missing `=`) leave `side-effects-cache` at its default `true`, which then makes pnpm 11 write a non-deterministic `sideEffects` field into one row of `v11/index.db` (for `ink@6.8.0` in our case). Setting those two env vars as derivation attrs in our flake works around it until the upstream typo is patched.

Verified locally with two cold Nix builds (GC between) producing identical narHash. Landed on main as [b5a9f87](https://github.com/Nano-Collective/nanocoder/commit/b5a9f87).

Thanks again for taking a look at this while #505103 was still in flight.

PR #531, opened 2026-05-28
fix: read_file returns empty marker instead of throwing on empty files

### LINKED ISSUE #530: read_file throws on empty files — breaks valid use cases
### Summary

`read_file` throws when reading an empty file, forcing the model to treat a successful read as a failure.

### Location

[source/tools/read-file.tsx:100-103](https://github.com/Nano-Collective/nanocoder/blob/main/source/tools/read-file.tsx#L100-L103)

```ts
if (content.length === 0) {
    throw new Error(`File "${args.path}" exists but is empty (0 tokens)`);
}
```

### Why this is a bug

Empty files are perfectly valid in real codebases:

- `.gitkeep` and similar placeholder files
- Freshly-created stub files (e.g. `touch new-module.ts` before writing)
- Log files truncated to zero
- Generated files whose first write has not happened yet

Throwing causes the model to treat a successful FS read as an error. In practice this leads to wasted tool calls — the model retries with `bash cat`, `list_directory`, or `metadata_only`, only to confirm what the first call already knew: the file exists and is empty.

### Suggested fix

Return the empty content with a short marker so the model can distinguish "file is empty" from "file does not exist":

```ts
if (content.length === 0) {
    return '[file is empty]';
}
```

This keeps the call successful, conveys the state in one token-cheap line, and aligns with how other tools surface "no results" (e.g. `search_files`).

### Labels

`bug`, `ux`

ragini-pandey:
## Description

`read_file` previously threw `File "..." exists but is empty (0 tokens)` whenever the target file was empty. Empty files are valid (`.gitkeep`, freshly-created stubs, log files truncated to zero), so the throw turned a successful FS read into an error and pushed the model into wasteful retries (`bash cat`, `list_directory`, `metadata_only`) just to confirm the same state.

This PR replaces the throw with a successful return of the marker string `[file is empty]`, so the model can distinguish "file is empty" from "file does not exist" without an exception path. Updates the corresponding AVA test.

Closes #530.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

### Automated Tests

- [x] New features include passing tests in `.spec.ts/tsx` files (existing empty-file test updated to assert the new return value)
- [x] All existing tests pass (`pnpm run test:ava source/tools/read-file.spec.tsx` — 38/38 pass)
- [x] Tests cover both success and error scenarios (empty file returns marker; nonexistent file still throws)

### Manual Testing

- [ ] Tested with Ollama
- [ ] Tested with OpenRouter

_Change is provider-agnostic — the fix is in the local tool handler, not the LLM client path._

## Checklist

- [x] Code follows project style guidelines (Biome ran via lint-staged on commit)
- [x] Self-review completed
- [x] Documentation updated (if needed) — none required; behavior change is internal to the tool
- [x] No breaking changes (or clearly documented) — return type unchanged (`string`); callers that previously caught the throw will simply see a successful response
- [x] Appropriate logging added using structured logging (see [CONTRIBUTING.md](../CONTRIBUTING.md#logging)) — N/A, no new log sites

### DIFF
--- a/source/constants.ts
+++ b/source/constants.ts
@@ -56,6 +56,7 @@ export const FILE_READ_CHUNKING_HINT_THRESHOLD_LINES = 500;
 export const FILE_READ_CHUNK_SIZE_LINES = 250;
 export const CHARS_PER_TOKEN_ESTIMATE = 4;
 export const MAX_LINE_LENGTH_CHARS = 10_000; // Lines longer than this are likely minified/binary
+export const EMPTY_FILE_MARKER = '[file is empty]';
 
 // === TERMINAL AND UI ===
 export const PATH_LENGTH_NARROW_TERMINAL = 30;
--- a/source/tools/read-file.spec.tsx
+++ b/source/tools/read-file.spec.tsx
@@ -643,7 +643,7 @@ test.serial('read_file throws error for nonexistent file', async t => {
 	);
 });
 
-test.serial('read_file throws error for empty files', async t => {
+test.serial('read_file returns empty marker for empty files', async t => {
 	t.timeout(10000);
 	const testDir = join(process.cwd(), 'test-read-empty-temp');
 
@@ -651,17 +651,14 @@ test.serial('read_file throws error for empty files', async t => {
 		mkdirSync(testDir, {recursive: true});
 		writeFileSync(join(testDir, 'empty.ts'), '');
 
-		await t.throwsAsync(
-			async () => {
-				await readFileTool.tool.execute!(
-					{
-						path: join(testDir, 'empty.ts'),
-					},
-					{toolCallId: 'test', messages: []},
-				);
+		const result = await readFileTool.tool.execute!(
+			{
+				path: join(testDir, 'empty.ts'),
 			},
-			{message: /exists but is empty/},
+			{toolCallId: 'test', messages: []},
 		);
+
+		t.is(result, '[file is empty]');
 	} finally {
 		rmSync(testDir, {recursive: true, force: true});
 	}
--- a/source/tools/read-file.spec.tsx
+++ b/source/tools/read-file.spec.tsx
@@ -4,6 +4,7 @@ import test from 'ava';
 import {render} from 'ink-testing-library';
 import React from 'react';
 import {themes} from '../config/themes';
+import {EMPTY_FILE_MARKER} from '../constants';
 import {ThemeContext} from '../hooks/useTheme';
 import {readFileTool} from './read-file';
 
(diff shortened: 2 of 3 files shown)


ragini-pandey:
Thanks for the Copilot review. Pushed [f6ab5ad0](https://github.com/Nano-Collective/nanocoder/pull/531/commits/f6ab5ad0) addressing the actionable items, and force-pushed to drop the Copilot Autofix commit (`'[file is empty]'` → `''`) — see rationale below.

**Adopted:**
- Extracted `EMPTY_FILE_MARKER` in [`source/constants.ts`](https://github.com/Nano-Collective/nanocoder/blob/fix/read-file-empty/source/constants.ts) so the literal can't drift between implementation and tests.
- Added two tests: empty file + `start_line`/`end_line`, and empty file + `metadata_only` (verifies the metadata path is not short-circuited by the empty check, since `metadata_only` runs first).

**Not adopted, with reasoning:**
- *Return a structured `{ content, isEmpty }` shape instead of a string marker.* All tool handlers in `ToolManager` return `string`; restructuring just this one would force every consumer (formatter, model contract, MCP bridge) to special-case it. The bracket-marker convention is already used elsewhere in this file (`[Medium file - …]`, `[Large file - …]`, `[Use read_file to view file contents]`), so this stays consistent.
- *Return `''` (Copilot Autofix).* That loses the signal the change is meant to provide: the model can't distinguish "file exists and is empty" from "got back nothing" and goes back to retrying with `bash cat` / `list_directory`, which is the exact regression the PR fixes.
- *Spec indentation drift.* Biome ran via lint-staged on commit; the file is formatted.

CI green locally: 40/40 tests pass on `pnpm run test:ava source/tools/read-file.spec.tsx`, `tsc --noEmit` clean.

ragini-pandey:
CI status on the latest commit ([f6ab5ad0](https://github.com/Nano-Collective/nanocoder/pull/531/commits/f6ab5ad0)) — flagging for maintainers since I don't have rerun rights:

**1. Unit Tests & Coverage — `FAILURE` (unrelated to this PR)**

Three tests failed, all in `source/wizards/steps/`:
- `field-input-view › boolean field highlights No when currentValue is "false"`
- `model-selection-list › slash enters search mode and typing filters models`
- `model-selection-list › d in search mode updates the query instead of completing`

These are Ink rendering / keyboard-input timing tests in code paths this PR doesn't touch. Verified locally:

```
$ pnpm run test:ava          # exit 0, all tests pass including the three above
$ pnpm run test:ava source/tools/read-file.spec.tsx   # 40/40 pass
```

This looks like CI flakiness (timing-sensitive Ink tests) rather than a regression. A rerun should clear it.

**2. Package Audit — `FAILURE` (pre-existing)**

```
high  │ tmp has Path Traversal via unsanitized prefix/postfix
Vulnerable versions: <0.2.6
```

This advisory exists on `main` independent of this PR — the only files this PR changes are `source/constants.ts`, `source/tools/read-file.tsx`, and `source/tools/read-file.spec.tsx`. No dependency changes.

**3. Everything else green**

Linting, Type Checks, Format Checks, Unused Dependencies, Verify Build, Semgrep, CodeQL — all ✅.

Happy to push an empty commit to retrigger CI if a rerun isn't easier on your end.

Issue #542, opened 2026-06-04
[Bug] Session bugs: autosave duplicate-session race, on-disk history truncation, and /resume not rebasing the session ID


JimStenstrom:
## Description

The session subsystem (`useSessionAutosave` plus the shared session‑ID wiring) has three bugs:

1. **Duplicate / orphaned sessions (race):** on the first save of a fresh chat, multiple `saveSession()` calls can run concurrently before `currentSessionId` is set, so each calls `createSession()` — a second session file + index entry is created and then orphaned.
2. **Silent data loss (truncation):** autosave truncates the message array to `maxMessages` (default 1000) and writes the truncated array **back over the existing session file**, permanently discarding older messages from disk — not just from the in‑memory/model context.
3. **`/resume` never rebases the session ID:** `setKeyGeneratorSessionId()` is defined, documented, and unit‑tested as the way to re‑key the shared session ID when a persisted session is resumed, but it is **never called in production**. `applySession` (the `/resume` handler) sets `currentSessionId` but not the key‑generator ID, so `SessionTracker` and React keys keep the random `randomBytes(4)` startup ID instead of the resumed session's ID. This leaves an acceptance criterion of #229 ("SessionTracker uses shared session ID" / rebase on resume) effectively unwired — folding it in here since it's the same subsystem.

## Environment

- **OS**: Any (code‑level, platform‑independent)
- **Node version**: ≥ 22
- **Nanocoder version**: 1.27.0 (`main`)
- **Provider**: Any (provider‑agnostic)
- **Model**: Any

## Steps to Reproduce

**A — Duplicate sessions (race):**
1. Start a fresh chat (no resumed session), autosave enabled (the default).
2. Send a message that produces a multi‑step turn (assistant text **plus** a tool call), so the conversation loop calls `setMessages` several times within the one turn.
3. Inspect the sessions directory / index: two session files can exist for the single conversation; `currentSessionId` points at only the last, orphaning the other (and it counts against `maxSessions`).

> Root cause: on the first turn `lastSaveRef.current === 0`, so the effect takes the immediate `void saveSession()` branch. Each `setMessages` re‑runs the effect and fires another `saveSession()` **before** the first one's `await createSession()` has resolved `setCurrentSessionId`. Both runs observe `currentSessionId === null` and both create a session.

**B — Truncation data loss:**
1. Set `sessions.maxMessages` low (e.g. `10`) in config — or simply run a conversation past the 1000 default.
[...]

PR #545, opened 2026-06-08
fix: resolve autosave race, history truncation, and resume rebase

akramcodez:
## Description

Resolves three issues in the session subsystem (#542):

1. **Duplicate Session Race Condition** - Autosave operations are now serialized through a promise chain (`saveChainRef`), preventing concurrent session creation during the initial save flow.
2. **On-disk Message History Truncation** - Full conversation history is now persisted to disk. The `maxMessages` limit is treated as a model-context concern rather than a persistence concern.
3. **Session ID Rebase on `/resume`** - Added the missing `setKeyGeneratorSessionId(session.id)` call inside `applySession` so shared session state correctly reflects the resumed session.

## Type of Change

* [x] Bug fix
* [ ] New feature
* [ ] Breaking change
* [ ] Documentation update

### Automated Tests

* [x] All existing tests pass (`pnpm test:all`)
* [x] Build succeeds (`pnpm run build`)
* [x] Changes verified locally

### Manual Testing

* [x] Tested session creation and autosave behaviour
* [x] Tested session resume flow
* [x] Verified full message history persists across saves

## Checklist

* [x] Self-review completed
* [x] No breaking changes introduced
* [x] Documentation updated where necessary

### DIFF
--- a/docs/configuration/index.md
+++ b/docs/configuration/index.md
@@ -146,7 +146,7 @@ Configure automatic session saving and retention. See [Session Management](../fe
 | `autoSave` | boolean | `true` | Enable/disable automatic session saving |
 | `saveInterval` | number | `30000` | Milliseconds between saves (minimum 1000) |
 | `maxSessions` | number | `100` | Maximum sessions to keep (minimum 1) |
-| `maxMessages` | number | `1000` | Maximum messages saved per session — older messages are truncated (minimum 1) |
+| `maxMessages` | number | `1000` | Maximum messages sent to the model in interactive/headless chat (minimum 1). Preserves on-disk history and system messages, capping only the context window. |
 | `retentionDays` | number | `30` | Auto-delete sessions older than this (minimum 1) |
 | `directory` | string | (platform default) | Custom storage directory for session files |
 
--- a/docs/features/session-management.md
+++ b/docs/features/session-management.md
@@ -74,6 +74,6 @@ Customize session behaviour in your `agents.config.json`:
 | `autoSave` | `true` | Enable/disable automatic saving |
 | `saveInterval` | `30000` | Milliseconds between saves (minimum 1000) |
 | `maxSessions` | `100` | Maximum sessions to keep (minimum 1) |
-| `maxMessages` | `1000` | Maximum messages saved per session — older messages are truncated (minimum 1) |
+| `maxMessages` | `1000` | Maximum messages sent to the model (context window capping) — on-disk history is NOT truncated (minimum 1) |
 | `retentionDays` | `30` | Auto-delete sessions older than this (minimum 1) |
 | `directory` | (platform default) | Custom storage directory |
--- a/source/acp/acp-conversation.ts
+++ b/source/acp/acp-conversation.ts
@@ -7,6 +7,7 @@ import {requestToolPermission} from '@/acp/acp-permission';
 import {requestUserChoice} from '@/acp/acp-question';
 import type {AcpSession} from '@/acp/acp-session';
 import {type AcpToolCallMeta, buildToolCallMeta} from '@/acp/acp-tool-call';
+import {getAppConfig} from '@/config/index';
 import {processToolUse} from '@/message-handler';
 import {parseToolCalls} from '@/tool-calling/index';
 import {resolveToolApproval} from '@/tools/approval-policy';
@@ -19,6 +20,7 @@ import type {
 	ToolCall,
 	ToolResult,
 } from '@/types/core';
+import {capMessagesForModel} from '@/utils/message-capping';
 import {toOptionString} from '@/utils/type-helpers';
 
 const MAX_TURNS = 50;
@@ -85,8 +87,12 @@ export async function runAcpConversation(
 			return {stopReason: 'end_turn'};
 		}
 
+		const sessionConfig = getAppConfig().sessions;
+		const maxMessages = sessionConfig?.maxMessages ?? 1000;
+		const cappedMessages = capMessagesForModel(messages, maxMessages);
+
 		const result = await client.chat(
-			[systemMessage, ...messages],
+			[systemMessage, ...cappedMessages],
 			tools,
 			callbacks,
 			abortController.signal,
--- a/source/hooks/chat-handler/conversation/conversation-loop.spec.ts
+++ b/source/hooks/chat-handler/conversation/conversation-loop.spec.ts
@@ -1,5 +1,5 @@
 import test from 'ava';
-import {clearAppConfig} from '@/config/index.js';
+import {clearAppConfig, getAppConfig} from '@/config/index.js';
 import {resetShutdownManager} from '@/utils/shutdown/shutdown-manager.js';
 import {processAssistantResponse, resetFallbackNotice, resetLastTurnHadReasoning} from './conversation-loop.js';
 import type {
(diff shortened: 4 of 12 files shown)


will-lamerton:
Thanks for putting this together. The resume-rebase piece (one new line in `applySession`, `useAppHandlers.tsx:445`) is exactly right — order is correct, the rebased id is in place before the first `generateKey` after the call, and the import is clean.

The other two items have problems that block merge as-is. Detailed below.

---

## 1. The race is not actually fixed

The chained-promise serialisation is necessary but not sufficient. `useSessionAutosave.ts` captures `const capturedSessionId = currentSessionId` at effect-fire time and uses that for the create-vs-update branch. On a fresh conversation:

1. Effect run with `messages=[u1]`, `currentSessionId=null` → `capturedSessionId = null`. `schedule()` enqueues `doSave_1`.
2. Conversation loop dispatches the assistant. Effect re-runs with `messages=[u1, a1]`, `currentSessionId=null` (React has not yet flushed any `setCurrentSessionId` from a previous save). `capturedSessionId = null` again. `doSave_2` enqueued.
3. `doSave_1` resumes after the `initPromiseRef` await → `capturedSessionId` is still `null` → `createSession()` → `setCurrentSessionId(id1)`.
4. `doSave_2` resumes → `capturedSessionId` is still `null` (it's a local const from step 2) → `createSession()` → `setCurrentSessionId(id2)`.

Two session files for one conversation, `id1` orphaned. Same outcome as the pre-PR code, just less wall-clock overlap. The header comment's claim that "only one `createSession()` call can be in-flight per conversation" is only true in the trivial sense (one at a time across the chain), not in the sense the issue is asking for.

A workable fix is to consult the *live* session id at execution time, not the captured snapshot. E.g. mirror `currentSessionId` into a ref via a tiny effect and have `doSave` read the ref after the `initPromiseRef` await:

```ts
const currentSessionIdRef = useRef(currentSessionId);
useEffect(() => { currentSessionIdRef.current = currentSessionId; }, [currentSessionId]);
// ...inside doSave, after `if (!initialized) return;`:
const liveId = currentSessionIdRef.current;
if (liveId) { /* update path */ } else { /* create path */ }
```

The `lastSaveRef` / `setTimeout` debounce also won't save you here, because the first save takes the `timeSinceLastSave >= saveInterval` branch and runs `schedule()` immediately, before the second effect run gets a chance to coalesce.

## 2. `maxMessages` is now silently dead and the doc comment is wrong

Removing `messages.slice(-maxMessages)` from the autosave path is the right direction. The header comment then says:

> "maxMessages bounds only what is sent to the model (enforced at prompt-build time), not what is stored in the session file."

I searched the codebase for any other consumer of `sessions.maxMessages` — the only readers are `source/config/index.ts:255-258` (schema/defaults), `source/types/config.ts:206` (type), and the docs (`docs/configuration/index.md:136,149`, `docs/features/session-management.md:64,77`). The prompt path (`source/ai-sdk-client/ai-sdk-client.ts:153` → `chat-handler.ts:60`) does not slice. Auto-compact is token-budget, not message-count, and is unrelated.

So the cap is gone, but the field, the type, the docs, and the doc comment all still promise a behaviour that doesn't exist. Anyone with `maxMessages: 1000` in `nanocoder-preferences.json` will see unbounded growth in `~/.nanocoder/sessions/*.json`, the opposite of the documented behaviour.

Two acceptable resolutions — please pick one:

- **Move the cap to the prompt-build path** (e.g. in `ai-sdk-client/ai-sdk-client.ts:153` `chat()` or wherever the message array is assembled for the LLM). Apply `messages.slice(-maxMessages)` there, leave the autosave writing the full array, and update the config docs to say the cap bounds what is sent to the model, not what is saved.
- **Remove the field, the docs, the type, and add a CHANGELOG entry** noting the cap is gone.

The current state (cap removed, field and docs unchanged, doc comment pointing at a non-existent enforcement site) is the worst of both options and shouldn't ship.

## 3. Tests

The PR adds zero new tests. Given the timing-sensitive nature of both #1 and #2, that's a real gap. Suggested starting point, modelled on `source/hooks/chat-handler/useChatHandler.spec.tsx`:

- `source/hooks/useSessionAutosave.spec.tsx` with a mocked `SessionManager` (track `createSession` / `saveSession` call counts and arguments). Render the hook with `currentSessionId = null`, fire the autosave effect twice in one tick as the conversation loop would, assert `createSession` is called exactly once and `saveSession` is called with the full messages array (no `slice`).
- A test for the `applySession` rebase wiring — even a small one — that asserts `setKeyGeneratorSessionId` is called with `session.id` and that the very next `generateKey` call uses the new id prefix. This is straightforward against the existing `key-generator.spec.ts` test patterns.

## Minor

- Header comment on `useSessionAutosave.ts` is well-intentioned but contains the false "enforced at prompt-build time" claim. Either fix the claim or fix the enforcement site.
- `/clear` (`useAppHandlers.tsx:145` and `app-util.ts:273`) calls `setCurrentSessionId(null)` but never resets the key-generator session id, so `/resume A` → `/clear` → new conversation still carries `A`'s id prefix in its React keys. Not blocking for this PR, but worth a follow-up issue.
- `pnpm run test:types` and `pnpm run test:lint` pass on the branch (only a pre-existing biome schema-version diagnostic, unrelated).

will-lamerton:
Big improvement on this iteration — the race fix and the `/clear` follow-up are both resolved. Two remaining things, one bigger than the other.

## 1. Race fix — confirmed closed

The synchronous `currentSessionIdRef.current = newSession.id` write inside `doSave` (lines 166 and 184) is what makes this work. The ref is the source of truth at execution time, the React state update is just for re-rendering purposes. Walking through a fresh conversation:

1. Effect run #1, `currentSessionId=null`, schedules `doSave_1` immediately. `capturedMessages=[u1]`. Ref mirrors `null`.
2. Conversation loop dispatches the assistant. Effect run #2 fires, `currentSessionId` still `null` (no React flush yet). `doSave_2` queued.
3. `doSave_1` resumes after the init await, reads ref → `null`, takes create branch. `createSession()` returns `id1`. **Synchronously** writes `currentSessionIdRef.current = id1` and schedules `setCurrentSessionId(id1)`.
4. `doSave_2` resumes after its own await, reads ref → `id1` (set synchronously in step 3). Takes the update branch. `readSession(id1)` → exists → `saveSession(session)`.

The ref write is independent of React's batching, so the timing works in any rendering mode. Removing `currentSessionId` from the dep array is also correct — the effect should re-run on message changes, not on session id changes, since the session id is the very thing the ref pattern protects.

## 2. `/clear` — confirmed fixed

`useAppHandlers.tsx:150` resetting to `randomBytes(4).toString('hex')` matches `ensureSessionId()`'s output shape. The comment explains the why. Good.

## 3. `maxMessages` cap — only one of four call sites, docs are stale

The cap is now applied in `source/hooks/chat-handler/conversation/conversation-loop.tsx:249-255` with the system message correctly prepended outside the slice. The doc comment in `useSessionAutosave.ts` has been updated to point at "the conversation loop" rather than "prompt-build time", which is now accurate for that one site.

But there are five `client.chat(...)` call sites in the codebase and only one is capped:

- `source/hooks/chat-handler/conversation/conversation-loop.tsx:257` — capped (new) ✓
- `source/acp/acp-conversation.ts:88` — uncapped
- `source/plain/conversation.ts:80` — uncapped
- `source/subagents/subagent-executor.ts:396` — uncapped (probably fine, sub-agents manage their own message construction)
- `source/utils/llm-summariser.ts:93` — uncapped (probably fine, constructed summary prompt)

The cap's *meaning* has changed from "disk truncation" to "LLM-call capping", but only the interactive chat path enforces it. If the intent is "cap the LLM call everywhere interactive", `acp-conversation.ts` and `plain/conversation.ts` need the same slice. If the intent is "cap only the conversation loop", that's a narrower contract that should be documented.

Worse, the docs are still describing the old behaviour. `docs/configuration/index.md:149` and `docs/features/session-management.md:77` both still say:

> `maxMessages` — number — `1000` — Maximum messages saved per session — older messages are truncated

A user reading the docs will expect truncation on disk; they'll get capping on the LLM call (in the interactive loop only) and unbounded growth on disk. That's a docs-vs-code mismatch.

Two acceptable resolutions — please pick one:

- Extract a small `capMessagesForModel(messages)` helper, apply it in `acp-conversation.ts` and `plain/conversation.ts` as well, and update the docs to say the cap bounds what is sent to the model in interactive chat. Sub-agents and the summariser can stay as-is.
- Or document the narrower contract: "maxMessages caps the messages sent to the model in interactive chat; persisted history and other LLM paths are uncapped."

The current state (one of four call sites updated, docs unchanged) is the worst of both options.

## 4. The `maxMessages` slice can orphan tool-call pairs (correctness, higher priority than the docs gap)

Separate from the rollout question above: the cap is a naive `messages.slice(-maxMessages)` applied directly to the live LLM request. That can cut between an assistant `tool_use` message and its matching `tool_result`. When the boundary lands there you send a `tool_result` with no preceding `tool_use` (or a dangling `tool_use` whose result got sliced off), and providers reject that sequence - Anthropic returns a 400 (`unexpected tool_use_id` / `tool_result` without `tool_use`), and the OpenAI-compatible path errors similarly.

This is a real regression risk, not a theoretical one. The pre-PR code had the same naive slice, but it only ever wrote to disk, so an invalid sequence never reached API validation. Moving the slice onto the request path turns it into a request failure whenever the cap actually bites - i.e. any tool-heavy session longer than `maxMessages`, or any user who sets `maxMessages` low.

Note that the existing orphan filter does **not** cover this. `createPrepareStepHandler` in `source/ai-sdk-client/chat/streaming-handler.ts:40-66` only skips tool results that *immediately follow an empty assistant message* - it walks forward from an empty assistant and drops the trailing tool messages. The slice case is different: the assistant `tool_use` message is removed entirely, so the kept array *starts* on a `role: 'tool'` result with no preceding assistant at all. That leading orphan never matches the filter's condition, so it survives conversion (`message-converter.ts:33` is a straight `map` with no repair) and reaches the provider. The codebase already knows orphaned tool results are fatal here - see the analogous 400 documented at `message-converter.ts:7-8` for empty assistant messages.

The fix is to snap the slice to a safe boundary rather than cutting blindly:

- Don't let the window start on a `tool` / `tool_result` message - walk back (or forward) so the first kept message is a complete turn.
- Don't let the window end on an unanswered `tool_use`.

Whatever `capMessagesForModel(messages)` helper comes out of item #3 should own this boundary logic, so every interactive call site gets correct slicing for free. Worth a dedicated test (a history whose `(len - maxMessages)` boundary falls mid tool-call pair, asserting the sent array starts and ends on a valid turn).

## 5. Tests

Still no tests added. The diff stat for `*.spec.*` is empty. The autosave hook is now meaningfully more complex (refs, three save paths, chained promise), and the `maxMessages` behaviour change is unverified. Concrete asks:

- `source/hooks/useSessionAutosave.spec.tsx` with a mocked `SessionManager`. Drive the autosave effect twice in one tick on a fresh conversation, assert `createSession` is called exactly once and `saveSession` is called with the full message array (no `slice`).
- A test in `conversation-loop.spec.tsx` that drives `processAssistantResponse` with a 1500-message history under `maxMessages: 1000` and asserts the client received the last 1000 (plus system message = 1001).
- A test for the boundary case in #4: a history where the `(len - maxMessages)` cut falls between a `tool_use` and its `tool_result`, asserting the sent array does not start on an orphaned `tool_result` or end on an unanswered `tool_use`.
- A test that `clearMessages` calls `setKeyGeneratorSessionId` with a fresh id, or more loosely that keys generated after `/clear` are not prefixed with the prior session's id.

`source/hooks/chat-handler/useChatHandler.spec.tsx` and `source/session/key-generator.spec.ts` are clean templates. Happy to draft the autosave spec in a follow-up if useful.

## Minor

- `pnpm run test:types` and `pnpm run test:lint` pass. `pnpm run test:ava source/hooks/chat-handler/conversation/conversation-loop.spec.ts` and `source/session/key-generator.spec.ts` both pass. Only a pre-existing biome schema-version diagnostic, unrelated to this PR.
- The `useMemo` wrapping `clearMessages` is mildly unusual for a function-returning-closure (the rest of the file uses `useCallback`), but it's the existing pattern. Not a blocker.

will-lamerton:
Reviewed the new commit. All four of the prior asks are addressed and the tests pass locally. Two minor observations, neither blocking.

## 1. `capMessagesForModel` is the right shape

Extracting the slice into a helper is the right call, and the tool-boundary walk-back is a thoughtful addition I didn't ask for. Walking the algorithm:

- If `messages[start]` is `role: 'tool'`, decrement `start` until it's not.
- Otherwise slice from `start`.

The walk-back stops at the assistant message that initiated the tool calls (since `assistant ≠ tool`), so the resulting slice starts with the assistant message and includes all of its tool results in order. The four test cases in `message-capping.spec.ts` cover the no-slice, normal-slice, single-tool-result, and multi-tool-result cases. All pass.

One edge case worth being aware of (not a bug, just behaviour): if a tool sequence is longer than `maxMessages`, the resulting slice can be *larger* than `maxMessages`. E.g. `[user, assistant(tool_calls), tool, tool, user]` with `maxMessages=2` produces a 4-message slice `[assistant, tool, tool, user]`. Tool-boundary expansion wins over the cap, which is the right call — sending a slightly longer but well-formed prompt beats breaking tool-call pairing. A one-line comment noting that the function returns "at most `maxMessages`, or more if a tool sequence straddles the boundary" would help future readers; not blocking.

## 2. `acp-conversation.ts` and `plain/conversation.ts` now use the helper

Both files now go through `capMessagesForModel` (lines 90-92 in `acp-conversation.ts`, lines 82-84 in `plain/conversation.ts`), with the system message correctly prepended outside the slice. The cap is now applied consistently across the three interactive LLM call sites. Good.

I left `subagent-executor.ts:396` and `llm-summariser.ts:93` out of the ask list because they manage their own message construction. That's still correct — neither path represents a conversational history that the user would want capped at `maxMessages`.

## 3. Docs are updated

`docs/configuration/index.md:149` and `docs/features/session-management.md:77` now read:

> `maxMessages` — number — `1000` — Maximum messages sent to the model in interactive/headless chat (minimum 1). Preserves on-disk history and system messages, capping only the context window.

And the equivalent wording in `session-management.md`. This matches the implementation. Good.

## 4. Tests added — all pass

- `source/utils/message-capping.spec.ts` (4 tests) — unit tests for the helper, including the tool-boundary walk-back cases.
- `source/hooks/useSessionAutosave.spec.ts` (6 tests) — regression coverage for the three bugs from #542, plus "demonstrates the old bug" companion tests for A and B and a companion for C. The chain test correctly models the live-ref-after-await pattern.
- `source/hooks/chat-handler/conversation/conversation-loop.spec.ts` (+2 tests) — one for the basic `maxMessages` cap, one for the tool-boundary case. Both pass.
- `source/hooks/useAppHandlers.spec.tsx` (+1 test) — `clearMessages resets key generator session ID`, which exercises the `/clear` follow-up fix end-to-end through the real `useAppHandlers` hook. Good integration coverage.

I ran the affected specs locally:

- `pnpm run test:ava source/utils/message-capping.spec.ts` → 4/4 pass
- `pnpm run test:ava source/hooks/useSessionAutosave.spec.ts` → 6/6 pass
- `pnpm run test:ava source/hooks/useAppHandlers.spec.tsx` → 11/11 pass (including the new one)
- `pnpm run test:ava source/hooks/chat-handler/conversation/conversation-loop.spec.ts` → 39/39 pass (including the two new ones)
- `pnpm run test:types` and `pnpm run test:lint` clean (only the pre-existing biome schema-version diagnostic)

Nice work on iterating this. The tool-boundary walk-back in `capMessagesForModel` in particular is a nice touch I didn't ask for but is clearly the right thing.

Issue #501, opened 2026-05-04
[Bug]


zerophilmister:
---

## Description

When the LLM provider returns a context-length error (e.g., `request exceeds the available context size`), Nanocoder enters an infinite retry loop instead of surfacing the error. On each retry, it appends a continuation message to the conversation history, causing the prompt to grow and guaranteeing every subsequent request also fails.

## Environment

- **OS**: Bazzite Linux (Fedora Silverblue-based)
- **Node version**: v25.9.0
- **Nanocoder version**: 1.25.2
- **Provider**: LM Studio (local OpenAI-compatible API, `http://127.0.0.1:1234/v1/chat/completions`)
- **Model**: `qwen/qwen3.6-35b-a3b` (also reproduced with `google/gemma-4-26b-a4b`)

## Steps to Reproduce

1. Load a model in LM Studio with the default context length of `4096`.
2. Launch Nanocoder.
3. Send any user message, e.g., `is .net installed?`
4. Observe the terminal and LM Studio logs.

## Expected Behavior

Nanocoder should detect the unrecoverable context-length error from the provider, halt execution, and display a clear message to the user such as:

> "The prompt (5,360 tokens) exceeds the model's context window (4,096 tokens). Please increase the provider's context length or reduce the number of enabled tools."

## Actual Behavior

LM Studio rejects the request immediately:

```
error: request (5360 tokens) exceeds the available context size (4096 tokens), try increasing it
```

Nanocoder interprets this as an incomplete response and automatically appends a user message: `"Please continue with the task."` (or the user manually types `continue`). The next request now contains more tokens than the previous one because the conversation history keeps growing. Steps repeat indefinitely, with the prompt growing by ~7 tokens per iteration:

- Request 1: 2 messages, 5,360 tokens
- Request 2: 3 messages, 5,367 tokens
- Request 3: 4 messages, 5,374 tokens
- ...
- Request 12: 13 messages, 5,423 tokens

The loop only stops when the user manually ejects the model or kills the Nanocoder process. Eventually the model may return an entirely empty response, which Nanocoder reports as:

```
Model returned empty response. This may indicate the model is not responding correctly or the prompt was unclear.
```

---

will-lamerton:
Hey @zerophilmister - thanks for surfacing this! @zerone0x swiftly patched - out in the next release :)

PR #1, opened 2025-08-09
Integrate upstream changes + Enhanced Features Suite

jasonpaulso:
## Summary
Merges latest upstream/main changes while preserving a comprehensive suite of enhancements that significantly expand Nanocoder's capabilities.

### 1. Custom Command System
- **Markdown-based commands** in `.nanocoder/commands/` directory
- **Template variables** with `{{parameter}}` syntax for dynamic content
- **Namespace organization** via folder structure (e.g., `/refactor:dry`)
- **YAML frontmatter** for rich metadata (descriptions, aliases, parameters)
- **Autocomplete integration** with Tab completion
- **Pre-installed commands**: `/test`, `/review`, `/refactor:dry`, `/refactor:solid`

### 2. Advanced Logging & Preferences
- **Configurable log levels**: silent, normal, verbose via `/debug` command
- **Persistent preferences** for provider/model selections per user
- **Session state management** that remembers settings across runs
- **Intelligent output filtering** based on user preferences

### 3. Enhanced MCP Integration
- **Dynamic MCP server management** with real-time tool loading
- **Advanced tool call parsing** supporting multiple MCP formats
- **Server status monitoring** via `/mcp` command
- **Seamless tool integration** that extends AI capabilities on-demand

### 4. OpenAI-Compatible API Provider
- **Universal API support** for LM Studio, vLLM, LocalAI, Ollama Web API
- **Flexible endpoint configuration** with optional authentication
- **Smart provider fallback** when services are unavailable
- **Multi-provider session management**

### 5. Enhanced User Experience
- **Smart autocomplete** with real-time command suggestions
- **Prompt history** accessible via `/history` command  
- **Rich terminal UI** with syntax highlighting and structured output
- **Real-time indicators** showing token usage, timing, and processing status

### Architecture Enhancements
- **Modular tool system** with dynamic registry management
- **Plugin-style extensibility** for adding new capabilities
- **TypeScript-first approach** with comprehensive type safety
- **Clean separation of concerns** between core functionality and extensions

### Developer Experience
- **Enhanced debugging tools** with verbose logging options
- **Comprehensive error handling** with graceful degradation
- **Project-specific configuration** via `agents.config.json`
- **Extensible command system** that developers can easily customize

## 🎯 Upstream Integration
Successfully merged all upstream changes including:
- New `/history` command for prompt management
- Tool organization improvements (`src/core/tools/` structure)
- UI/UX enhancements with bordered content and improved colors
- Code organization refinements

[...]

PR #546, opened 2026-06-08
Force compacting and nudging continue when LLM server throws error because of exceeded context

lordoski:
### Problem
During long-running agent sessions, the LLM can exhaust its token budget (especially with reasoning models) and return empty responses. Previously, after MAX_EMPTY_TURNS nudge retries, the loop would give up immediately — leaving tasks incomplete even though compaction could recover them. Additionally, when the LLM server throws a context-length error mid-task, there was no mechanism to automatically compress the conversation history and resume work resiliently.

### Solution
After exhausting all empty-turn nudges, the loop now force-compacts the conversation context using aggressive mechanical compression before surfacing an error. This gives the model a fresh chance to continue with a reduced context window.

### Changes
- source/constants.ts: Added MAX_COMPACT_RETRIES = 1 constant
- conversation-loop.tsx: New force-compact retry logic in the empty-response branch
- Dynamically imports tokenizer for context compression
- Filters system messages from compressed output before recursion
- Shows user-facing info message with reduction percentage
- Error message now includes total attempts (nudges + compact retries)
- conversation-loop.spec.ts: 8 new tests covering force-compact behavior

### Tests Added
- Attempts compact-and-retry after exhausting empty turns
- Error message includes total attempts including compact retries
- Compact-and-retry resets emptyTurnCount for new cycle
- Falls through to error when compression throws during force-compact
- Compact-and-retry shows info message with reduction percentage
- Compact-and-retry adds nudge message to compressed context
- MAX_COMPACT_RETRIES constant limits retry cycles (prevents infinite loops)
- compactRetryCount parameter is passed through recursion correctly

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

### Automated Tests

### Manual Testing

- [ ] Tested with Ollama
- [ ] Tested with OpenRouter
- [x] Tested with llamacpp

## Checklist

- [x] Self-review completed


will-lamerton:
Hey @lordoski 

Thanks for the contribution, sorry it's taken me a day or so to get to this! The shape is good and the new tests are thorough, but I have a few nits to address and a question to answer before I can approve. There's also a merge conflict with `main` to resolve.

## Blocking

1. **Merge conflict with `main`.** Please rebase and resolve before merge.
2. **Question on PR body needs an answer.** The body mentions recovering from "context-length errors thrown by the LLM server mid-task", but the diff only adds force-compact on the empty-response path - no `context_length_exceeded` catch anywhere. Is that recovery in scope here, or a follow-up? If in scope, the diff is missing it; if a follow-up, the body should say so explicitly.
3. **Nit #1 below is now stronger than I first wrote it** - see the updated version.

## Non-blocking nits (but would be great to address)

1. **Skip the force-compact branch when retrying won't help.** Two cases:
   - **`result.reductionPercentage === 0`** (or `compressedUser.length === updatedMessages.length`) — compression didn't shrink anything, so the retry will hit the same wall. Currently the user sees a misleading `"auto-compacted (0% reduction). Retrying…"` followed by another empty turn.
   - **Last error was a known context-length error.** Main's `0d8867a` (Fixes #501) makes `parseAPIError` recognise `"exceeds the available context size"` etc., and `chat-handler.ts:511` *throws* on a real `APICallError` instead of returning empty — so a hard out-of-context case usually doesn't reach this branch. But if it does (e.g. the AI SDK wrapper hid the cause), PR #546 will eat one force-compact cycle + one empty-turn cycle before falling through. Use `extractRootError` + `parseAPIError` to detect a context-length error and short-circuit to the give-up path, so the user gets the existing clear "Context too large" message immediately.

2. **Redundant `setMessages` on line 812.** It's immediately superseded by `setMessages(messagesAfterCompact)` on line 830, and the recursive call uses the second value. Not a bug (React batches), but the line 812 call muddies intent - I think you can drop it.

3. **`getLastBuiltPrompt()` vs `params.systemMessage` divergence.** Compression is built from the cached `getLastBuiltPrompt()` (with a hard-coded fallback) but the recursive LLM call sends `params.systemMessage`. Same in practice for the main loop, but a subagent-style caller could diverge them. Worth a one-line comment.

4. **Use `pino`, not `console.warn`, in the catch block.** The repo uses pino for structured logging (see `chat-handler.ts:489` for the `logger.warn` pattern in the same flow). Replace the `catch (_err) { /* fall through */ }` with a `logger.warn({ err, compactRetryCount }, 'force-compact failed; falling through')` so production debugging isn't blind. Tests should use `t.context.logger` per the existing pattern.

Thanks again!

lordoski:
> Hey @lordoski
> 
> Thanks for the contribution, sorry it's taken me a day or so to get to this! The shape is good and the new tests are thorough, but I have a few nits to address and a question to answer before I can approve. There's also a merge conflict with `main` to resolve.
> 
> ## Blocking
> 1. **Merge conflict with `main`.** Please rebase and resolve before merge.
> 2. **Question on PR body needs an answer.** The body mentions recovering from "context-length errors thrown by the LLM server mid-task", but the diff only adds force-compact on the empty-response path - no `context_length_exceeded` catch anywhere. Is that recovery in scope here, or a follow-up? If in scope, the diff is missing it; if a follow-up, the body should say so explicitly.
> 3. **Nit [Integrate upstream changes + Enhanced Features Suite #1](https://github.com/Nano-Collective/nanocoder/pull/1) below is now stronger than I first wrote it** - see the updated version.
> 
> ## Non-blocking nits (but would be great to address)
> 1. **Skip the force-compact branch when retrying won't help.** Two cases:
>    
>    * **`result.reductionPercentage === 0`** (or `compressedUser.length === updatedMessages.length`) — compression didn't shrink anything, so the retry will hit the same wall. Currently the user sees a misleading `"auto-compacted (0% reduction). Retrying…"` followed by another empty turn.
>    * **Last error was a known context-length error.** Main's `0d8867a` (Fixes [[Bug] #501](https://github.com/Nano-Collective/nanocoder/issues/501)) makes `parseAPIError` recognise `"exceeds the available context size"` etc., and `chat-handler.ts:511` _throws_ on a real `APICallError` instead of returning empty — so a hard out-of-context case usually doesn't reach this branch. But if it does (e.g. the AI SDK wrapper hid the cause), PR [Force compacting and nudging continue when LLM server throws error because of exceeded context #546](https://github.com/Nano-Collective/nanocoder/pull/546) will eat one force-compact cycle + one empty-turn cycle before falling through. Use `extractRootError` + `parseAPIError` to detect a context-length error and short-circuit to the give-up path, so the user gets the existing clear "Context too large" message immediately.
> 2. **Redundant `setMessages` on line 812.** It's immediately superseded by `setMessages(messagesAfterCompact)` on line 830, and the recursive call uses the second value. Not a bug (React batches), but the line 812 call muddies intent - I think you can drop it.
> 3. **`getLastBuiltPrompt()` vs `params.systemMessage` divergence.** Compression is built from the cached `getLastBuiltPrompt()` (with a hard-coded fallback) but the recursive LLM call sends `params.systemMessage`. Same in practice for the main loop, but a subagent-style caller could diverge them. Worth a one-line comment.
> 4. **Use `pino`, not `console.warn`, in the catch block.** The repo uses pino for structured logging (see `chat-handler.ts:489` for the `logger.warn` pattern in the same flow). Replace the `catch (_err) { /* fall through */ }` with a `logger.warn({ err, compactRetryCount }, 'force-compact failed; falling through')` so production debugging isn't blind. Tests should use `t.context.logger` per the existing pattern.
> 
> Thanks again!

Hello @will-lamerton! Dont worry, Ive been busy too. Regarding the points: 

## Blocking
1. Rebased.
2. I tried the approach of adding a `context_length_exceeded` catch but the solution wasn't reliable because at least on my end (on premise) these kind of errors are mapped as "Model produced no output". For example: 
Image

On nanocoder: 
image

In this kind of situations (as the image shows), when I receive the "Model produced no output" error, I just do /compact --mechanical and then send another continue. Everything keeps working.

So I thought: What if instead of just trying with "continue" nudges 2 times before throwing the "Model produced no output" error, we add a last retry where we compact the chat and then send another "continue" nudge?  This idea turned out to be simple and very effective at least in my workflow where I can confidently leave the agent work con long tasks without failing because of context exceeded.  Ive been working with this feature since.

3. I didnt understand this point, sorry what do you need me to do? Do you mean this: 
> Successfully merged all upstream changes including:
> - New /history command for prompt management
> - Tool organization improvements (src/core/tools/ structure)
> - UI/UX enhancements with bordered content and improved colors
> - Code organization refinements


## Non-Blocking
I'll be looking into them.

Thanks for the feedback!

lordoski:
Ok @will-lamerton regarding the non blocking points: 
1.1: Ok, this was actually solved by leaving just one retry with compact. More than that is not necessary I think. 
1.2: Ok, I didnt know about this closed reported bug: https://github.com/Nano-Collective/nanocoder/issues/501 but the problem seems to keep happening even with the last version. 🤔 (I updated it a couple of days ago). Anyways, IMHO nanocoder should try to keep doing the work and avoid halting because it exceeded the token limit (the llm itself is the only one who know this, any token calculation is an approximation). Makes the system more resilient and let me actually disconnect from the workstation. 

2. Ok
3. Ok 🤖 
4. Ok

will-lamerton:
Thanks @lordoski, this all makes sense and I can see you've already pushed the non-blocking fixes - appreciate it.

## Blocking

1. Rebase: 👍 thanks.

2. That fully answers it, the screenshots make it more clear to me. The key insight here is that on your server a context-exceeded condition surfaces as an empty response ("Model produced no output"), not as a typed `context_length_exceeded` / `APICallError` - so the empty-response branch *is* the right home for this recovery and there's no separate catch to add.

3. Sorry, that was a GitHub auto-link, not anything real. The hashtag number in my comment got turned into a link to repo PR 1, but I just meant my non-blocking nit numbered "1" below - nothing to do with the upstream-merge PR. You already engaged with it in your reply, so there's nothing to do here. My fault for the confusion :)

## Non-blocking

1. You've changed my mind on half of this:

- The "detect a context-length error and short-circuit to give-up" idea I think should be dropped. Your point is right with force-compacting reduces the context, so retrying after a compact is the correct move even for a genuine context-length case, since it might now fit. And in your environment `parseAPIError` never sees it anyway (it's empty output, not an `APICallError`). Short-circuiting would fight the whole point of the PR, so ignore that part of my nit.
- The one piece worth keeping is the `result.reductionPercentage === 0` case. If compaction shrank nothing, the retry will hit the exact same wall, and the user just sees "auto-compacted (0% reduction). Retrying…" followed by another empty turn. So in that one case, skip the retry and fall straight through to the give-up message. Everywhere else, retry as you do now.

Capping at one retry (`MAX_COMPACT_RETRIES = 1`) is correct I think.

2, 3, 4: I can see all three in `d37463d` - the single `setMessages`, the divergence comment, and the `getLogger().warn` in the catch. All good, thank you.

One last thing on the tests, also non-blocking: they pass, but the new force-compact ones are a bit loose - they mostly assert `chatCallCount >= 3` and branch on `if (compactInfo) {...} else {...}`, so they'd still go green even if the compact path silently no-opped. Could you add one test that stubs the tokenizer / `compressMessages` and asserts the compact branch actually ran and recursed - e.g. the "auto-compacted" info message fired and `chat` was called again with the compacted messages? That pins the behaviour down so a future refactor can't quietly break it.

On your wider point about resilience - I agree. A local token count is always an approximation, funnily enough I pushed some changes last night to try and hone and make this more accurate but, the approximatation will just get smaller. Only the model truly knows its limit, so trying to recover before halting is the right default. I'm happy with this approach. Thanks again for the contribution and for being patient through the back-and-forth.

Issue #541, opened 2026-06-04
[Bug] execute_bash advertises a 2-minute timeout but BashExecutor has none (+ no process-group kill, unbounded output, AbortSignal ignored)


JimStenstrom:
## Description

The `execute_bash` tool advertises to the model that **"Commands time out after 2 minutes by default"** (`source/tools/execute-bash.tsx:70`), but `BashExecutor` implements **no timeout at all** — a long‑running command runs indefinitely. The same executor is missing several other process‑lifecycle controls: it spawns the shell without a process group (so cancellation orphans child processes), buffers stdout/stderr without any bound (an OOM vector), and ignores the AI SDK `AbortSignal` (so aborting a conversation or subagent never stops the shell).

These share one root cause — `BashExecutor` has no lifecycle management — and one fix area, so I'm filing them together; happy to split into focused PRs if maintainers prefer.

## Environment

- **OS**: Any (code‑level, platform‑independent)
- **Node version**: ≥ 22
- **Nanocoder version**: 1.27.0 (`main`)
- **Provider**: Any (provider‑agnostic)
- **Model**: Any model that calls `execute_bash`

## Steps to Reproduce

**A — No timeout (documented but unimplemented):**
1. In a non‑interactive run (`--plain`, headless, or a daemon‑triggered skill), have the agent run `execute_bash` with `sleep 600` (or `tail -f /dev/null`).
2. The promise from `BashExecutor.execute()` never resolves; nothing kills the process. The run hangs forever — there is no Escape to cancel in these modes.

**B — Orphaned child processes on cancel:**
1. Interactively run `execute_bash` with `sleep 600 | cat` (or `yes > /dev/null & sleep 600`).
2. Press Escape to cancel.
3. `ps` shows the `sleep`/`yes`/`cat` children still running — `cancel()` only `SIGTERM`s the `sh -c` parent.

**C — Unbounded output (OOM):**
1. Have the agent run `execute_bash` with `yes` or `cat /dev/urandom | base64`.
2. Node RSS climbs without bound until the process is OOM‑killed.

**D — AbortSignal ignored:**
1. A subagent or the main loop runs a long `execute_bash`, then the parent aborts (the SDK `AbortSignal` fires).
2. The shell keeps running — the tool drops `_options.abortSignal`.

## Expected Behavior

- Commands time out after the advertised default (~2 minutes, ideally configurable) and the process is killed when they do.
[...]

PR #547, opened 2026-06-09
fix: add timeout, output limits, and abort support

akramcodez:
## Description

Resolves the remaining valid findings from Issue #541 in the `execute_bash` tool and `BashExecutor`.

### Changes

* Added a default per-command timeout (`TIMEOUT_BASH_DEFAULT_MS = 120000`) so commands no longer run indefinitely.
* Added output buffering limits (`BASH_MAX_OUTPUT_BYTES = 5MB`) to prevent unbounded memory growth from commands that generate excessive output.
* Wired `AbortSignal` support from the tool layer through to `BashExecutor`, allowing running commands to be cancelled programmatically.
* Added regression tests covering timeout handling, output truncation, AbortSignal cancellation, and process-group cleanup behaviour.
* Verified that the orphaned child-process issue reported in #541 had already been resolved in a previous change through process-group termination.

## Type of Change

* [x] Bug fix
* [ ] New feature
* [ ] Breaking change
* [ ] Documentation update

### Automated Tests

* [x] New features include passing tests in `.spec.ts/tsx` files
* [x] All existing tests pass (`pnpm test:all` completes successfully)
* [x] Tests cover both success and error scenarios

### Manual Testing

* [x] Tested command timeout behaviour
* [x] Tested command cancellation via `AbortSignal`
* [x] Verified output truncation under large-output scenarios
* [ ] Tested MCP integration (if applicable)

## Checklist

* [x] Self-review completed
* [x] Documentation updated (if needed)
* [x] No breaking changes (or clearly documented)

### DIFF
--- a/source/constants.ts
+++ b/source/constants.ts
@@ -77,6 +77,8 @@ export const DELAY_COMMAND_COMPLETE_MS = 100;
 // === BASH EXECUTION ===
 export const INTERVAL_BASH_PROGRESS_MS = 500;
 export const BASH_OUTPUT_PREVIEW_LENGTH = 150;
+export const TIMEOUT_BASH_DEFAULT_MS = 120_000;
+export const BASH_MAX_OUTPUT_BYTES = 5 * 1024 * 1024;
 
 // === FILE SCANNER ===
 export const MAX_FILES_TO_SCAN = 1000;
--- a/source/services/bash-executor.spec.ts
+++ b/source/services/bash-executor.spec.ts
@@ -1,5 +1,5 @@
 import test from 'ava';
-import {BashExecutor} from './bash-executor';
+import { BashExecutor } from './bash-executor';
 
 console.log(`\nbash-executor.spec.ts`);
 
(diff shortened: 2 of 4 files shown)


will-lamerton:
Reviewed the executor changes - this is good to go. The `detached: true` + `process.kill(-pid, 'SIGTERM')` pair is the right primitive, the output-cap accounting fix in the follow-up commit is correct, and combined with the existing `handleCancel` path (which enumerates active executions and calls `bashExecutor.cancel()` on each), Escape now reliably terminates a long-running command and its whole child tree. Before this, `cancel()` only killed the `sh` wrapper and left children running, so this PR is the missing half that makes Escape->cancel actually work.

Verified all four scenarios from the issue against the modified executor:

- Timeout: `sleep 5` with `timeoutMs: 500` resolves in ~504ms with the timeout error.
- Process group: spawned `sh` -> `node setInterval`, captured the child PID, `cancel()`, child PID dead 500ms later.
- Output cap: `yes hello | head -c 6000000` truncates at ~5MB with the truncation marker appended once.
- AbortSignal: `sleep 5` with an aborted signal resolves in ~304ms with the cancel error.

`tsc --noEmit` and `biome lint` clean on the three changed files; the existing 29 executor specs pass.

One ask before merge: the new behavior has no regression tests - `bash-executor.spec.ts` is unchanged. Since the executor is the natural unit-test target, could you add specs for:

1. timeout resolves with the timeout error after `timeoutMs`
2. `signal` abort resolves with the cancel error
3. output cap trips and appends the truncation marker exactly once
4. `cancel()` on a detached process kills a spawned child (process-group assertion)

The four scenarios above translate almost directly into specs, so it should be a test-only change. Other than that, ready to merge! :)

will-lamerton:
Great job!

PR #555, opened 2026-06-10
fix: prevent silent overwrite of corrupted config files

### LINKED ISSUE #551: [Bug] Corrupted Config Files Can Be Silently Overwritten
## Description

If a configuration file contains invalid JSON, Nanocoder currently treats it as an empty configuration rather than surfacing a parsing error to the user.

This can lead to accidental data loss. A user who makes a simple syntax mistake (for example, a missing comma) may be taken through the setup flow as though no configuration exists. Completing the wizard can then overwrite the original configuration file, causing the user's existing setup to be lost.

## Environment

* **OS**: Any
* **Node version**: ≥ 22
* **Nanocoder version**: Current main
* **Provider**: Any
* **Model**: Any

## Steps to Reproduce

1. Create a valid provider or MCP configuration.
2. Introduce an invalid JSON syntax error into the configuration file.
3. Start Nanocoder.
4. Proceed through the configuration/setup flow.
5. Save the new configuration.

## Expected Behavior

Nanocoder should clearly inform the user that the configuration file is invalid and prevent accidental overwrites until the parsing issue is resolved.

## Actual Behavior

The invalid configuration is treated as though it does not exist. The user can proceed through setup and overwrite the original configuration file without being informed that a parsing error occurred.

## Additional Context

This does not break application startup, but it can create confusion and may result in accidental loss of existing configuration data.

Dhirenderchoudhary:
## Description

When a configuration file contained invalid JSON, the wizard silently treated it as empty and allowed the user to proceed through setup, potentially overwriting their existing (albeit broken) config.

The fix:
- Surfaces the JSON parse error to the user via the wizard's error banner, explaining that the file has invalid syntax.
- Tracks a configCorrupted flag that blocks writeConfigToDisk from silently overwriting the broken file.
- When the user opens the editor (Ctrl+E), the existing broken file is opened directly so they can fix the JSON syntax.
- When the editor saves valid JSON, or when the user deletes the config file, the corruption flag is cleared and normal operation resumes.

Closes #551

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

### Automated Tests

### Manual Testing

- [ ] Tested with Ollama
- [ ] Tested with OpenRouter

## Checklist

- [x] Self-review completed

### DIFF
--- a/source/wizards/base-config-wizard.spec.tsx
+++ b/source/wizards/base-config-wizard.spec.tsx
@@ -191,3 +191,82 @@ test.serial('displays an error when the configuration file contains invalid JSON
 		unlinkSync(configPath);
 	} catch {}
 });
+
+test.serial('blocks saving corrupted config', async t => {
+	const configFileName = `.bad-save-${Date.now()}.json`;
+	const configPath = join(process.cwd(), configFileName);
+	writeFileSync(configPath, '{ invalid }', 'utf-8');
+
+	const {lastFrame, stdin} = renderWithTheme(
+		<BaseConfigWizard<FakeItems>
+			title="Title"
+			focusId="corrupt-save-wizard"
+			configFileName={configFileName}
+			initialItems={{entries: []}}
+			parseConfig={() => {
+				throw new Error('Parse error');
+			}}
+			buildConfig={items => items}
+			hasItems={items => items.entries.length > 0}
+			renderConfigureStep={({onComplete}) => {
+				setTimeout(() => onComplete({entries: [{name: 'new'}]}), 10);
+				return <></>;
+			}}
+			renderSummaryItems={noopRenderSummary}
+			projectDir={process.cwd()}
+			onComplete={() => {}}
+		/>,
+	);
+
+	stdin.write('\r');
+	await new Promise(r => setTimeout(r, 100));
+
+	stdin.write('\r');
+	await new Promise(r => setTimeout(r, 50));
+
+	const output = lastFrame()!;
+	t.regex(output, /Cannot save: the existing configuration file contains invalid/);
+	
+	try {
+		unlinkSync(configPath);
+	} catch {}
+});
+
+test.serial('deleting corrupted config clears corruption state', async t => {
+	const configFileName = `.bad-del-${Date.now()}.json`;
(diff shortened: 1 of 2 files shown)


akramcodez (dismissed):
Thanks for the contribution! LGTM 👍

will-lamerton:
Thanks for tackling this @Dhirenderchoudhary - the `configCorrupted` flag is a clean fix - nice work.

One small follow-up: the three new tests in `base-config-wizard.spec.tsx` write real files into `process.cwd()` (`.bad-*.json`) and rely on the `unlinkSync` in the `try/catch` to clean up. If a test fails before the catch runs - e.g. an `t.regex` assertion fires after the file is already on disk - the dot-files leak into the working tree and risk being committed. Could you move the writes under `os.tmpdir()` in a unique subdir, or use `t.teardown(() => rmSync(configPath, { force: true }))` so cleanup is guaranteed regardless of how the test exits? Non-blocking, but it would make the suite a lot more robust. Other tests use this method if you need something to refer to.

Thanks :)

Dhirenderchoudhary:
@will-lamerton I've updated the tests.

 I moved the bad config file writes to unique subdirectories under `os.tmpdir()` and replaced the manual `try/catch` cleanup with `t.teardown(() => rmSync(...))`. 

While doing this, I also caught a bug in `BaseConfigWizard` where it was hardcoding `process.cwd()` instead of honoring the `projectDir` prop when resolving the project-level config path. That's fixed now too. All tests are green and pushed for review!


Thanks for the review

will-lamerton (approved):

will-lamerton:
> @will-lamerton I've updated the tests.
> 
> I moved the bad config file writes to unique subdirectories under `os.tmpdir()` and replaced the manual `try/catch` cleanup with `t.teardown(() => rmSync(...))`.
> 
> While doing this, I also caught a bug in `BaseConfigWizard` where it was hardcoding `process.cwd()` instead of honoring the `projectDir` prop when resolving the project-level config path. That's fixed now too. All tests are green and pushed for review!
> 
> Thanks for the review

Thank you very much! Merging 🔥

PR #557, opened 2026-06-10
fix: Move stop handler setup before starting resources and wrap lockfile write in try/catch that calls stop() on failure.

### LINKED ISSUE #549: [Bug] Daemon Startup Failure Leaks Watcher and IPC Resources
## Description

If daemon startup succeeds through watcher and IPC initialization but fails while writing the lockfile, previously created resources are not cleaned up.

## Environment

* **OS**: Any
* **Node version**: ≥ 22
* **Nanocoder version**: Current main
* **Provider**: Any
* **Model**: Any

## Steps to Reproduce

1. Start the daemon.
2. Allow `watcher.start()` and `ipcServer.start()` to succeed.
3. Cause `writeLockfile()` to throw (permissions, disk issues, etc.).

## Expected Behavior

All previously started resources should be cleaned up before startup fails.

## Actual Behavior

The startup process throws and exits without stopping:

* Chokidar FSWatcher
* IPC server/socket
* Associated daemon resources

## Root Cause

`startDaemon()` throws before returning a `DaemonHandle`. The shutdown handler that would normally call `handle.stop()` is never registered, and the top-level error handler exits immediately.

rakshith1928:
## Description

Fixes a resource leak in the daemon startup sequence. When `writeLockfile()` fails (due to permissions, disk issues, etc.) after the file watcher and IPC server have already started, the resources were not cleaned up because the shutdown handler was registered after the lockfile write operation.

Changes - `source/daemon/daemon.ts`: Moved `stop` function definition and `stopHandler.fn` assignment before `watcher.start()` and `ipcServer.start()`. Added try/catch around `writeLockfile()` that calls `await stop()` on failure before re-throwing.

Closes #549 

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

### Automated Tests

### Manual Testing

- [ ] Tested with Ollama
- [ ] Tested with OpenRouter

## Checklist

- [x] Self-review completed

### DIFF
--- a/source/daemon/daemon.ts
+++ b/source/daemon/daemon.ts
@@ -179,16 +179,6 @@ export async function startDaemon(opts: DaemonOptions): Promise<DaemonHandle> {
 		cron.register(sub.filter.cron);
 	}
 
-	await watcher.start();
-	await ipcServer.start();
-
-	await writeLockfile({
-		pid: process.pid,
-		socketPath: getSocketPath(opts.projectRoot),
-		startedAt: Date.now(),
-		projectRoot: opts.projectRoot,
-	});
-
 	let stopPromise: Promise<void> | null = null;
 	const stop = (): Promise<void> => {
 		// Idempotent: subsequent callers get the in-flight Promise so they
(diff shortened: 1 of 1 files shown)


akramcodez (approved):
Thanks for the contribution! I pulled the changes locally and tested the lockfile-failure path.

The fix correctly cleans up daemon resources when startup fails after watcher and IPC initialization, and it addresses the issue described in issue.

LGTM 👍

will-lamerton:
Looks good from my side @rakshith1928 !

Feel free to add yourself as a contributor to our website via a PR which I will approve :)

PR #579, opened 2026-06-18
fix: forward resolved context limit to Ollama

### LINKED ISSUE #485: [Feature]
## Description

Give nanocoder the ability to modify context tokens for specific models or sub agents

## Use Case

Using local models the context length is one key part for speed and not every request needs a full context length of a specific model. E.g. a subagent that should only Explore folders for file names may needs less context than a code reviewing or Unit test agent. 

## Proposed Solution

Set in the agents.config.json specific context values for specific models (like num_ctx for ollama) or that the context-max command also tries to set the context for the chosen model. 

## Alternatives Considered

## Additional Context

- [ x] I have searched existing issues to ensure this is not a duplicate
- [ x] This feature aligns with the project's goals (local-first AI assistance)

If you have thoughts on how this could be implemented, share them here

zerone0x:
## Description

This PR fixes the missing runtime half of #485 for Ollama-backed models.

Nanocoder already let users configure context limits via:
- `/context-max` / `--context-max`
- provider-level `contextWindow`
- per-model `contextWindows`
- subagent `contextWindow`

But those limits were only used inside Nanocoder for context accounting and auto-compact decisions. They were **not** being forwarded to Ollama itself, so the actual model runtime could still use a different context size.

This change now forwards the resolved Nanocoder context limit to Ollama as `options.num_ctx` on each request, so the configured limit affects the real runtime behavior too.

It also updates the docs to make that behavior explicit.

## Type of Change

- [ ] New feature
- [x] Bug fix
- [ ] Breaking change
- [x] Documentation update

### Automated Tests

Ran:
- `npm exec --yes pnpm -- test:ava source/ai-sdk-client/chat/provider-options.spec.ts`
- `npm exec --yes pnpm -- test:types`
- `npm exec --yes pnpm -- exec biome check source/ai-sdk-client/chat/provider-options.ts source/ai-sdk-client/chat/provider-options.spec.ts source/ai-sdk-client/chat/chat-handler.ts docs/configuration/providers/ollama.md docs/features/subagents.md`

### Manual Testing

- [ ] Tested with Ollama
- [ ] Tested with OpenRouter

Manual testing was not run in this environment.

## Checklist

- [x] Self-review completed

**Closes #485**

### DIFF
--- a/docs/configuration/providers/ollama.md
+++ b/docs/configuration/providers/ollama.md
@@ -34,6 +34,15 @@ By default, Ollama uses a 2048 token context window which is too small for agent
 OLLAMA_NUM_CTX=32768 ollama serve
 ```
 
+Nanocoder also forwards its resolved context limit to Ollama as `options.num_ctx` on each request. That means these settings flow through automatically:
+
+- `/context-max` or `--context-max`
+- provider-level `contextWindow`
+- per-model `contextWindows`
+- subagent `contextWindow`
+
+Use the Ollama server default (`OLLAMA_NUM_CTX`) as your baseline ceiling, then tighten individual Nanocoder sessions or models with the settings above when you want smaller/faster contexts.
+
 Or set it permanently in your environment.
 
 ### Signs of an Insufficient Context Limit
--- a/docs/features/subagents.md
+++ b/docs/features/subagents.md
@@ -125,7 +125,7 @@ Always use your tools — never guess.
 
 The `provider` must match a provider name configured in your `agents.config.json`.
 
-If you set `contextWindow`, Nanocoder creates that subagent with its own context limit override. This is useful when a lightweight research agent should run with a smaller local-model context than your main coding agent.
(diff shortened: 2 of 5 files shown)


akramcodez (commented):
Thanks for the contribution - I think the overall direction is correct, and forwarding the resolved context limit to Ollama is absolutely the right approach.

However, after validating the issue and tracing the context resolution flow, I don't think we can confidently close this issue yet.

The main blocker is the subagent precedence issue. The original feature request explicitly calls out subagents using different context limits, but the current resolution order means a subagent override is ignored whenever a per-model `contextWindows` entry exists. I was able to reproduce and verify this behavior locally.

Additionally, I wasn't able to verify whether Ollama's OpenAI-compatible endpoint actually honors the injected `options.num_ctx` field at runtime. Since the effectiveness of this fix depends on that behavior, I'd like to see either manual validation or automated coverage demonstrating that the runtime context limit changes when using the `/v1/chat/completions` endpoint.

Once the subagent override issue is addressed and the Ollama runtime behavior is verified, I'd be happy to take another look.

will-lamerton:
Hey @zerone0x - just checking in on this PR - would love to get it merged :)

Issue #575, opened 2026-06-15
[Feature] Navigable Slash Commands (Arrow Keys + Enter Selection)


rakshith1928:
## Description

Add keyboard navigation (up/down arrows) and Enter selection to the slash command completion menu.

## Use Case

Currently, typing / shows command completions but users must either type the full command name or press Tab only when there's exactly one match. 

## Proposed Solution

Files to modify:
1. source/hooks/useUIState.ts - Add selectedCompletionIndex state
2. source/components/user-input.tsx - Handle arrow keys + Enter, highlight selected item
Behavior:
- / typed → completions show, index 0 auto-selected
- ↓/↑ → navigate completions (wrap around), not history
- Enter (when completions visible) → select → fill /command in input, hide list
- Typing more chars → filter + reset to index 0
- Esc → clear input, hide completions, reset index

## Alternatives Considered

Mouse click support
Tab to cycle: Already used for file autocomplete; Enter is standard for selection

## Additional Context

- [x] I have searched existing issues to ensure this is not a duplicate
- [x] This feature aligns with the project's goals (local-first AI assistance)

## Implementation Notes (optional)

The foundation already exists — just needs wiring:
- Command registry already finds matches (/mod → /mode, /model, /model-database)
- UserInput already computes matches and shows the list
- UI state already shares the completion list between components
Only 2 small additions needed:
1. useUIState.ts — track which item is highlighted (index)
2. user-input.tsx — arrow keys move highlight, Enter fills the command
Tests needed in:
- source/hooks/useUIState.spec.ts (new state + reset)
- source/components/user-input.spec.tsx (navigation, selection, rendering) highlights not working

rakshith1928:
Can i work on this

PR #578, opened 2026-06-16
navigation with arrow keys for slash commands with highlighting

rakshith1928:
## Description
Add keyboard navigation (↑/↓) and Enter selection to slash command completions, matching modern CLI UX.
Also fixes a double-submit race condition when selecting completions(which i encountered after my 1st commit).
Closes #575 (maybe create a new issue for mouse click feature in future)

**Changes:**
- Type `/` → completions appear with first item auto-selected (▸ /command)
- ↑/↓ → navigate completions
- Enter → selects highlighted command → fills `/command` in input, hides completions
- Enter (again) → executes the filled command
- Tab → no-op when completions visible (avoids conflict with file autocomplete)
- Esc → clears input, hides completions, resets selection

Also The Completion type is defined in source/types/components.ts (line 18), not in types/index.ts. The original code had a wrong import that would fail type-checking.

 ## Bug Fix:
- Fixed double-submit race condition where pressing Enter to select a completion
  would also execute the stale partial command (e.g., `/cl` instead of `/clear`)
- Root cause: Two `useInput` handlers (UserInput + TextInput) both processing Enter
- Solution: Single-handler architecture - TextInput ignores Enter (`handleEnter=false`);
  UserInput handles completion selection then submit with guard flag preventing
  completion re-trigger

 ## Files changed:
 - source/hooks/useUIState.ts - added selectedCompletionIndex state
 - source/components/user-input.tsx - navigation, selection, highlighting ( selectedCommad uses colors.info others use --    colors.primary) single-handler Enter logic
 - source/components/text-input.tsx - added handleEnter/onEnter props
 - source/hooks/useUIState.spec.ts - state tests
 - source/components/text-input.spec.ts - handleEnter logic tests
 - source/components/user-input.spec.tsx - added tests to simulate real user input using stdin.write().

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

## Demo Video

https://github.com/user-attachments/assets/4180a34a-f151-4501-9883-dfc15fff8e1c

Future upgarde maybe adding mouse click option .
Check issue #575 

## Testing
- pnpm test:all passes
- pnpm test:ava source/hooks/useUIState.spec.ts - state tests
- pnpm test:ava source/components/text-input.spec.ts - handleEnter logic tests  
- pnpm test:ava source/components/user-input.spec.ts- integration tets

### Automated Tests

- [x] New features include passing tests in `.spec.ts/tsx` files
[...]

PR #580, opened 2026-06-18
feat: add keyboard selection for slash completions

miloquinn:
## Summary
- Track the selected slash-command completion in shared UI state
- Use Up/Down to move through visible slash completions with wraparound
- Use Enter to fill the selected command into the input
- Use Escape to dismiss the completion list without triggering the clear-input prompt
- Highlight the selected completion with the same marker pattern as file suggestions

## Why
The slash completion menu was display-only unless there was exactly one match. When multiple commands matched, users could not navigate the menu or select a highlighted command with Enter, even though the UI already showed a completion list.

This implements the keyboard-selection behavior requested in #575 while preserving existing file-autocomplete priority.

## Testing
- `npm run test:types`
- `npx ava source/hooks/useUIState.spec.ts source/components/user-input.spec.tsx --timeout=60s`
- `npm run test:format -- source/hooks/useUIState.ts source/hooks/useUIState.spec.ts source/components/user-input.tsx source/components/user-input.spec.tsx`
- `git diff --check`

Note: `test:format` reports existing Biome config schema/deprecation infos, but no formatting errors after this change.

### DIFF
--- a/source/components/user-input.tsx
+++ b/source/components/user-input.tsx
@@ -94,10 +94,12 @@ export default function UserInput({
 		showClearMessage,
 		showCompletions,
 		completions,
+		selectedCompletionIndex,
 		pendingFileMentions,
 		setShowClearMessage,
 		setShowCompletions,
 		setCompletions,
+		setSelectedCompletionIndex,
 		setPendingFileMentions,
 		resetUIState,
 	} = uiState;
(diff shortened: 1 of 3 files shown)


will-lamerton:
Hi @miloquinn - thanks so much for this PR - unfortunately the work was tackled for this issue in PR #578 and as this was opened and polished first, we've merged that one. Thank you so much for the work though - appreciate your contribution and if there's another issue you want to take on I'll happily assign you so there's no crossover :)

PR #592, opened 2026-06-21
fix(frontmatter): recognise empty frontmatter blocks

MFA-G:
## Summary

`splitFrontmatter` failed to recognise an **empty** frontmatter block — a file that opens with `---` and closes with `---` on the next line, with no lines in between:

```
---
---
Body content
```

The existing regex required at least one captured line between the delimiters (`([\s\S]*?)\r?\n---`), so an empty block fell through to the no-frontmatter branch. Two consequences:

- `hasFrontmatter` was reported as `false` and the literal `---\n---` markers leaked into `body`.
- For subagent markdown files, `extractRawFrontmatter` then threw `No YAML frontmatter found in file` (see `source/subagents/markdown-parser.ts`), even though the file *does* have a (valid, empty) frontmatter block.

## Fix

Make the inner frontmatter capture group optional so an empty block is matched, returning an empty `frontmatter` string and a clean `body`:

```diff
-const frontmatterRegex = /^---\s*\r?\n([\s\S]*?)\r?\n---\s*\r?\n?([\s\S]*)$/;
+const frontmatterRegex = /^---\s*\r?\n(?:([\s\S]*?)\r?\n)?---\s*\r?\n?([\s\S]*)$/;
```

I verified against a matrix of inputs that the **only** behavioural change is the previously-unhandled empty-block case — every other input (standard, multi-line, CRLF, no trailing newline, embedded `---` in the body, lone delimiter, no frontmatter) splits identically to before.

## Tests

Added `source/utils/frontmatter.spec.ts` (the module had no spec). It covers `splitFrontmatter` across all the cases above — including a regression test for the empty block — plus `parseYamlObject` (valid object, nested/array values, blank input, scalar, top-level array, invalid YAML).

## Validation

- `pnpm exec ava source/utils/frontmatter.spec.ts` — 15 passed
- `pnpm run test:ava` — full suite, 5775 passed
- `pnpm run test:types` — clean
- `pnpm exec biome lint/check` (changed files) — clean
- `pnpm run test:knip` — clean

### DIFF
--- a/source/subagents/markdown-parser.spec.ts
+++ b/source/subagents/markdown-parser.spec.ts
@@ -37,6 +37,43 @@ test.serial('extractFrontmatter - throws on missing frontmatter', t => {
 	);
 });
 
+test.serial(
+	'extractFrontmatter - empty frontmatter block surfaces a validation error',
+	t => {
+		// An empty `---\n---` block should be treated as an empty object so it
+		// flows through to schema validation, producing a clear "name is
+		// required" error rather than a misleading "must be an object" parse
+		// failure.
+		const content = `---
+---
+You are a subagent without metadata.`;
+
+		t.throws(() => extractFrontmatter(content), {
+			message: /name is required/,
+		});
+	},
+);
+
+test.serial(
+	'parseSubagentMarkdown - empty frontmatter block surfaces a validation error',
+	async t => {
+		const tmpPath = join(tmpdir(), 'empty-frontmatter-agent.md');
+		const content = `---
+---
+You are a subagent without metadata.`;
+
+		await writeFile(tmpPath, content, 'utf-8');
+
+		try {
+			await t.throwsAsync(() => parseSubagentMarkdown(tmpPath), {
+				message: /name is required/,
+			});
+		} finally {
+			await unlink(tmpPath);
+		}
+	},
+);
+
 test.serial('extractBody - extracts body content', t => {
 	const content = `---
 name: test
--- a/source/subagents/markdown-parser.ts
+++ b/source/subagents/markdown-parser.ts
@@ -18,10 +18,9 @@
  */
 
 import * as fs from 'node:fs/promises';
-import {parse as parseYaml} from 'yaml';
 import {parseSubscribeBlock} from '@/skills/parse-subscribe';
 import type {SkillTrigger} from '@/types/skills';
-import {splitFrontmatter} from '@/utils/frontmatter';
+import {parseYamlObject, splitFrontmatter} from '@/utils/frontmatter';
 import type {
 	ParsedSubagentFile,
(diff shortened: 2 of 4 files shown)


akramcodez (commented):
@MFA-G  Thanks for the contribution - this is a nice catch, and the regex change itself looks correct. The new test coverage is also excellent and does a great job covering both `splitFrontmatter` and `parseYamlObject`.

However, after tracing the full user flow from the original issue, I think there is still one remaining edge case before we can confidently close this.

The issue description mentions that subagent markdown files with an empty frontmatter block currently fail because they are treated as having no frontmatter:

This PR fixes that part by ensuring `hasFrontmatter` is now correctly reported as `true`.

However, `extractRawFrontmatter` in `source/subagents/markdown-parser.ts` still calls `parseYaml(raw)` directly. For an empty frontmatter block, `raw` becomes an empty string and `parseYaml('')` evaluates to `null`, which means the code will now throw:

```text
YAML frontmatter must be an object
```

As a result, the user still receives a parsing error before reaching the frontmatter validation stage.

**Recommendation:**

Please update `extractRawFrontmatter` in `source/subagents/markdown-parser.ts` to use the existing `parseYamlObject` utility rather than calling `parseYaml` directly.

`parseYamlObject('')` already returns `{}` (and is covered by the new tests), which would allow empty frontmatter to continue through validation and produce a much clearer error such as:

```text
Invalid frontmatter: name is required and must be a non-empty string
```

This feels closer to the intent of the original issue and fully resolves the user-facing workflow rather than only fixing the regex layer.

Once that's addressed, I'd be happy to approve.

MFA-G:
@akramcodez Thanks for tracing the full flow — you're right, fixing only the regex layer left `extractRawFrontmatter` throwing `YAML frontmatter must be an object` for an empty block before validation could run.

Addressed in 17d9e2c: `extractRawFrontmatter` now uses the shared `parseYamlObject` helper instead of calling `parseYaml` directly. `parseYamlObject('')` returns `{}`, so an empty frontmatter block flows through to schema validation and produces the clearer `Invalid frontmatter: name is required and must be a non-empty string` error, matching the intent of the original issue.

Added two tests covering the full user-facing path: `extractFrontmatter` and `parseSubagentMarkdown` with an empty `---\n---` block both now surface the `name is required` validation error. Typecheck, lint, and the parser/frontmatter test suites all pass.

akramcodez (approved):
@MFA-G Thanks for addressing the previous review. I've re-reviewed the latest changes and the remaining edge case is now handled correctly.

Nice work - I'm happy to approve and think this is ready to merge.

will-lamerton:
Thanks for this PR @MFA-G  - feel free to add yourself as a contributor to our website via a PR which I will approve :)

PR #596, opened 2026-06-23
fix: increase terminal max-width from 120 to 1024

### LINKED ISSUE #595: Hardcoded 120-column max width ignores large terminals
## Summary

`useTerminalWidth` caps content width to 120 columns regardless of terminal size, leaving most of a wide terminal empty.

## Location

`source/hooks/useTerminalWidth.ts` (compiled to `dist/hooks/useTerminalWidth.js`):

```ts
const calculateBoxWidth = (columns) => Math.max(Math.min(columns - 4, 120), 40);
```

Also `source/constants.ts` sets `DEFAULT_TERMINAL_WIDTH = 120` — unclear whether this was intended as a cap or just a fallback.

## Impact

On a 190-column terminal, usable width is clamped to ~116 (120 − 4 padding), wasting ~74 columns of horizontal space. Forces users to either:
- Resize their terminal smaller, or
- Patch the dist file after every `npm update`

## Suggested fix

- Use the full terminal width as the max (or make it configurable)
- At minimum remove the `Math.min(columns - 4, 120)` cap so it respects actual terminal width
- Or expose a setting in `nanocoder-preferences.json` / `agents.config.json` for max-width

## Workaround

```bash
sed -i "s/min(columns - 4, 120)/min(columns - 4, 999)/" \
  "$NVM_DIR/versions/node/$(ls -1 "$NVM_DIR/versions/node" | tail -1)/lib/node_modules/@nanocollective/nanocoder/dist/hooks/useTerminalWidth.js"
```

zhaider3c:
## Summary

Removes the hardcoded 120-column cap in `useTerminalWidth`, replacing it with 1024. This lets content fill wide terminal windows instead of leaving ~74 columns of empty space on a 190-col terminal.

## Changes

- `source/hooks/useTerminalWidth.tsx` — `Math.min(columns - 4, 120)` → `Math.min(columns - 4, 1024)`
- `source/constants.ts` — `DEFAULT_TERMINAL_WIDTH = 120` → `DEFAULT_TERMINAL_WIDTH = 1024`

## Related

Closes #595


akramcodez (commented):
@zhaider3c Thanks for the contribution! The implementation correctly updates the terminal width limits and addresses the original issue.

One thing still needs updating before this can be merged: the corresponding unit test still expects the previous maximum width (`120`). Since the implementation now uses `1024`, the test should be updated accordingly to reflect the new behavior and ensure the change is covered.

Once the test expectation is updated, I'd be happy to approve.

will-lamerton:
Thanks for tackling this - the diff is small, focused, and easy to review. That said, I agree with the framing in the linked issue that something should give on the 120 cap, but I'd push back on going all the way to 1024. A few thoughts:

**Why the cap exists**

The 120 ceiling isn't an accident or a forgotten constant - it's doing real work. CLI UX research and most well-designed TUIs converge on ~100–120 columns because:

- Long single-line content (diffs, table rows, code snippets wrapped in boxes) reads worse when stretched. The eye loses the line and re-finds it on every row.
- Side-by-side panels, two-column layouts, and aligned ASCII art only look balanced when the box itself has a known upper bound — otherwise a 300-col terminal gives you a full-width box on the left and a wasteland on the right.
- It caps pathological cases: a user pipes output into a non-TTY reporter that reports `columns: 99999`, or resizes the window very wide mid-session. Without an upper bound, transient frames can blow up `cli-table3` column math in `source/markdown-parser/table-parser.ts:51` and the file-ops previews.

So I don't think the right fix is "remove the cap." A configurable, sensible default is.

**The middle ground I'd suggest**

Something like `160` or `200` for the default - high enough that a 190-col terminal (the exact case from the issue) gets its full ~186 usable columns back, low enough that genuinely wide terminals don't produce comically stretched boxes. Concretely:

- `source/constants.ts:50` — `DEFAULT_TERMINAL_WIDTH = 160` (or `200`)
- `source/hooks/useTerminalWidth.tsx:8` — `Math.max(Math.min(columns - 4, 160), 40)`
- Better still: read it from `nanocoder-preferences.json` with the constant as the default, which is the third option the issue itself proposes. That sidesteps the "what's the right number" debate entirely.

Let me know your thoughts :)

PR #599, opened 2026-06-28
feat: queue user messages while busy

### LINKED ISSUE #597: [Feature] Queue user messages while the agent is busy, with arrow-key navigation to edit them
**Summary**
Allow the user to keep typing and submit additional messages while an agent turn (LLM streaming + tool execution) is in flight, instead of being locked out of the input. Submitted messages should be queued and dispatched in order when the current turn ends, and the user should be able to see the queue and navigate it with ↑/↓ to re-load a queued message back into the input for editing — modeled on the behaviour in Claude Code.

**Context**
Today the chat input is fully replaced by a "Press Esc to cancel" spinner the moment a turn starts:

- `source/app/sections/interactive-app.tsx:198` passes `inputDisabled={chatHandler.isGenerating || appState.isToolExecuting}` into `ChatInput`.
- `source/app/components/chat-input.tsx:158-163` only mounts `UserInput` when neither tool confirmation, tool execution, nor question prompts are pending AND the client is ready — i.e. never while busy.
- `source/components/user-input.tsx:618-630` is the "disabled" branch: it renders a spinner and an early-return, so even though `TextInput` is mounted, all keypresses (including typing) are blocked by `useInput` at `source/components/user-input.tsx:431-433` (`if (disabled) { return; }`).

This is the "user input disappears until the agent loop finishes" behaviour the request describes. There is no existing user-input queue to extend:

- `source/utils/message-queue.tsx` is only for system-level `logInfo`/`logError`/etc. messages, not user submissions. It pipes `ErrorMessage`/`InfoMessage`/`SuccessMessage`/`WarningMessage` into the static chat queue — nothing about preserving a pending user message across turns.
- `source/components/chat-queue.tsx` only renders already-finalised message components via Ink's `Static`; it has no concept of "next user message to send".
- The `onConversationComplete` callback fired from `source/hooks/chat-handler/conversation/conversation-loop.tsx:367-369, 670-672, 929-931` (wired up at `source/app/App.tsx:178-185` to reset `isConversationComplete`/tool counts/task list) is the natural drain point for a queue — once a turn ends, the next queued message can be dispatched.

Existing patterns to follow:

- The two existing "global handler slot" queues (`source/utils/question-queue.ts`, `source/utils/tool-approval-queue.ts`, `source/utils/tool-confirm-queue.ts`) and their shared helper `source/utils/global-handler-slot.ts` show how the codebase bridges a tool's suspended Promise with the Ink UI via a module-level singleton + a `set…` wiring call from `App.tsx`. The new user-input queue can be a lighter-weight version of this: state lives in a hook/ref owned by `App.tsx`, exposed to `UserInput` via props.
- `source/components/user-input.tsx:330-389` already handles `assemblePrompt` + `extractImageReferences` + `promptHistory.addPrompt` inside `handleSubmit`. The submit path can branch on `isBusy`: instead of calling `onSubmit` directly, push onto the queue and reset the input.
- Arrow-key history navigation already exists in `source/components/user-input.tsx:391-465` (`handleHistoryNavigation`) and is wired through `useInputState` / `promptHistory`. The new queue-navigation mode would be a sibling state machine: ↑/↓ move a "selected queue index" cursor, Enter loads the selected message back into the input for editing, and another key (e.g. Ctrl+Delete) removes it from the queue.

**Proposed approach**
- Introduce a `useMessageQueue` (or extend `useAppState` / `useChatHandler`) that owns an ordered list of pending user messages, each carrying the same triple the current submit path uses: `{message, displayValue, images}`. Expose `queuedMessages`, `enqueueMessage`, `dequeueMessage`, `removeFromQueue`, and `updateQueueItem`.
- Lift the queue state into `App.tsx` (next to `appState.chatComponents`) so it can be read by both `InteractiveApp` (for rendering the queue list above/below the input) and `useChatHandler` (for draining on completion). Wire the drain by adding a `drainMessageQueue` call inside the `onConversationComplete` callback at `source/app/App.tsx:178-185` and inside the error/`finally` path in `source/hooks/chat-handler/useChatHandler.tsx:200-208`.
- Replace the hard `inputDisabled` swap in `source/app/components/chat-input.tsx` with a softer "input stays enabled while busy" path: render `UserInput` whenever the client/MCP are ready and no decision-state modal owns the slot (tool confirmation, question prompt, subagent approval), and pass `isBusy={chatHandler.isGenerating || appState.isToolExecuting}` instead of `inputDisabled`.
- Modify `source/components/user-input.tsx`:
  - When `isBusy` is true, `handleSubmit` pushes onto the queue and clears the input instead of calling `onSubmit`. Re-show the queue strip below the input.
  - Add a queue-navigation state machine similar to `handleHistoryNavigation`: ↑/↓ move a selected index over the queue, Enter/Esc re-loads the selected message into the input buffer (re-assembling its `InputState` via `setInputState`), Ctrl+Delete removes the selected item, Escape-with-empty-input clears the selection.
  - Render the queue strip (e.g. `▸ 1. first line… 2. second line…` with the active row highlighted) above or just below the input box, matching the position of the existing `showCompletions` / `isFileAutocompleteMode` lists in `source/components/user-input.tsx:649-712`.
  - Keep the existing "Press Esc to cancel" indicator but only as a small status line; the input itself stays interactive.
- In `source/app/sections/interactive-app.tsx`, stop wiring `inputDisabled`. Pass `isBusy` and `queuedMessages` into `ChatInput` → `UserInput` so the queue strip renders next to the spinner instead of replacing the input.
- Update tests:
  - `source/components/user-input.spec.tsx` and `source/app/components/chat-input.spec.tsx` for the new submit-while-busy path, queue rendering, and arrow-key navigation.
  - Add a hook-level test that `onConversationComplete` drains the queue in FIFO order and that an error in the current turn still drains correctly.

**Acceptance criteria**
- While `chatHandler.isGenerating` or `appState.isToolExecuting` is true, the text input remains mounted, accepts typed characters, and shows a visible queue strip of pending submissions.
- Submitting (Enter) while busy adds the message to the end of the queue and clears the input; the message is **not** sent to the LLM until the in-flight turn ends.
- When the in-flight turn ends (via `onConversationComplete` or the error/`finally` path in `useChatHandler`), the next queued message is dispatched automatically in FIFO order, and the queue strip updates.
- ↑/↓ arrows navigate the queue strip when the input is empty (or via a dedicated modifier key) and load the selected message back into the input buffer for editing; a separate key (e.g. Ctrl+Delete) removes a queued message.
- Image attachments and `[@file]` placeholders are preserved through enqueue/dequeue (i.e. the queue stores the same `{message, displayValue, images}` triple the submit path already produces).
- An empty queue behaves identically to today: pressing Enter sends immediately; Escape cancels the running turn.
- No regressions in the existing decision states — tool confirmation, question prompts, and subagent approval still own the slot when they are pending (the input is hidden during those, just like today).

**Out of scope**
- Cross-session queue persistence (queue lives only in memory for the current run).
- Reordering queued messages via drag/move (only delete + re-type is required for the first cut).
- Showing the queue count in the status/title bar (can be a follow-up).
- Changing the behaviour of `/commands` while busy (slash commands should still bypass the queue and execute immediately, mirroring today's behaviour).

Dhirenderchoudhary:
## Description

  Adds an in-memory queued user-message flow while the agent is busy. The chat input stays interactive during LLM streaming/tool execution,
  submitted messages are queued FIFO, queued items can be selected with ↑/↓, edited with Enter, and removed with Ctrl+Delete.

  Fixes #597

  ## Type of Change

  - [ ] Bug fix
  - [x] New feature
  - [ ] Breaking change
  - [ ] Documentation update

  ## Testing

  ### Automated Tests

  Notes:
  - `npm run build` passed
  - `npm run test:format` passed
  - `npm run test:lint` passed
  - `npm run test:knip` passed
  - Changed-area AVA passed: 69 tests
  - `node_modules/.bin/tsc --noEmit` passed
  - Full fail-fast AVA reached 406 passing tests, then stopped on existing timeout in `source/auth/chatgpt-codex.spec.ts`

  ### Manual Testing

  - [ ] Tested with Ollama
  - [ ] Tested with OpenRouter

  ## Checklist

  - [x] Self-review completed

### DIFF
--- a/source/app/App.tsx
+++ b/source/app/App.tsx
@@ -38,8 +38,10 @@ import {useSessionAutosave} from '@/hooks/useSessionAutosave';
 import {ThemeContext} from '@/hooks/useTheme';
 import {TitleShapeContext, updateTitleShape} from '@/hooks/useTitleShape';
 import {UIStateProvider} from '@/hooks/useUIState';
+import {useUserMessageQueue} from '@/hooks/useUserMessageQueue';
 import {useVSCodeServer} from '@/hooks/useVSCodeServer';
 import {generateKey} from '@/session/key-generator';
+import type {ImageAttachment} from '@/types/core';
 import type {ThemePreset} from '@/types/ui';
 import {createPinoLogger} from '@/utils/logging/pino-logger';
 import {setGlobalMessageQueue} from '@/utils/message-queue';
@@ -75,6 +77,15 @@ export default function App({
 
 	// Use extracted hooks
 	const appState = useAppState(initialDevelopmentMode);
+	const userMessageQueue = useUserMessageQueue();
+	const queuedUserSubmitRef = React.useRef<
+		| ((
+				message: string,
+				displayValue: string,
+				images?: ImageAttachment[],
+		  ) => Promise<void>)
+		| null
+	>(null);
 	const {exit} = useApp();
 	const {isTrusted, handleConfirmTrust, isTrustLoading, isTrustedError} =
 		useDirectoryTrust();
@@ -159,6 +170,28 @@ export default function App({
 		}
 	}, []);
 
+	const drainQueuedUserMessage = React.useCallback(() => {
+		queueMicrotask(() => {
+			void userMessageQueue.drainNextMessage(async message => {
+				const submitQueuedMessage = queuedUserSubmitRef.current;
+				if (!submitQueuedMessage || !appState.client || !appState.toolManager) {
+					return false;
+				}
+
+				await submitQueuedMessage(
+					message.message,
+					message.displayValue,
+					message.images,
+				);
+				return true;
+			});
+		});
+	}, [
+		appState.client,
+		appState.toolManager,
+		userMessageQueue.drainNextMessage,
+	]);
+
 	// Setup chat handler
 	const chatHandler = useChatHandler({
 		client: appState.client,
@@ -180,6 +213,7 @@ export default function App({
 			appState.setCompactToolCounts(null);
 			appState.compactToolCountsRef.current = {};
 			appState.setLiveTaskList(null);
+			drainQueuedUserMessage();
 		},
 		reasoningExpandedRef: appState.reasoningExpandedRef,
 		compactToolDisplayRef: appState.compactToolDisplayRef,
@@ -404,6 +438,10 @@ export default function App({
 		activeEditor: vscodeServer.activeEditor,
 	});
 
+	React.useEffect(() => {
+		queuedUserSubmitRef.current = handleUserSubmit;
+	}, [handleUserSubmit]);
+
 	// Setup non-interactive mode
 	const {nonInteractiveLoadingMessage} = useNonInteractiveMode({
 		nonInteractivePrompt,
@@ -605,6 +643,7 @@ export default function App({
 						handleToolConfirmation={handleToolConfirmation}
 						handleQuestionAnswer={handleQuestionAnswer}
 						handleUserSubmit={handleUserSubmit}
+						userMessageQueue={userMessageQueue}
 						handleIdeSelect={handleIdeSelect}
 					/>
 				</UIStateProvider>
--- a/source/app/components/chat-input.spec.tsx
+++ b/source/app/components/chat-input.spec.tsx
@@ -102,6 +102,48 @@ test('ChatInput shows tool execution indicator when executing', t => {
 	unmount();
 });
 
+test('ChatInput keeps UserInput visible while a tool is executing', t => {
+	const mockToolCall = {
+		id: 'test-1',
+		function: {name: 'test_tool', arguments: {}},
+	};
+
+	const props = createDefaultProps({
+		isToolExecuting: true,
+		isBusy: true,
+		inputDisabled: true,
+		pendingToolCalls: [mockToolCall],
+		currentToolIndex: 0,
+	});
+
+	const {lastFrame, unmount} = renderWithTheme(<ChatInput {...props} />);
+	const output = lastFrame();
+	t.truthy(output);
+	t.regex(output!, /hat would you like me to help with\?/);
+	t.regex(output!, /Press Esc to cancel/);
+	unmount();
+});
+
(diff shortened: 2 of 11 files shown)


akramcodez (commented):
Thank you for tackling this issue! The queue UI, the `useUserMessageQueue` hook, and the arrow key navigation look great and are working as intended. The UI tests in `user-input.spec.tsx` also cover the keyboard navigation well.

However, comparing the PR against the issue requirements and our Review Golden Rules, there are a couple of missing pieces that need to be addressed before we can merge this confidently.

### 1. Missing Error/Finally Drain Path

The queue is currently drained inside `onConversationComplete` in `App.tsx`, but it won't be drained if an error occurs before the conversation loop starts (for example, during prompt building). In that case, `handleChatMessage` exits early, `onConversationComplete` is never called, and any queued messages remain stuck until the user manually starts another turn. Since the issue explicitly requires queued messages to continue after failures, the queue drain should also be triggered from the `catch`/`finally` path in `useChatHandler`. A good approach would be to pass a `drainQueuedUserMessage` callback (or the queue itself) into `useChatHandler` so it can safely drain the queue even when the current turn fails early.

### 2. Missing Hook-Level Tests for the Drain Behavior

The PR is also missing the hook-level tests requested in the issue. There should be a test verifying that queued messages are drained in FIFO order after a successful conversation, and another confirming that the queue is still drained when the current turn exits through the error path. Without these tests, this critical behavior isn't protected against future regressions.

### 3. Out-of-date Branch

The branch is also a little behind `main` and is missing several recent fixes. While Git will likely merge it without issues, rebasing onto the latest `main` before merging is recommended to avoid any unexpected conflicts or regressions with recent changes, such as the recent network error handling updates in `error-parser.ts`.

Once the error-path draining and the missing hook-level tests are addressed, this should be good to merge.

will-lamerton (changes requested):
Hey @Dhirenderchoudhary - a really cool PR. Thank you. One issue I am encountering is any queued messages double send, I'm not sure why. Can you take a look?

Image

will-lamerton (approved):
Nice job! All working and looking good!

PR #611, opened 2026-06-29
feat: add retry command for last user turn

### LINKED ISSUE #608: [Feature] Add `/retry` command to re-run the last user turn (optionally with a different model)
**Summary**
Add a new `/retry` slash command that re-submits the last user turn against the current model (or, with `--model <id>`, against a chosen model). It is genuinely missing from the family of "act on the last turn" commands — `/copy` reads the last assistant message and checkpoint-restore rolls back history, but neither re-runs the prompt.

**Context**
The extension path is documented in `CLAUDE.md` ("Command System" section): create the file under `source/commands/` exporting a `Command` object and add an entry to `lazyCommands` in `source/commands/lazy-registry.ts`. The neighboring commands set the conventions for this work:

- `/copy` (`source/commands/copy.ts:1`) is a pure handler with `(args, messages, metadata)` — but it only *reads* the last assistant message, it doesn't re-trigger a turn.
- `/resume` (`source/commands/resume.ts:1`, dispatched via `source/app/utils/handlers/session-handler.ts:23` `handleResumeCommand`, hooked at `source/app/utils/app-util.ts:550`) is the established "stub + intercepted handler" pattern for commands that need app state.
- `/model` and `/context-max` use `createStubCommand` (`source/commands/create-stub-command.ts:1`) and are intercepted in `handleSpecialCommand` at `source/app/utils/app-util.ts:233`, dispatched via the `SPECIAL_COMMANDS` table at `source/app/utils/app-util.ts:41`.

`/retry` cannot live as a pure `Command.handler`. To re-run a turn it needs `appHandlers.handleMessageSubmit` (defined at `source/hooks/useAppHandlers.tsx:521`), and to honor `--model` it needs the model-switching path in `source/hooks/useModeHandlers.tsx:98` `handleModelSelect(selectedProvider, selectedModel)` — both are app-level state, so the command must follow the same stub + intercepted-handler pattern as `/resume`, `/model`, and `/context-max`.

The "find the last user message" idiom already exists in the codebase: `source/hooks/useSessionAutosave.ts:134` filters by `role === 'user'` and takes the trailing entry, and `source/commands/copy.ts:12` does the same for assistant — so the lookup itself is a one-liner with an established shape.

**Proposed approach**
- Add a new file `source/commands/retry.ts` that exports a stub `Command` (using `createStubCommand` from `source/commands/create-stub-command.ts`), mirroring `source/commands/resume.ts:1`. Description: `Re-run the last user turn (use --model <id> to switch models first)`.
- Register it in `source/commands/lazy-registry.ts` alongside the other `LazyCommand` entries (alphabetical/logical placement near `resume` is natural).
- Add `RETRY: 'retry'` to the `SPECIAL_COMMANDS` table at `source/app/utils/app-util.ts:41` and dispatch it from the existing handler pipeline. A new `handleRetryCommand(commandParts, options)` is the cleanest fit — modeled on `handleResumeCommand` in `source/app/utils/handlers/session-handler.ts:23` — and should be invoked from the `processCommand` switch in `source/app/utils/app-util.ts:542` (right after `handleResumeCommand`).
- The handler should:
  1. Look up the last `role === 'user'` message from `options.messages` (same pattern as `source/hooks/useSessionAutosave.ts:134`); error with a friendly `errorMsg` if there is none.
  2. Parse args for `--model <id>` (and optionally `--provider <id>` to cross providers). Reuse `parseCustomCommandArgs` from `source/app/utils/app-util.ts:65` so quoted model ids work.
  3. If a model was specified, call into the model-select path: same-provider case uses `client.setModel` + `setCurrentModel` (mirroring `source/hooks/useModeHandlers.tsx:110`); cross-provider case uses `createLLMClient(provider, model)` and replaces the client (mirroring `source/hooks/useModeHandlers.tsx:140`). The handler will need these as new `MessageSubmissionOptions` callbacks (`onSwitchModel`, or by exposing `handleModelSelect`).
  4. Await `options.handleMessageSubmit(content, content)` with the user message's content (display and images can be omitted — a retry re-uses the same prompt text and intentionally drops any images from the original turn, matching the principle that `/retry` is text-only; revisit if users want image fidelity).
  5. `onCommandComplete?.()` to release the input.
- Add a spec file `source/commands/retry.spec.tsx` (or `retry.spec.ts`) covering: no prior user turn → error, prior turn → re-submits content, `--model` arg → model switches before re-submit, `--model` with explicit `--provider` → switches provider+model. The `MessageSubmissionOptions` callbacks can be stubbed the way `source/commands/copy.spec.ts` and `source/commands/resume` tests already do.

**Acceptance criteria**
- Typing `/retry` in an empty conversation surfaces a clear error (no user turn to retry) and does not crash.
- Typing `/retry` after at least one user turn re-submits that turn's text via the existing chat-submit pipeline, producing a fresh assistant response against the current model and provider.
- Typing `/retry --model <model-id>` (same provider) switches the model on the live client (no new client spin-up) and then re-submits.
- Typing `/retry --model <model-id> --provider <provider-id>` (cross-provider) builds a new client via `createLLMClient`, updates `currentProvider`/`currentModel`, and then re-submits.
- `/retry` appears in the slash-command picker and in `/help` with its description, using the lazy-load mechanism in `source/commands/lazy-registry.ts` (no eager import).
- Quoted model ids (e.g. `/retry --model "some model id"`) parse correctly via `parseCustomCommandArgs`.
- A new spec file passes; the existing `pnpm run test:all` suite still passes.

**Out of scope**
- Editing the retried prompt in place (e.g. `/retry <new text>`). The v1 contract is "re-run the last turn verbatim"; a follow-up could add prompt mutation, but it complicates UX and is a separate decision.
- Retrying with a different provider but the *same* model id (resolver would still apply; handled by the same `createLLMClient` path, but no special validation is added in v1).
- Re-attaching images from the original turn (turns are text-only on retry by default per the approach above).
- UI affordances beyond the slash command — no keybinding, no menu entry. A follow-up could add a `Shift+Ctrl+R` binding in `source/app/components/chat-input.tsx` if there's demand.

Dhirenderchoudhary:
## Description

Adds a new `/retry` command to re-run the last user turn, optionally with revised text. The command preserves the existing conversation flow and avoids duplicating unrelated state or unsafe behavior.

closes #608 
## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

### Automated Tests

### Manual Testing

- [x]Tested with Ollama
- [ ] Tested with OpenRouter

## Checklist

- [x] Self-review completed
- [x] No breaking changes

No new logging was required because `/retry` reuses the existing user-turn handling flow.

### DIFF
--- a/docs/features/commands.md
+++ b/docs/features/commands.md
@@ -37,6 +37,7 @@ Type `/` in the chat input to see available commands. All commands start with `/
 | `/schedule` | Read-only view of cron subscriptions declared by skills (see [Skills → Event subscriptions](skills.md#event-subscriptions)) |
 | `/skills` | List and inspect loaded skills; scaffold new bundle skills with AI assistance (see [Skills](skills.md)) |
 | `/resume` | Resume a previous chat session (aliases: `/sessions`, `/history`). See [Session Management](session-management.md) |
+| `/retry` | Re-run the last user turn. Use `/retry --model <id>` or `/retry --provider <name> --model <id>` to switch models first |
 | `/rename` | Rename the current session. Name must be non-empty and 100 characters or less. See [Session Management](session-management.md) |
 | `/explorer` | Interactive file browser to navigate, preview, and select files for context |
 | `/tune` | Configure runtime model behaviour — tool profiles, compaction, native tools, model parameters (see [Tune](tune.md)) |
--- a/source/app/App.tsx
+++ b/source/app/App.tsx
@@ -385,6 +385,7 @@ export default function App({
 		enterExplorerMode: modeHandlers.enterExplorerMode,
 		enterIdeSelectionMode: modeHandlers.enterIdeSelectionMode,
 		enterTune: modeHandlers.enterTune,
+		handleModelSelect: modeHandlers.handleModelSelect,
 		handleChatMessage: chatHandler.handleChatMessage,
 		dismissActiveEditor: vscodeServer.dismissActiveEditor,
 	});
--- a/source/app/components/modal-selectors.tsx
+++ b/source/app/components/modal-selectors.tsx
@@ -24,7 +24,7 @@ export interface ModalSelectorsProps {
 	} | null;
 
 	// Handlers - Model Selection
-	onModelSelect: (provider: string, model: string) => Promise<void>;
+	onModelSelect: (provider: string, model: string) => Promise<unknown>;
 	onModelSelectionCancel: () => void;
 
 	// Handlers - Model Database
--- a/source/app/utils/app-util.spec.ts
+++ b/source/app/utils/app-util.spec.ts
@@ -6,6 +6,7 @@ import {
 	parseContextLimit,
 	parseCustomCommandArgs,
 } from './app-util.js';
+import {lazyCommands} from '@/commands/lazy-registry';
 import type {MessageSubmissionOptions} from '@/types/index';
 import type {Session} from '@/session/session-manager';
 import {sessionManager} from '@/session/session-manager';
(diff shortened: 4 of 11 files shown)


akramcodez (dismissed):
Thank you for this PR! The implementation is clean, well-structured, and matches the issue requirements. I especially like the decision to wire `/retry` alongside `/resume` and the type-safe handling around `handleModelSelect`.

The test coverage is comprehensive, covering both success and failure scenarios along with the relevant edge cases. Everything looks solid, approved and ready to merge!

will-lamerton (changes requested):
Hey @Dhirenderchoudhary - solid PR. `/retry` is wired end-to-end: stub command, lazy registration, dispatched handler, model switch, re-submit, friendly error path, and a thorough test set. A few small notes below:

## Suggestions

- **Drop the unrelated `interactive-app.tsx:149-151` wrapper.** This change replaces `onModelSelect={modeHandlers.handleModelSelect}` with an `async (provider, model) => { await modeHandlers.handleModelSelect(provider, model); }` lambda. It has no functional effect (the return type `Promise<boolean>` is structurally compatible with `Promise<void>`) and isn't called out in the PR description. Either revert it or split it into a separate "chore" commit so the `/retry` diff stays scoped.
- **Consider moving the dispatch into `handleSpecialCommand`.** The issue's "Proposed approach" suggested adding `RETRY: 'retry'` to the `SPECIAL_COMMANDS` table and dispatching from `handleSpecialCommand` (alongside `/model`, `/context-max`). You went the `/resume` route (direct call from `handleSlashCommand`) — also fine, just a noted deviation. If you keep the current placement, a short comment near `app-util.ts:551` pointing to the `/retry` and `/resume` pair being the "handlers-with-args" subgroup would help future readers.
- **The `useModeHandlers.tsx:203-205` `finally { exitMode() }` change is a strict improvement** (previously `exitMode()` was outside the `try` for the cross-provider path, so a thrown error from `addToChatQueue` inside the success branch would have leaked the mode), but it also adds an extra `exitMode()` call path in the same-provider branches where `exitMode()` is already called before `return`. That's a no-op in practice (presumably idempotent), but worth a glance during code review to make sure the underlying `setMode(null)` is indeed idempotent. (I didn't see anything suggesting it isn't.)
- **Whitespace-only last user message is silently skipped** (`retry-handler.ts:18` — `message.content.trim()`). Probably the right call (retrying an empty prompt would be confusing), but if you'd rather be explicit, queue an info message and bail. Not blocking.
- Missing docs for the new command.

Nice work - happy to merge once the small `interactive-app.tsx` wrapper is reverted or extracted and the command is documented :)

PR #615, opened 2026-06-30
fix: patch malformed SSE termination string from providers

### LINKED ISSUE #614: [Bug]: "Unexpected token 'D'" crash with OpenAI-compatible providers sending double spaces in data:  [DONE]
## Description

Nanocoder crashes with an `Unexpected token 'D', " [DONE]" is not valid JSON` error when using certain OpenAI-compatible APIs (like Atlas Cloud). This happens because these providers send `data:  [DONE]` (with two spaces) instead of the standard `data: [DONE]`. The strict `@ai-sdk/openai-compatible` parser strips only one space, leaving `" [DONE]"`, which fails strict string comparison and falls through to `JSON.parse()`.

## Environment

- **OS**: Linux
- **Node version**: (insert your node version here)
- **Nanocoder version**: 1.28.0
- **Provider**: OpenAI-compatible API (Atlas Cloud)
- **Model**: `deepseek-ai/deepseek-r1-0528`

## Steps to Reproduce

1. Configure an OpenAI-compatible provider in `agents.config.json` that sends `data:  [DONE]` at the end of SSE streams (e.g., Atlas Cloud).
2. Start a chat session using this provider and any model.
3. Wait for the model to finish generating the response.

## Expected Behavior

The stream should complete successfully and the response should be processed.

## Actual Behavior

Nanocoder crashes at the end of the stream with the error `Unexpected token 'D', " [DONE]" is not valid JSON` and silently attempts to retry the request 4 times before failing.

## Logs/Screenshots

Image

akramcodez:
## Description

Closes #614.

This PR fixes a bug where Nanocoder crashes with an `Unexpected token 'D', " [DONE]" is not valid JSON` error at the end of streams when using certain OpenAI-compatible APIs (like Atlas Cloud or DeepSeek endpoints). 

These providers send a slightly malformed termination string: `data:  [DONE]` (with two spaces). Because the Vercel `@ai-sdk/openai-compatible` parser strips only one space and does a strict string comparison, it fails to recognize the `[DONE]` signal. It then falls back to `JSON.parse(" [DONE]")`, which immediately crashes the application.

**Changes:**
- Added a `TransformStream` interceptor to `createUndiciFetch` that normalizes multiple spaces in SSE streams (replacing `/data:\s+\[DONE\]/g` with `data: [DONE]`).
- Added a unit test in `provider-factory.spec.ts` using `MockAgent` to ensure the double-space termination string is patched correctly before reaching the AI SDK.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

### Automated Tests

### Manual Testing

- [ ] Tested with Ollama
- [ ] Tested with OpenRouter

## Checklist

- [x] Self-review completed

### ScreenShot
image

### DIFF
--- a/source/ai-sdk-client/providers/provider-factory.spec.ts
+++ b/source/ai-sdk-client/providers/provider-factory.spec.ts
@@ -3,8 +3,8 @@ import * as fs from 'node:fs';
 import * as os from 'node:os';
 import * as path from 'node:path';
 import type {AIProviderConfig} from '@/types/index';
-import {Agent} from 'undici';
-import {createProvider} from './provider-factory.js';
+import {Agent, MockAgent} from 'undici';
+import {createProvider, createUndiciFetch} from './provider-factory.js';
 
 test('createProvider creates provider with basic config', async t => {
 	const config: AIProviderConfig = {
@@ -397,4 +397,25 @@ test.serial('createProvider throws when github-copilot has no stored credential'
 		}
 		fs.rmSync(tmpDir, {recursive: true, force: true});
 	}
+});
+
+test('createUndiciFetch patches double spaces in SSE data: [DONE]', async t => {
+	const mockAgent = new MockAgent();
+	mockAgent.disableNetConnect();
+	
+	const mockPool = mockAgent.get('https://api.atlascloud.ai');
+	mockPool.intercept({
+		path: '/v1/chat/completions',
+		method: 'POST'
+	}).reply(200, 'data:  [DONE]', {
+		headers: { 'content-type': 'text/event-stream' }
+	});
+
+	const fetchFn = createUndiciFetch(mockAgent as unknown as Agent);
+	const response = await fetchFn('https://api.atlascloud.ai/v1/chat/completions', {
+		method: 'POST'
+	});
+	
+	const text = await response.text();
+	t.is(text, 'data: [DONE]');
 });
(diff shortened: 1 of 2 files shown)


will-lamerton:
Hey @akramcodez - the changes correctly address an issue where certain SSE responses containing double spaces in `[DONE]` messages break parsing, but further verification is needed to determine if this is isolated to the Atlas Cloud provider.

This PR introduces a `TransformStream` in `createUndiciFetch` at `source/ai-sdk-client/providers/provider-factory.ts:74` that specifically targets and repairs the `data: [DONE]` stream format.

### Required changes
None.

### Issue coverage
The PR addresses the symptom observed in the issue (parser failure due to malformed SSE messages). However, the underlying cause needs confirmation:

- **Is this Atlas-specific?** The test case in `source/ai-sdk-client/providers/provider-factory.spec.ts:413` specifically targets `api.atlascloud.ai`. You mentioned testing Atlas Cloud and it working fine; it is possible this only affects certain models, legacy API versions, or specific load-balancer configurations within the Atlas infrastructure.
- **Robustness:** The current implementation in `source/ai-sdk-client/providers/provider-factory.ts:74` performs a simple `str.includes('data: [DONE]')` check and a regex replacement on every chunk. While functional for the described issue, ensure that this transformation does not inadvertently corrupt chunks that might split a multi-byte character or the target string across two separate stream chunks.

### Suggestions
- **Regex refinement:** In `source/ai-sdk-client/providers/provider-factory.ts:80`, you are replacing `data:\s+\[DONE\]` with `data: [DONE]`. This is good, but if the intention is only to fix the double space, consider narrowing the regex to strictly target `data:\s\s\[DONE\]` to avoid unintended modifications to otherwise valid (though unconventional) formatting.
- **Broaden testing:** Since you are interested in whether this is provider-wide, it would be beneficial to run a quick regression test against other providers (e.g., Anthropic, OpenAI) using this modified fetch to ensure this transformation does not cause side effects for them.tream` is the correct, memory-efficient way to handle this on-the-fly streaming transformation.

If you wouldn't mind researching and coming back to me on the above that would be great!

PR #616, opened 2026-06-30
feat: add diff_edit tool for nano profile models (#604)

### LINKED ISSUE #604: [Feature] Add diff-format edit path for very weak local models
**Summary**
Add a search/replace-block (and/or unified-diff) edit tool, modeled on Aider's edit formats, selectable for models too weak to use the existing XML/JSON tool-call path or the exact-string `string_replace` editor. Wired through the existing tool-profile system so tiny models can drop down to this lower rung automatically.

**Context**
The codebase today has a single content-based editor (`source/tools/file-ops/string-replace.tsx`) that demands an exact `old_str`/`new_str` match. That works well for capable models but is fragile on weak local models that hallucinate whitespace, drift indentation, or struggle to reproduce long blocks byte-for-byte. The system prompt and tool definitions, by contrast, already bend toward weak models: `source/tools/tool-profiles.ts:23-30` defines a `nano` profile (≤4B params, 5 tools, ultra-slim prompt, single-tool mode) and the prompt has parallel ultra-slim variants for `nano` (see `isNanoProfile` in `source/tools/tool-profiles.ts:135-137`).

The edit tool itself does not bend — it is the same exact-match tool for nano models as for everything else. The local-tool-calling-reliability story already has multiple tiers (native function calling → XML fallback → JSON fallback, with malformed-output repair on both — see `docs/battlemap.md:181-183`). The edit tool needs its own rung below that: a diff-block format that Aider proved works on tiny local models. `search_file_contents` for `search/replace|diff.*format|edit_format` in `source/` returned no hits, so nothing in this shape exists yet.

Concretely, the work would land in three places:
- A new tool at `source/tools/file-ops/diff-edit.tsx` (alongside `source/tools/file-ops/string-replace.tsx` and `source/tools/file-ops/write-file.tsx`), exporting a `NanocoderToolExport` per the pattern in `source/types/core.ts:194`. It can reuse `createFileToolApproval('diff_edit')`, `hasSeenFile`/`markFileSeen`, `getCachedFileContent`/`invalidateCache` (`source/utils/read-tracker.ts`, `source/utils/file-cache.ts`), and the VS Code preview integration via `sendFileChangeToVSCode` already wired up in `source/tools/file-ops/string-replace.tsx`.
- Registration in `source/tools/index.ts:25-41` (`staticTools` array — the single source of truth).
- The profile and inference tables in `source/tools/tool-profiles.ts:23-30` and `source/types/config.ts:239` (`ToolProfile = 'auto' | 'full' | 'minimal' | 'nano'`), so that the new tool is gated to the right tier and the drift-guard test at `source/tools/tool-profiles.spec.ts:11-26` ("every profile names only registered tools") continues to pass.

**Proposed approach**
- Pick a single primary format to ship first (search/replace-block is the smaller surface; unified-diff is the more powerful superset). Recommended: search/replace-block with the model emitting fenced blocks like:
  ```
  <<<<<<< SEARCH
  ...exact old content...
  =======
  ...new content...
  >>>>>>> REPLACE
  ```
  Apply it line-by-line, validate that the search block is unique in the file (same uniqueness rule `string-replace.tsx:42-50` already enforces), and reuse the read-before-edit gate from `string-replace.tsx:218-226` (`hasSeenFile`).
- Keep `string_replace` as-is. The new tool is an additional rung, not a replacement. Selectable per-model: extend `ToolProfile` (or add a sibling setting `editFormat: 'string_replace' | 'diff_block'`) so the model-size heuristic in `inferToolProfile` (`source/tools/tool-profiles.ts:95-104`) can map ≤4B → nano profile that includes `diff_edit`, ≤15B → minimal profile that includes both editors, larger models → default `string_replace` only.
- Mirror the full `NanocoderToolExport` shape from `source/tools/file-ops/string-replace.tsx:268-274`: `tool`, `formatter` (render a colour-coded unified-style preview via `formatStringReplacePreview` patterns at `source/tools/file-ops/string-replace-preview.tsx`), `validator`, `approval`.
- Apply the patch atomically: parse all SEARCH/REPLACE blocks, validate every search block is found and unique before any write, then `writeFile` once and `invalidateCache` + `markFileSeen` per `source/tools/file-ops/string-replace.tsx:57-67`. Return a line-range summary modelled on `string-replace.tsx:69-95` so the model gets the same updated-file feedback.
- Extend `TOOL_PROFILES` and `TOOL_PROFILE_DESCRIPTIONS` / `TOOL_PROFILE_TOOLTIPS` in `source/tools/tool-profiles.ts:23-45` so the new tool is documented as part of `nano`.
- Add tests alongside the existing `string-replace.spec.tsx` pattern (parser fuzz tests against the same content the validator uses, uniqueness rejection, missing-search-block rejection, atomic-apply semantics) and add `diff_edit` to the drift-guard expectations.

**Acceptance criteria**
- A new `diff_edit` tool is exported from `source/tools/file-ops/` and registered in `source/tools/index.ts`, conforming to the `NanocoderToolExport` shape at `source/types/core.ts:194`.
- The tool accepts a `path` and a `diff` (or `blocks`) string containing one or more SEARCH/REPLACE fenced blocks, parses them, validates every search block exists exactly once in the on-disk file, applies them atomically, and writes the result via the existing `writeFile` + `invalidateCache` + `markFileSeen` path.
- The validator enforces the same read-before-edit gate (`hasSeenFile`) used by `source/tools/file-ops/string-replace.tsx:218-226`.
- The tool is included in the `nano` profile in `source/tools/tool-profiles.ts` and is selectable (either as part of `nano` or via an explicit `editFormat` setting) for weak local models; `string_replace` remains the default for `minimal` and `full`.
- `source/tools/tool-profiles.spec.ts` continues to pass, including the drift-guard test that every name in a profile resolves to a registered tool.
- `/tune`'s profile picker (`source/app/components/tune-selector.tsx`) reflects the change in description/tooltip, with wording consistent with the existing `TOOL_PROFILE_DESCRIPTIONS` / `TOOL_PROFILE_TOOLTIPS` copy.
- New spec file covers: single-block apply, multi-block atomic apply, search-block not found, search-block ambiguous (>1 match), empty `path`, unread file rejected by validator, and malformed input (unterminated fence, missing separator) rejected with a clear error.
- An end-to-end smoke test exercises a model emitting a diff block against a real file and asserts the post-edit content matches the union of the REPLACE blocks.

**Out of scope**
- Replacing `string_replace` or any change to its behaviour for existing profiles.
- Native (non-tool-call) streaming of diff blocks into the conversation; the first cut routes through the same tool-call pipeline as other editors.
- Full unified-diff (`--- a/` `+++ b/` `@@` hunks) parser — ship SEARCH/REPLACE first; unified-diff can follow if needed.
- Changes to the XML/JSON tool-call fallback parsers in `source/hooks/chat-handler/conversation/`.

Dhirenderchoudhary:
## Description

Adds a new `diff_edit` tool (`source/tools/file-ops/diff-edit.tsx`) designed specifically for weak local models (nano profile) that struggle with the exact-string matching required by the existing `string_replace` tool. 

Key changes:
- Implements parsing and atomic application of SEARCH/REPLACE blocks.
- Adds strict validation: search blocks must be unique and non-empty, and files must be read before editing (`hasSeenFile`).
- Includes VS Code extension integration for diff previews.
- Updates the `nano` profile to use `diff_edit` instead of `string_replace`.
- Excludes `diff_edit` from `plan` mode in `tool-manager.ts`.
- Full test coverage for parser, validator, formatter preview, and approval conditions.

Fixes #604

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

### Automated Tests

### Manual Testing

- [x] Tested with Ollama
- [ ] Tested with OpenRouter

## Checklist

- [x] Self-review completed

### DIFF
--- a/source/components/question-prompt.tsx
+++ b/source/components/question-prompt.tsx
@@ -55,10 +55,24 @@ export default function QuestionPrompt({
 		setSelectedIndex(0);
 	}
 
+	// Fix stale closures in useInput by keeping a ref to the latest state
+	const stateRef = useRef({
+		items,
+		selectedIndex,
+		isFreeformMode,
+		onAnswer,
+	});
+	stateRef.current = {
+		items,
+		selectedIndex,
+		isFreeformMode,
+		onAnswer,
+	};
+
 	const submitAnswer = (answer: string) => {
 		if (answeredRef.current) return;
 		answeredRef.current = true;
-		onAnswer(answer);
+		stateRef.current.onAnswer(answer);
 	};
 
 	const handleSelect = (item: OptionItem | undefined) => {
@@ -77,8 +91,9 @@ export default function QuestionPrompt({
 	};
 
 	useInput((input, key) => {
+		const state = stateRef.current;
 		// In freeform mode the TextInput owns typing; only handle Escape (back).
-		if (isFreeformMode) {
+		if (state.isFreeformMode) {
 			if (key.escape) {
 				setIsFreeformMode(false);
 				setFreeformValue('');
@@ -90,14 +105,14 @@ export default function QuestionPrompt({
 			submitAnswer('User declined to answer');
 			return;
 		}
-		if (items.length === 0) return;
+		if (state.items.length === 0) return;
 
 		if (key.upArrow || input === 'k') {
-			setSelectedIndex(i => (i - 1 + items.length) % items.length);
+			setSelectedIndex(i => (i - 1 + state.items.length) % state.items.length);
 		} else if (key.downArrow || input === 'j') {
-			setSelectedIndex(i => (i + 1) % items.length);
+			setSelectedIndex(i => (i + 1) % state.items.length);
 		} else if (key.return) {
-			handleSelect(items[selectedIndex]);
+			handleSelect(state.items[state.selectedIndex]);
 		}
 	});
 
--- a/source/tools/file-ops/diff-edit.spec.tsx
+++ b/source/tools/file-ops/diff-edit.spec.tsx
@@ -152,6 +152,37 @@ test('diff_edit applies multiple blocks atomically', async t => {
 	t.regex(result, /Successfully applied 2 diff blocks/);
 });
 
+test('diff_edit rejects duplicate search blocks and leaves file unchanged', async t => {
+	const filePath = await createTestFile('test.ts', 'alpha\n');
+
+	await t.throwsAsync(
+		async () => {
+			await executeDiffEdit({
+				path: filePath,
+				diff: [diffBlock('alpha', 'alphaX'), diffBlock('alpha', 'beta')].join(
+					'\n',
+				),
+			});
+		},
+		{message: /Search block 2 duplicates an earlier search block/},
+	);
+
+	t.is(await readFile(filePath, 'utf-8'), 'alpha\n');
+});
+
+test('diff_edit returns updated file contents for the model', async t => {
+	const filePath = await createTestFile('test.ts', 'alpha\n');
+
+	const result = await executeDiffEdit({
+		path: filePath,
+		diff: diffBlock('alpha', 'beta'),
+	});
+
+	t.regex(result, /Successfully applied 1 diff block/);
+	t.regex(result, /Updated file contents:/);
+	t.regex(result, /1: beta/);
+});
+
 test('diff_edit rejects missing search content and leaves file unchanged', async t => {
 	const filePath = await createTestFile('test.ts', 'alpha\ngamma\n');
 
@@ -247,4 +278,11 @@ test('diff_edit formatter renders a preview', async t => {
 	t.regex(lastFrame()!, /test\.ts/);
 	t.regex(lastFrame()!, /const oldValue/);
 	t.regex(lastFrame()!, /const newValue/);
+});
+
+test('diff_edit description tells models not to wrap diff in code fences', t => {
+	t.regex(
+		diffEditTool.tool.description,
+		/do not wrap.*code fence|code fence.*do not wrap/i,
+	);
 });
--- a/source/tools/file-ops/diff-edit.spec.tsx
+++ b/source/tools/file-ops/diff-edit.spec.tsx
@@ -1,0 +1,250 @@
+import {mkdtemp, readFile, rm, writeFile} from 'node:fs/promises';
+import {join, relative} from 'node:path';
+import test from 'ava';
+import {render} from 'ink-testing-library';
+import React from 'react';
+import {themes} from '../../config/themes.js';
+import {ThemeContext} from '../../hooks/useTheme.js';
+import {resolveToolApproval} from '../approval-policy.js';
+import {clearReadTracker, markFileSeen} from '../../utils/read-tracker.js';
+import {diffEditTool, parseDiffEditBlocks} from './diff-edit.js';
+
+console.log(`\ndiff-edit.spec.tsx - ${React.version}`);
+
+function TestThemeProvider({children}: {children: React.ReactNode}) {
+	const themeContextValue = {
+		currentTheme: 'tokyo-night' as const,
+		colors: themes['tokyo-night'].colors,
+		setCurrentTheme: () => {},
+	};
+
+	return (
+		<ThemeContext.Provider value={themeContextValue}>
+			{children}
+		</ThemeContext.Provider>
+	);
+}
+
+let testDir: string;
+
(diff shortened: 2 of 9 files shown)


will-lamerton:
Hi @Dhirenderchoudhary - looking good on the surface. I notice it's still a draft, so, mark as ready to review when you're happy and I can do so :)

Dhirenderchoudhary:
@will-lamerton ready for review

will-lamerton:
Hey @Dhirenderchoudhary0001 - thanks for this PR :D 

The new tool, profile swap, validator, and spec coverage all line up with issue #604. A handful of suggestions before merging :)

**Suggestions**

- `source/tools/file-ops/diff-edit.tsx:269-281` - the formatter's "VS Code diff preview" branch parses, validates, and applies the blocks a second time purely to compute the preview. If `getCachedFileContent` or `validateBlocks` throws, the `try/catch` swallows it silently and the inline preview is still rendered (good). But if the model later edits the same path twice, `vscodeChangeIds.set(absPath, changeId)` overwrites the prior id without `closeDiffInVSCode` on the old one - call `closeDiffInVSCode(previousId)` before overwriting so the stale VS Code tab closes.
- `source/tools/file-ops/diff-edit.tsx:53-54` - `countOccurrences` uses `content.split(search).length - 1`. If `search` is empty (already guarded at parse time) this would return `Infinity`, and if `search` contains a `$`-pattern that `String.prototype.replace` interprets, behaviour is fine here because you validate `=== 1` then call `replace` literally - but the same `search` may also appear as a substring inside `replace`, which `applyBlocks` would re-match on the next iteration. Worth a unit test asserting that an `alpha` search block followed by a `alphaX` replace block doesn't double-apply.
- `source/tools/file-ops/diff-edit.tsx:162-180` - the `inputSchema` description tells the model to use the markers but the issue's example payload wraps the whole block in a triple-backtick fence. Consider stating "do NOT wrap in a code fence" in the description so a small model following the issue doc verbatim doesn't get a `Missing <<<<<<< SEARCH marker` error from your `parseDiffEditBlocks`.
- `source/tools/file-ops/diff-edit.tsx:206-218` -`DiffEditPreview` re-parses the diff on every render. If the formatter is invoked during a streaming transcript it will be called per-token; cheap, but worth memoising on `args.diff`.
- `source/tools/file-ops/diff-edit.tsx:245-249` -when `result` is set you display `result.split('\n\n')[0]` (the success line). The actual `result` value also contains the `Updated file contents:` block, which is the part the model is meant to re-read; that data is correct for the tool result but the inline preview only shows the headline. That's fine for the UI, but make sure the full string is what gets fed back into the model's context - confirm by reading `source/tools/file-ops/diff-edit.tsx:163` (`execute` returns the full string) vs how `tool-manager.ts` threads it through. From the diff it looks right; flagging only because the formatter-vs-result divergence is a classic foot-gun.

Dhirenderchoudhary:
@will-lamerton All required things are done have a look.

Dhirenderchoudhary:
@will-lamerton 
I checked and I thinl the Semgrep failure. It is not coming from the `diff_edit` changes in this PR.

The scan is failing on existing repo-level security/config findings in `.github/workflows/*`, `.github/dependabot.yml`, and `pnpm-workspace.yaml`:
- GitHub Actions are using mutable tags like `actions/checkout@v4`
- Dependabot is missing cooldown config
- pnpm workspace is missing supply-chain hardening settings

My PR only touches `source/tools/file-ops/diff-edit.tsx` and its spec, so I think this should either be handled separately as a repo security-hardening PR or the Semgrep policy should be adjusted.

If you allow can i push a PR for Semgrep fixes

will-lamerton:
> @will-lamerton I checked and I thinl the Semgrep failure. It is not coming from the `diff_edit` changes in this PR.
> 
> The scan is failing on existing repo-level security/config findings in `.github/workflows/*`, `.github/dependabot.yml`, and `pnpm-workspace.yaml`:
> 
> * GitHub Actions are using mutable tags like `actions/checkout@v4`
> * Dependabot is missing cooldown config
> * pnpm workspace is missing supply-chain hardening settings
> 
> My PR only touches `source/tools/file-ops/diff-edit.tsx` and its spec, so I think this should either be handled separately as a repo security-hardening PR or the Semgrep policy should be adjusted.
> 
> If you allow can i push a PR for Semgrep fixes

Yes please - feel free :D

Issue #619, opened 2026-07-01
[Feature]: Continuous Semantic Memory


akramcodez:
## Summary

One limitation of current AI coding assistants is that they effectively restart from scratch every new conversation. While they understand the current prompt well, they lose architectural decisions, coding conventions, previous bug fixes, and historical context between sessions.

I'd like to explore a local-first semantic memory system for Nanocoder that gradually builds a "project brain" over time, allowing the assistant to remember previous work without requiring developers to continually re-explain the project.

This issue aims to foster discussion on the design and architecture before we begin implementation.

## Motivation

Current workflows often look like:

1. Explain architecture
2. Explain conventions
3. Explain previous refactors
4. Explain why certain approaches were rejected
5. Repeat every new session

Over time this becomes significant prompt overhead.

**Example:**

_User:_ Refactor the auth flow.

_Current AI:_ Starts from zero.

_Desired Nanocoder:_ "Last week this project migrated from JWT to Clerk. Middleware was intentionally avoided. Should I follow the same pattern?"

## Goals

The goal is not to remember entire conversations word-for-word, but rather to gradually remember durable project knowledge, such as:

- Architectural decisions
- Project conventions
- Previous refactors
- Coding style
- Recurring TODOs
- Important bug fixes
- Design rationale

## Proposed Architecture (Grounded in Nanocoder)

Nanocoder already stores sessions and prompt history using local files. To stay consistent with the existing architecture and keep the first iteration lightweight, we can introduce a new **SemanticMemoryManager** that stores structured semantic memories in repository-scoped JSON files behind a clean storage abstraction.

This allows us to validate the feature with minimal complexity while leaving room to introduce a more advanced storage backend in the future if semantic search or embeddings become necessary.

### The Flow

1. **Interactive session ends:** After a successful session or `nanocoder run` execution.
2. **Summarizer Agent:** A background process uses the configured LLM (or optional local model) to extract durable facts from the completed session.
3. **Storage:** Extracted memories are stored in a repository-scoped JSON file (for example `~/.nanocoder/memory/<repo-hash>/memory.json`) managed through `SemanticMemoryManager`.
4. **Retrieval:** Before constructing a prompt, Nanocoder loads the stored memories, selects the most relevant entries using lightweight relevance scoring (keywords, tags, metadata, recency), and prepares them for injection.
5. **Context Injection:** Relevant memories are appended to the system prompt as "Project Context".

### Visual Architecture

```mermaid
graph TD
[...]

PR #678, opened 2026-07-17
feat: inject semantic memories into prompts(Phase 3)

Dhirenderchoudhary:
## Description

Adds Phase 3 of semantic memory: retrieves relevant repo-scoped memories and appends them to the system prompt as `Project Context` before the model call. This uses the existing `SemanticMemoryManager.findRelevantMemories()` path and keeps injection limited to the top relevant memories.

Phase 3: #619

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

### Automated Tests

### Manual Testing

- [ ] Tested with Ollama
- [ ] Tested with OpenRouter

## Checklist

- [x] Self-review completed

### DIFF
--- a/source/hooks/chat-handler/types.ts
+++ b/source/hooks/chat-handler/types.ts
@@ -1,5 +1,9 @@
 import type React from 'react';
 import type {CustomCommandLoader} from '@/custom-commands/loader';
+import type {
+	MemoryFinder,
+	ProjectContextOptions,
+} from '@/memory/project-context';
 import type {Task} from '@/tools/tasks/types';
 import type {ToolManager} from '@/tools/tool-manager';
 import type {TuneConfig} from '@/types/config';
@@ -49,6 +53,8 @@ export interface UseChatHandlerProps {
 	subagentsReady?: boolean;
 	privacySessionMapRef?: React.MutableRefObject<Record<string, string>>;
 	privacyEnabled?: boolean;
+	memoryFinder?: MemoryFinder;
+	projectContextOptions?: ProjectContextOptions;
 }
 
 export interface ChatHandlerReturn {
--- a/source/hooks/chat-handler/useChatHandler.spec.tsx
+++ b/source/hooks/chat-handler/useChatHandler.spec.tsx
@@ -416,6 +416,67 @@ test('useChatHandler - drains queued message when setup fails before conversatio
 	rendered.unmount();
 });
 
+test('useChatHandler - injects project context from memory finder', async t => {
+	let hookResult: ChatHandlerReturn | null = null;
+	let sentMessages: Message[] = [];
+	const client: LLMClient = {
+		...createMockClient(),
+		chat: async (messages, _tools, callbacks) => {
+			sentMessages = messages;
+			callbacks.onFinish?.();
+			return {
+				choices: [
+					{
+						message: {
+							role: 'assistant',
+							content: 'ok',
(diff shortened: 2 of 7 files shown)


akramcodez:
Hi @Dhirenderchoudhary, thanks for shipping Phase 3! The overall direction looks good, but after reviewing the implementation against the Phase 3 requirements in #619, I found a few blockers that should be addressed before we consider this phase complete.

### 1. Performance Regression: `SemanticMemoryManager` is recreated on every message

**Where:** `source/hooks/chat-handler/useChatHandler.tsx`

**What is wrong:**
`appendRelevantProjectContext()` currently relies on its default `memoryFinder`, which creates a new `SemanticMemoryManager()` for every user message.

**Why it is wrong:**
`SemanticMemoryManager` determines the repository scope by running `git config --get remote.origin.url`. Although the manager has an internal cache, recreating it for every message bypasses that cache and results in a blocking Git command being executed repeatedly in the main chat flow.

**What should be done:**
Instantiate `SemanticMemoryManager` once (for example using `useMemo` or by injecting it through `UseChatHandlerProps`) and pass that shared instance into `appendRelevantProjectContext(...)` instead of relying on the default constructor.


### 2. Phase 3 retrieval heuristics are only partially implemented

**Where:** `source/memory/semantic-memory-manager.ts`

**What is wrong:**
The issue defines the retrieval heuristics as:

* keyword matching
* category matching
* recency
* repository scope

The current implementation scores memories using keyword matching and then falls back to recency, but it does not incorporate category matching into the ranking.

**Why it is wrong:**
This means the retrieval logic doesn't fully satisfy the Phase 3 requirements described in the issue.

**What should be done:**
Extend the relevance scoring so that matching memory categories also contribute to the overall relevance score before sorting.


### 3. Configurable token budget has not been implemented

**Where:** `source/memory/project-context.ts`

**What is wrong:**
The issue requires injected memories to respect a **configurable token budget**. The current implementation instead uses a hardcoded item limit (`PROJECT_CONTEXT_LIMIT = 3`).

**Why it is wrong:**
A fixed number of memories is not equivalent to a configurable token budget and doesn't allow the injection size to be tuned for different model context windows.

**What should be done:**
Replace the hardcoded limit with a configurable budget (or introduce the configuration mechanism planned for this feature) so the injected context size can be adjusted.


### 4. Tests are no longer hermetic and the new behavior isn't fully covered

**Where:** `source/hooks/chat-handler/useChatHandler.spec.tsx`

**What is wrong:**
`useChatHandler` now instantiates a real `SemanticMemoryManager`, causing tests to depend on the local semantic memory store instead of a mocked implementation.

Additionally, there isn't a test verifying that the retrieved project context is actually appended to the system prompt.

**Why it is wrong:**
This makes the tests dependent on local machine state and leaves the primary Phase 3 behavior unverified.

**What should be done:**

* Inject a mock `SemanticMemoryManager` (or memory finder) into `useChatHandler`.
* Update the existing tests to use that mock.
* Add a test that verifies the generated system prompt includes the injected `## Project Context` returned by the mock implementation.


Overall, the retrieval and prompt injection flow is heading in the right direction, but I don't think we can consider Phase 3 complete until these issues are addressed. Once they're fixed, I'll be happy to review again. Great work so far!

Dhirenderchoudhary:
@akramcodez please have a look

PR #684, opened 2026-07-19
feat: improve model picker UX for large model catalogs

### LINKED ISSUE #683: [Feature] Model picker doesn't scale well for large model catalogs
## Description

The current model picker works well when there are only a handful of models (e.g. 2–4), but it quickly becomes frustrating once a provider exposes dozens or hundreds of models.

Providers such as OpenRouter, Ollama, and GitHub Models can expose very large model catalogs. In these cases, the current picker becomes difficult to navigate and doesn't scale well.

Current pain points include:

- The picker renders every available model in a single long list.
- There is no way to filter or search for a model by typing.
- Navigation relies entirely on repeatedly pressing the ↑/↓ arrow keys.
- As the selection moves downward, the highlighted row eventually reaches the bottom of the screen, forcing the user to continually follow it with their eyes.
- If the desired model is missed, the user has to scroll back through the same long list and repeat the process.
- The currently selected model is easy to lose track of while navigating.

This makes switching models unnecessarily time-consuming for providers with large model catalogs.

## Screenshots/Vedio

https://github.com/user-attachments/assets/d1050cbb-a13f-42ac-bf68-321bbb5201a5

## Environment

- **OS**: Windows 11
- **Provider**: OpenRouter
- **Model**: (Any)

## Steps to Reproduce

1. Configure a provider with a large model catalog (e.g. OpenRouter).
2. Open the `/model` picker.
3. Attempt to switch to a model near the bottom of the list using the keyboard.
4. Continue navigating with the ↓ key.

## Expected Behavior

The model picker should remain efficient regardless of how many models are available.

Ideally it should:

- Support search-as-you-type filtering.
- Display a fixed-height scrollable list.
- Keep the highlighted item in a stable position while the list scrolls.
- Open with the currently selected model highlighted.

## Actual Behavior

The picker renders the entire model list and requires navigating every item sequentially.

As the highlighted row moves toward the bottom of the visible list, it becomes increasingly difficult to keep track of the current position. Overshooting a model requires scrolling back through the list again, making model switching tedious for providers with large catalogs.

## Logs/Screenshots

See attached recording demonstrating the current navigation experience.

## Recommended fix

The current picker is built on top of `ink-select-input`, which works well for small lists but doesn't scale well once providers expose dozens or hundreds of models.

A custom Ink component could provide a much better navigation experience while preserving the existing keyboard-first workflow.

## Proposed improvements

- Search-as-you-type filtering using fuzzy matching.
- Fixed-height scrollable window instead of rendering the entire list.
- Centered/sliding selection so the highlighted row remains in a stable position while the list scrolls underneath it.
- Current model preselection when opening the picker.
- Responsive visible row count based on terminal height.

## Implementation notes

Replace `ink-select-input` with a custom `FilterableSelectList` component that provides:

- Fuzzy filtering
- Fixed-height scrolling
- Centered/sliding window navigation
- Current model preselection
- Keyboard-first navigation

This preserves the current experience for small model lists while scaling much better for providers exposing large model catalogs.

rakshith1928:
## Description

Fixes #683 

Makes nanocoder's `/model` picker behave like what modern cli's have : **fuzzy search filter + capped, centered scrolling window** so the list never overflows the terminal, with the current model preselected and marked `(current)`.

### Problem

`/model` renders every configured model in one tall `SelectInput` (ink-select-input ^6.2.0). With 300+ configured models, the titled box grows taller than the terminal — it eats the whole screen, has no search, and only `↑/↓` + `Escape` navigate.

Root cause:
- `source/components/item-selector.tsx:98` — `<SelectInput items={items} />` has no `maxHeight`/virtualization/search
- `source/components/ui/titled-box.tsx` — `TitledBoxWithPreferences` receives no `height` cap; box grows to fit all rows
- `source/components/model-selector.tsx:24-40` — flat list built once, no filter state

### Approach

Extract a new pure **presentational** component `FilterableSelectList` mirroring the proven filter + sliding-window pattern in `source/wizards/steps/model-selection-list.tsx:35-63`. `ItemSelector` gains an opt-in `searchable` prop; when true (and not loading/error/empty) it renders `FilterableSelectList` instead of `SelectInput`. `ModelSelector` flips `searchable` on and preselects the current model. **No new runtime dependencies.** Reuses `fuzzyScore` from `source/utils/fuzzy-matching.ts`.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

| File | Change |
|---|---|
| `source/components/filterable-select-list.tsx` | **New.** Pure presentational: filter + sliding-window + key handlers. `visibleCount` default 12, height-aware clamp via `useStdout`. |
| `source/components/filterable-select-list.spec.tsx` | **New.** 16 tests. |
| `source/components/item-selector.tsx` | Add `searchable` / `visibleCount` / `initialSelectedValue` props; render `FilterableSelectList` when opted in. |
| `source/components/item-selector.spec.tsx` | **New.** 5 tests. |
| `source/components/model-selector.tsx` | Pass `searchable`, compute `currentIndex` for preselect. |
| `source/components/model-selector.spec.tsx` | 3 new wiring-only tests. |

## Key behaviors

- **Always-on filter** — any printable char feeds the query (no `/` prefix). `Type to filter · ↑↓ navigate · Enter select · Esc cancel` hint.
- **Capped centered window** — `visibleCount = 12` (matches wizard `MODEL_SELECTION_VISIBLE_ITEMS`). `effectiveVisibleCount = max(1, min(visibleCount, (stdout.rows ?? 24) - 4))` so short terminals clamp and can't reproduce the original overflow.
- **Scroll math** mirrors `model-selection-list.tsx:53-63` — `scrollStart = max(0, min(highlightedIndex − floor(visible/2), len − visible))`.
- **Current model preselect** — stays at natural list index, marked `(current)`; `initialSelectedValue` preselects without re-sorting. Missing current model → no preselect (fallback to index 0).
- **Navigation** — `↑/↓`, `Home`/`End`, `PageUp`/`PageDown`, `Backspace`/`Delete` erase from query, `Enter` selects, `Escape` cancels.

### Before

https://github.com/user-attachments/assets/d577d8ef-0bae-4420-b48a-81d03a0f6ee5

### After

https://github.com/user-attachments/assets/b859b54a-0f4a-4f18-8bb2-41bc3bf8e310

## One non-obvious decision (single owner for Escape)

Ink 6's `useInput` is **broadcast** — every registered handler fires on every keystroke (`use-input.js:117`, `App.js:85-88`). A naive "keep both escape handlers" double-fires `onCancel`.

The original plan said delete `ItemSelector`'s handler and rely on the section-level authority (`interactive-app.tsx:225-237`, `{isActive: cancellable}`). **Investigation proved that wrong** — `cancellable` (`interactive-app.tsx:143-151`) is `false` during `ModelSelector` loading/error, so Escape would be **lost** there.

**Fix:** `ItemSelector` keeps a single `useInput` gated by Ink's `isActive` option so it owns Escape only in loading/error branches where `FilterableSelectList` is unmounted:
```tsx
{isActive: !searchable || loading || error != null}
```

(error != null, not error !== null — error is string | null | undefined; the latter wrongly keeps the handler active when error is omitted and double-fires.) Verified by two regression tests: Escape fires in the error branch, and does NOT double-fire in the normal searchable path.

### Automated Tests

Note: manual smoke (run /model, arrow-down ~6 times to confirm centered 12-row window, type gpt to confirm filtering, type zzzzz for empty state) is described in the plan file.

### Manual Testing

- [ ] Tested with Ollama
- [x] Tested with OpenRouter

## Checklist

- [x] Self-review completed

## Out of scope (follow-ups)
- Generalizing searchable to other long-list selectors (session-selector, checkpoint-selector, tune-selector, settings-selector) — FilterableSelectList is already generic (ItemSelectorOption<TValue>[]), so each later flip is ~10 lines.
- Mouse support, category grouping, fuzzy match highlighting — opencode-only features, intentionally deferred.

### DIFF
--- a/source/components/filterable-select-list.spec.tsx
+++ b/source/components/filterable-select-list.spec.tsx
@@ -1,0 +1,305 @@
+// source/components/filterable-select-list.spec.tsx
+import test from 'ava';
+import {render} from 'ink-testing-library';
+import React from 'react';
+import {FilterableSelectList} from './filterable-select-list.js';
+import {renderWithTheme} from '../test-utils/render-with-theme.js';
+
+console.log('\nfilterable-select-list.spec.tsx');
+
+const items = [
+	{label: 'alpha (openai)', value: '0'},
+	{label: 'beta (ollama)', value: '1'},
+	{label: 'gamma (openrouter)', value: '2'},
+];
+
+test('renders all items when query is empty', t => {
+	const {lastFrame} = renderWithTheme(
+		<FilterableSelectList items={items} onSelect={() => {}} />,
+	);
+	const out = lastFrame()!;
+	t.regex(out, /alpha \(openai\)/);
+	t.regex(out, /beta \(ollama\)/);
+	t.regex(out, /gamma \(openrouter\)/);
+});
+
+test('filters by label substring case-insensitively', t => {
+	const many = [
+		{label: 'gpt-4o (openai)', value: '0'},
+		{label: 'gpt-4o-mini (openai)', value: '1'},
+		{label: 'llama3 (ollama)', value: '2'},
+	];
+	const {lastFrame, stdin, unmount} = renderWithTheme(
+		<FilterableSelectList items={many} onSelect={() => {}} />,
+	);
+	stdin.write('gpt');
+	// allow ink to process key — 50ms matches repo precedent (model-selector.spec.tsx:206)
+	return new Promise<void>(resolve => {
+		setTimeout(() => {
+			const out = lastFrame()!;
+			t.regex(out, /gpt-4o \(openai\)/);
+			t.regex(out, /gpt-4o-mini \(openai\)/);
+			t.notRegex(out, /llama3/);
+			unmount();
+			resolve();
+		}, 50);
+	});
+});
+
+test('empty query shows all items in original order', t => {
+	const many = [
+		{label: 'zeta (a)', value: '0'},
+		{label: 'alpha (b)', value: '1'},
+	];
+	const {lastFrame, unmount} = renderWithTheme(
+		<FilterableSelectList items={many} onSelect={() => {}} />,
+	);
+	const out = lastFrame()!;
+	t.regex(out, /zeta \(a\)[\s\S]*alpha \(b\)/);
+	unmount();
+});
+
+test('ranks exact > contains for fuzzy filter', t => {
+	const many = [
+		{label: 'claude-3 (anthropic)', value: '0'},
+		{label: 'claudette (x)', value: '1'},
+	];
+	const {lastFrame, stdin, unmount} = renderWithTheme(
+		<FilterableSelectList items={many} onSelect={() => {}} />,
+	);
+	stdin.write('claude');
+	return new Promise<void>(resolve => {
+		setTimeout(() => {
+			const out = lastFrame()!;
+			const i0 = out.indexOf('claude-3 (anthropic)');
+			const i1 = out.indexOf('claudette (x)');
+			t.true(i0 >= 0 && i1 >= 0);
+			t.true(i0 < i1);
+			unmount();
+			resolve();
+		}, 50);
+	});
+});
+
+test('caps visible window to visibleCount', t => {
+	const many = Array.from({length: 30}, (_, i) => ({
+		label: `m-${i}`,
+		value: String(i),
+	}));
+	const {lastFrame, unmount} = renderWithTheme(
+		<FilterableSelectList items={many} visibleCount={12} onSelect={() => {}} />,
+	);
+	const out = lastFrame()!;
+	const visible = many
+		.map(m => m.label)
+		.filter(label => out.includes(label));
+	t.is(visible.length, 12);
+	unmount();
+});
+
+test('centers highlight in the middle of the window', t => {
+	const many = Array.from({length: 30}, (_, i) => ({
+		label: `m-${i}`,
+		value: String(i),
+	}));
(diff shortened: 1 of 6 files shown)


rakshith1928:
## Re: CI failures (knip + semgrep)
These two CI failures are pre-existing on main and are not introduced by this PR. I deliberately left them out of this branch to keep the change scoped to the model picker.
## knip  usePrivacyContext unused
- Reported at source/context/privacy-context.tsx:13.
- This export is unused across the entire codebase (only its own definition references it). It was left behind by the stateless-API migration in #645 (refactor: migrate to stateless API and history-boundary rehydration).
- It exists on main today and will fail knip there too. I'll address it in a separate cleanup PR rather than bundling an unrelated file change into this feature PR.
## semgrep - 45 findings
- All 45 are in infra/config files only: .github/dependabot.yml, .github/workflows/*.yml, .npmrc, pnpm-workspace.yaml.
- Zero findings in source/ — the searchable picker code is clean.
- These are supply-chain hardening items (pin action SHAs, add cooldown/min-release-age/minimumReleaseAge/blockExoticSubdeps/trustPolicy). They are main-level hygiene and best handled in a dedicated CI-hardening PR.

**Diff confirmation**
A git diff origin/main -- <those files> is empty — none of them are touched by this branch. This PR adds only filterable-select-list.tsx, item-selector.tsx, model-selector.tsx (+ co-located specs).

will-lamerton:
Hey @rakshith1928 - looks brilliant :D

PR #685, opened 2026-07-19
fix: validate checkpoint name to prevent path traversal

### LINKED ISSUE #686: [Bug] Path traversal in CheckpointManager allows deleting/reading paths outside the workspace
## Description

Checkpoint names aren't validated on load or delete. `getCheckpointDir` joins the name into a path directly:

```ts
private getCheckpointDir(name: string): string {
	return path.join(this.checkpointsDir, name); // nosemgrep
}
```

`/checkpoint delete` passes raw user input as the name, so `../../foo` points outside `.nanocoder/checkpoints/`. `deleteCheckpoint` calls `fs.rm(dir, {recursive: true, force: true})`, so it can delete a directory outside the workspace. Only `saveCheckpoint` validates the name; load/delete/metadata/exists don't. The `path.join` here has a `// nosemgrep`, so it was flagged and skipped.

## Steps to Reproduce

1. Start nanocoder in a project.
2. `/checkpoint create test`.
3. `/checkpoint delete ../../<a-dir-outside-the-project>`.

## Expected Behavior

Name rejected, nothing outside the checkpoints dir touched.

## Actual Behavior

Path resolves outside the checkpoints dir and gets removed.

## Additional Context

Fix in #685.

YASHSHARMAOFFICIALLY:
## Description

Checkpoint names weren't validated on load or delete. `getCheckpointDir` joined the name into a path directly, and every read/delete method goes through it:

```ts
private getCheckpointDir(name: string): string {
	return path.join(this.checkpointsDir, name); // nosemgrep
}
```

`/checkpoint delete` passes raw user input as the name, so `../../foo` points outside `.nanocoder/checkpoints/`. `deleteCheckpoint` runs `fs.rm(..., {recursive: true, force: true})`, so it could delete a directory outside the workspace. Only `saveCheckpoint` validated the name before.

Fix validates the name and checks the resolved path stays inside the checkpoints dir, in `getCheckpointDir` so load/delete/metadata/exists are all covered in one place.

## Type of Change

- [x] Bug fix

## Testing

- [x] New tests in `checkpoint-manager.spec.ts` for traversal in load and delete
- [x] Existing checkpoint tests still pass

## Checklist

- [x] Self-review completed
- [x] No breaking changes

Closes #686

### DIFF
--- a/source/services/checkpoint-manager.spec.ts
+++ b/source/services/checkpoint-manager.spec.ts
@@ -312,6 +312,47 @@ test.serial(
 	},
 );
 
+test.serial(
+	'CheckpointManager rejects path traversal in delete',
+	async t => {
+		const tempDir = await createTempDir();
+		// A file outside the checkpoints dir that a traversal name would target.
+		const outside = path.join(tempDir, 'outside');
+		await fs.mkdir(outside, {recursive: true});
+		try {
+			const manager = new CheckpointManager(tempDir);
+
+			await t.throwsAsync(
+				async () => {
+					await manager.deleteCheckpoint('../../outside');
+				},
+				{message: /Invalid checkpoint name|invalid characters/},
+			);
+
+			// The escape target must still be there.
+			t.true(existsSync(outside));
+		} finally {
+			await cleanupTempDir(tempDir);
+		}
+	},
+);
+
+test.serial('CheckpointManager rejects path traversal in load', async t => {
+	const tempDir = await createTempDir();
+	try {
+		const manager = new CheckpointManager(tempDir);
+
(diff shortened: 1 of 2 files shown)


A-S-Manoj (approved):
Fix validates checkpoint names at the single choke point (getCheckpointDir()) that all accessors route through, closing the gap where only saveCheckpoint was validated. It blocks unsafe characters and confirms the resolved path stays within checkpointsDir, directly addressing the traversal issue in #686. Verified with the new tests reproducing the exact repro steps — both pass. LGTM, approving.

akramcodez (approved):
LGTM 👍

PR #700, opened 2026-07-22
feat: add interactive subagent session attachment for real-time monitoring and debugging

### LINKED ISSUE #679: Interactive Subagent Session Inspection
## Description

Add the ability to attach to a running subagent session from the terminal UI for interactive debugging.

When an orchestrator delegates work to a subagent, the user should be able to temporarily switch from the parent session into the subagent's live execution, then return to the parent when desired. This would make developing and debugging custom subagents significantly easier while preserving the existing orchestration workflow.

## Use Case

When developing custom subagents, it can be difficult to understand why a delegated task succeeds, fails, or appears to stall.

Currently, the parent agent shows that a subagent has been delegated work and eventually receives its final output, but there is no way to inspect the subagent while it is running.

This makes it difficult to diagnose issues such as:

* Reasoning loops
* Repeated or unnecessary tool calls
* Prompt conflicts
* Poor stopping behaviour
* Excessive context gathering
* General prompt engineering issues

Many reasoning models stream their thinking when they are the active session. However, once they are running as delegated subagents, that visibility is no longer available, making prompt iteration largely trial and error.

Having the ability to inspect a running subagent would greatly improve the workflow for anyone building custom agents.

## Proposed Solution

Allow users to attach to a running subagent from the TUI, similar to how OpenCode allows switching between delegated agent sessions.

For example:

```text
Main Session
├─ Research Agent    Running...
├─ Test Agent        Waiting...
└─ Reviewer          Complete
```

Selecting (or using a keyboard shortcut on) a running subagent would switch the UI into that subagent's session.

While attached, the UI would display the same information that would normally be available if that agent had been launched directly, including:

* Streaming model output
* Streaming reasoning/thinking (when supported by the provider/model)
* Tool calls
* Tool results
* Progress updates
* Token usage

A breadcrumb could indicate the current context:

```text
Main Session
└── Research Agent
```

Pressing **Esc** (or another shortcut) would immediately return to the parent session while allowing the subagent to continue executing.

Once the subagent completes, the parent workflow continues exactly as it does today.

## Alternatives Considered

An alternative would be a developer/debug mode that simply mirrors a subagent's streamed output into the parent session.

While this would provide additional visibility, I believe being able to temporarily attach to the delegated session would offer a much cleaner user experience, especially as subagents become more complex or begin spawning additional subagents.

Another possibility would be to save complete subagent transcripts to trajectory logs for later inspection. While useful for post-mortem debugging, it would not provide the same interactive experience as observing the subagent while it is running.

## Additional Context

This request is **not** asking for hidden chain-of-thought to be exposed.

Rather, when a provider already supports streaming reasoning or thinking, this feature would allow users to observe that output while the delegated subagent is executing, just as they can when the same agent is running as the active session.

This would make developing and debugging custom subagents considerably easier without changing the default orchestration experience for normal users.

* [x] I have searched existing issues to ensure this is not a duplicate
* [x] This feature aligns with the project's goals (local-first AI assistance)

One possible implementation would be to treat each subagent as a navigable session in the TUI.

When a subagent is launched, it would appear in a session tree beneath the parent. Users could attach to any active subagent to observe its live execution and detach at any time to return to the parent session.

This approach would preserve the current orchestration model while providing much-needed observability for users developing and debugging custom agents.

akramcodez:
Closes #679

### Description
This PR adds the ability to attach to a running subagent session from the terminal UI for interactive debugging. This makes developing and debugging custom subagents significantly easier while preserving the existing orchestration workflow.

### What changed
- **Subagent Session Store:** Added a lightweight `SubagentSessionStore` to decouple raw LLM token/reasoning streaming from the parent orchestration state.
- **Executor Hooks:** Updated `SubagentExecutor` to pump tokens, chunks, and tool events straight into the new session store as they are generated, cleaning up when the subagent finishes.
- **Subagent View UI:** Added a conditional `SubagentView` component that renders the live subagent trajectory. It includes a breadcrumb header indicating the active subagent (`Main Session > Subagent Name`).
- **Interactive Toggles:** Added a UI hint to `agent-progress.tsx` indicating that `Ctrl+S` is available. Bound `Ctrl+S` to switch the UI context to the running subagent, and `Esc` to immediately detach and return to the parent.

### Testing
- Verified that attaching/detaching does not interfere with the parent agent's execution state or generation stream.
- The session safely cleans up when the subagent finishes, even if the user is currently attached (automatically detaching the user).

### DIFF
--- a/.changeset/interactive-subagent-inspection.md
+++ b/.changeset/interactive-subagent-inspection.md
@@ -1,0 +1,5 @@
+---
+"@nanocollective/nanocoder": minor
+---
+
+Added the ability to attach to a running subagent session from the terminal UI for interactive debugging. This feature allows users to inspect exactly what a subagent is doing in real-time, including streaming text and reasoning. You can press `Ctrl+S` while a subagent is running to attach to it, cycle between multiple running subagents, and press `Esc` to detach.
--- a/source/app/App.tsx
+++ b/source/app/App.tsx
@@ -136,13 +136,29 @@ export default function App({
 		if (key.ctrl && input === 'c') {
 			handleExit();
 		}
-		if (key.ctrl && input === 's' && !appState.attachedAgentId) {
+		if (key.ctrl && input === 's') {
 			const progresses = Array.from(getAllSubagentProgress().entries());
-			const running = progresses.find(
-				([_, p]) => p.status !== 'complete' && p.status !== 'error',
-			);
-			if (running) {
-				appState.setAttachedAgentId(running[0]);
+			const runningAgents = progresses
+				.filter(([_, p]) => p.status !== 'complete' && p.status !== 'error')
+				.map(([id]) => id);
+
+			if (runningAgents.length === 0) {
+				if (appState.attachedAgentId) {
+					appState.setAttachedAgentId(null);
+				}
+			} else {
+				if (!appState.attachedAgentId) {
+					appState.setAttachedAgentId(runningAgents[0]);
+				} else {
+					const currentIndex = runningAgents.indexOf(appState.attachedAgentId);
+					if (currentIndex !== -1 && runningAgents.length > 1) {
+						appState.setAttachedAgentId(
+							runningAgents[(currentIndex + 1) % runningAgents.length],
+						);
+					} else if (currentIndex === -1) {
+						appState.setAttachedAgentId(runningAgents[0]);
+					}
+				}
 			}
 		}
 	});
--- a/source/app/App.tsx
+++ b/source/app/App.tsx
@@ -16,6 +16,7 @@ import {SuccessMessage} from '@/components/message-box';
 import SecurityDisclaimer from '@/components/security-disclaimer';
 import StreamingMessage from '@/components/streaming-message';
 import StreamingReasoning from '@/components/streaming-reasoning';
+import {SubagentView} from '@/components/subagent-view';
 import type {TitleShape} from '@/components/ui/styled-title';
 import {
 	shouldPromptExtensionInstall,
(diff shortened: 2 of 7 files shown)


will-lamerton:
Hey @akramcodez - this is really nice. Love it.

_Two issues and one question:_

**Issues:**
- The only bug I've found is when the message is streaming from the AI in the sub-agent session, the on a narrow terminal in vscode it looks like this which needs sorting (see screenshots):

Screenshot 2026-07-22 at 18 43 07

Screenshot 2026-07-22 at 18 42 02

- Can you also add a changeset for this? :)

**Question**

- The agent is able spawn up to 5 agents at once - how are we handling that when it comes to deciding which sub-agent session to allow us to view?

akramcodez:
I was thinking of handling it by making Ctrl+S context-aware. The first press attaches to the first active subagent, and subsequent presses cycle through the other running subagent sessions. If a session finishes, it'll be skipped automatically, and if no subagents remain, it'll return to the main session. This keeps the interaction lightweight without introducing an additional picker UI.

will-lamerton:
Hey @akramcodez - thanks for addressing the review points. The message display now looks a lot better! Just to let you know though Ctrl+S doesn't seem to cycle sub-agent sessions. I like this idea however, I can never get it to cycle past the first.

You can test this by saying: "Please run two sub-agents in parallel - one exploring git tools, the other exploring file tools. The aim of both is to find improvements" -  this usually gets the model to launch two sub-agents at the same time to test this feature.

Can you also add a changeset for this PR?

Once that's address I think we're good to merge :)

akramcodez:
Hey @will-lamerton! Thanks for the feedback. I've fixed the Ctrl+S cycling issue (it turned out to be a stale closure) and added the missing changeset. Ctrl+S now correctly cycles through active subagent sessions, skips completed ones, and returns to the main session when none remain. Whenever you get a chance, I'd appreciate another review. Thanks!

Issue #622, opened 2026-07-02
[Feature]  Realtime Voice Mode for Nanocoder


RONAK-AI647:
## Description

Brief description of the feature you'd like to see

I looked into what it would take to add voice interaction to Nanocoder — not "add Whisper," but a real bidirectional voice session: speak a request, get it transcribed, executed, and (optionally) spoken back.

This is a large feature (native audio bindings, VAD, interruption handling, wake word are all non-trivial), so I want to scope this as an RFC first and propose a very small, safe first PR rather than committing to the whole thing up front.

### Why provider-agnostic design matters here

Most voice-mode implementations I've seen (including my own earlier draft) lean on a single vendor's live/streaming API (e.g. Gemini Live). That's a poor fit for Nanocoder specifically, since the whole point of the project is "bring your own model." So instead of "Voice Mode," I'd propose this as a **Realtime Interaction Layer**, where voice is just one transport:

```
RealtimeSession
├── Text (existing)
├── Voice
└── future transports (screen share, etc.)
```

### Proposed first step (low-risk, no core changes)

I don't think the first PR should touch the request pipeline at all. Instead:

- Add a `voice.*` block to the settings schema (`enabled`, `activationMode`, `voiceName`) — additive only
- Add a `/voice` slash command stub (currently just toggles a flag, no audio yet)
- Add a push-to-talk keybinding entry
- Add a `VoiceStatusBar` Ink component (idle / listening / processing / speaking states, ASCII-only, no waveform)

This is fully testable without any audio hardware and doesn't touch `GeminiClient`, `ContentGenerator`, or any heavily-tested core file.

### What comes after (separate PRs, only once the above lands)

1. Extend `ContentGenerator` with an optional `liveConnect()` method (defaults unimplemented for providers that don't support it)
2. `packages/voice/` package for actual audio capture/playback, as `optionalDependencies` so non-voice users are unaffected
3. VAD + interruption handling — explicitly flagged as needing maintainer sign-off before merging, since it touches abort/stream logic used by everyone

### What I'm explicitly not proposing yet

- Wake word detection (heavy native dependency / licensing questions — separate discussion)
- OAuth-based auth support for live sessions (unclear if that's even possible with current SDKs — needs input from someone with visibility into that)

### Expected Outcomes

[...]

PR #708, opened 2026-07-27
feat(voice): add standalone local audio I/O plugin for Realtime Voice Mode[ PR 2/6 ] 

RONAK-AI647:
## Description

This PR introduces the foundational audio engine for the upcoming Realtime Voice Mode (PR2). It adds a lightweight, standalone Node.js plugin **(@nanocollective/nanocoder-voice)** that handles all local audio capture, playback, and processing.
For #622 

## Type of Change

- [x] New feature

## Changes

- **Capture & Playback**: Cross-platform microphone and speaker support using sox (rec/play), with automatic fallbacks for native macOS (afplay) and Linux (aplay).

- **Local Speech-to-Text (STT):** Safely spawns and streams whisper-cli (whisper.cpp) to transcribe audio natively without memory leaks.
 
- **Local Text-to-Speech (TTS**): Feeds text into piper for rapid, on-device voice synthesis.

- **Safety First**: Implements strict duration limits, stdout truncation, and process timeouts (e.g., 60s for STT, 30s for TTS) to ensure the CLI never hangs indefinitely.

- **Zero-Friction Tests**: Includes a mocked AVA test suite that verifies the spawned streams and timeouts without requiring actual audio hardware on CI machines.


akramcodez:
@RONAK-AI647 Hey! I went through the PR and overall the plugin structure looks really solid. The package isolation, optional dependency setup, workspace registration, and hermetic tests all look good. I am requesting changes before merge. The Windows recording bug and the missing `playAudio` timeout are the two items that need to be resolved; the others are important API decisions that are easier to fix now than to undo later.

**1. Blocker: Windows microphone args are in the wrong order**
`plugins/voice/src/microphone.ts` line 23: `args.unshift('-d')` is called after the duration args have already been pushed into the array. When a duration is provided on Windows, the final command becomes `sox -d -q test.wav trim 0 5`. Sox interprets `-q` as a second input file rather than a global flag, which will cause recording to fail. The correct order is `sox -q -d test.wav trim 0 5`. Please build the Windows args as a separate array rather than mutating the shared one. There is also no test covering the Windows path with a duration, so please add one.

**2. `detached: true` without `proc.unref()` is misleading**
`plugins/voice/src/stt.ts` line 23 and `tts.ts` line 20: `detached: true` without a matching `proc.unref()` call does not prevent the event loop from waiting for the child to exit. The option creates a new process group but has no practical effect here. More importantly, when the timeout fires, `proc.kill('SIGTERM')` sends the signal only to the direct child PID and not its entire process group, so any sub-processes spawned by `whisper-cli` (which it commonly creates when loading model shards) will be left running. Since daemon behavior is not needed in PR 2, the simplest fix is to remove `detached` entirely.

**3. `playAudio` has no timeout and no cancellation**
`plugins/voice/src/speaker.ts`: `transcribeAudio` has a 60 s timeout and `synthesizeSpeech` has a 30 s timeout, but `playAudio` has neither. If `afplay` or `aplay` hangs because the audio device is locked, `playAudio` will never resolve or reject. This becomes a hard blocker in PR 5 since barge-in requires cancelling active playback. Please add a `timeoutMs` parameter consistent with the pattern in `stt.ts` and `tts.ts`, and add a test for the timeout path.

**4. No `AbortSignal` on any public function**
`microphone.ts`, `speaker.ts`, `stt.ts`, `tts.ts`: When `useVoice` is implemented in PR 3, it will need to cancel in-flight transcription or synthesis when the user interrupts. Without `AbortSignal` on the function signatures now, PR 3 will require a breaking change to a public package API. I recommend adding `signal?: AbortSignal` as an optional last parameter to all four functions while the API is still being designed. The PR 2 implementation just needs to register an abort listener that calls `proc.kill('SIGTERM')`.

**5. Minor: `manual-test.md` has no actual test steps**
`plugins/voice/manual-test.md`: The file explains prerequisites but stops there. Please add at least one concrete command using the `REC_CMD`, `WHISPER_CMD`, and `PIPER_CMD` environment variable overrides so contributors can verify the plugin without real hardware, along with expected outputs or file artifacts for each step.

Once these are addressed I am happy to take another look. The architecture is clean and the direction is right.

RONAK-AI647:
@akramcodez addressed them, take a look !!

PR #712, opened 2026-07-28
fix: improve CLI and Node binary discovery for NVM/Volta users

akramcodez:
## Description

The VS Code extension previously failed to locate the globally installed `nanocoder` CLI for users relying on Node version managers like NVM or Volta. This happened because the VS Code Remote extension host often inherits a minimal `PATH` where NVM/Volta is not initialized. Furthermore, even if the user manually configured `nanocoder.cliPath`, the CLI process would crash because `#!/usr/bin/env node` could not find `node` in that same minimal `PATH`.

This PR fixes both issues by:
1. Adding a fallback discovery step that explicitly checks common global installation directories (`~/.nvm`, `~/.npm-global`, `~/.volta`, `~/.local/share/fnm`, etc.).
2. Automatically prepending the discovered CLI's directory to the child process `PATH` environment variable. This ensures that `env node` correctly resolves the Node binary installed alongside it.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Changeset

*(Note: Run `pnpm changeset` and mark this as a `patch` bump for `nanocoder-vscode`)*

### Automated Tests

- [x] New features include passing tests in `.spec.ts/tsx` files *(Added `cli-discovery.spec.ts`)*

### Manual Testing

- [ ] Tested with Ollama
- [ ] Tested with OpenRouter

## Checklist

- [x] Self-review completed


will-lamerton:
Hey @akramcodez - could you include a changeset for this? :)

will-lamerton:
Thanks for this @akramcodez - the direction is right and the discovery fallback itself works. A few things need addressing before this can go in.

## The fallback discovery (step 4) looks good

Once `which nanocoder` fails under a minimal PATH, the new step correctly finds the binary in `~/.nvm/versions/node/*/bin`, `~/.volta/bin`, fnm's default alias, and the usual system prefixes. The newest-first version sort is a sensible heuristic and the `readdirSync` is properly guarded.

Two gaps worth filling while you're here:

- pnpm's global bin is missing (`~/Library/pnpm` on macOS, `~/.local/share/pnpm` on Linux). Worth adding given this project ships with pnpm. `~/.bun/bin` too.
- On Windows the fallback paths look for `nanocoder`, not `nanocoder.cmd`, so step 4 is effectively unix-only. Fine if that's intentional, but the `where.exe` branch above it implies Windows is in scope.

## The PATH prepend needs guarding

`acp-process-manager.ts:60` fires for every `cliPath`, including the local-dev form where `cliPath` is `node /abs/path/dist/cli.js`. `path.dirname` of that string is `"node /abs/path/dist"`, so a nonsense entry containing a space gets prepended to PATH. Not fatal, but the dev path gets none of the intended benefit. Please guard with `!cliPath.startsWith('node ')`.

More importantly, prepending unconditionally can shadow the correct node. If discovery lands on `/usr/local/bin/nanocoder`, this puts `/usr/local/bin` ahead of the login-shell PATH, so an old `/usr/local/bin/node` beats the user's nvm node. That's exactly the failure mode the comment three lines above it exists to prevent (added in 3f9a505a). Suggest only prepending when the CLI's own directory actually contains node:

```ts
const cliDir = path.dirname(cliPath);
if (fs.existsSync(path.join(cliDir, 'node'))) {
  env.PATH = env.PATH ? `${cliDir}${path.delimiter}${env.PATH}` : cliDir;
}
```

Related: this also means `~/.npm-global/bin` in the fallback list never gets the node fix, since npm's global bin dir doesn't contain a node binary. A user in exactly that configuration still hits the crash the PR describes. Worth calling out as a known limitation if you don't want to solve it here.

## The new test never runs

AVA's `files` glob in the root `package.json` is `source/**/*.spec.ts`, so `plugins/vscode/src/cli-discovery.spec.ts` is outside the suite entirely. Biome also excludes `plugins`, and root `tsc --noEmit` only includes `source/**/*`, so nothing in CI touches this file.

Running it directly fails:

```
$ npx ava plugins/vscode/src/cli-discovery.spec.ts
  Error: Cannot find module 'vscode'
  Require stack: .../plugins/vscode/src/cli-discovery.ts
```

`cli-discovery.ts` imports `vscode`, which only resolves inside the extension host. So the "new features include passing tests" checkbox isn't met, and "all existing tests pass" holds only because this file is never picked up.

Options: extract the pure discovery logic into a vscode-free module that can live under `source/` and be tested there, or drop the test file. There's also no coverage for the PATH-prepend half of the change, which is the part carrying the regression risk above.

## Changeset

`.changeset/config.json` has `"ignore": ["nanocoder-vscode"]`, so this changeset produces nothing:

```
$ npx changeset status
🦋  info NO packages to be bumped at patch
```

The vsix ships bundled inside the main package, so if this should appear in release notes it wants a changeset on `@nanocollective/nanocoder` instead. Also, `Closes #712` in the changeset body refers to this PR's own number rather than an issue.

## Minor

- Three inline `require('path')` calls in a file that already uses ESM imports (`import * as os from 'os'` at the top). It only works because esbuild emits CJS, and would break if the bundle format ever changes. A top-level `import * as path from 'path'` is cleaner and matches the file.
- `if (cliPath && cliPath.includes(...))` - `cliPath` is already narrowed to `string` by the early return at line 46, so the truthiness check is dead.

## Summary

Happy to merge once the PATH prepend is guarded (both the `node <script>` form and the node-shadowing case), the test either works or comes out, and the changeset is retargeted. The discovery fallback itself is a genuine improvement and I'd like to get it in.

PR #714, opened 2026-07-29
fix: show output for user-typed ! bash commands in the transcript.

cschlick:
## Description

Currently, !commands do not show their output. This enables something like `!git status` or `!ls` to give human readable feedback instead of just going to the model.

Root cause: the output section in `BashProgress` depends on `!state.isComplete`, so output only exists as the live preview while a command runs. Fast commands finish before a frame renders, and the completed card shows only Command / Status / Tokens. Ctrl+O can't reveal it later .

Fix: a new `showOutput` prop on `BashProgress` (default `false`) renders the captured output on completion . Output is truncated to `BASH_OUTPUT_DISPLAY_LINES = 20` (arbitrarily chosen) with an `… (+N earlier lines)` note, mirroring the pattern in `streaming-message.tsx`. Only the user-typed `!` path opts in, the model-invoked `execute_bash` keeps its compact display in both compact and expanded modes.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Changeset

### Automated Tests

- [x] All existing tests pass (`pnpm test:all` completes successfully)*

*One unrelated spec fails identically on a clean `main` checkout:
`settings-tabs.spec.tsx`

### Manual Testing

- [x] Tested with Ollama
- [x] Tested with OpenRouter
- [x] Tested with OpenAI-compatible API (ChatGPT/Codex provider)

## Checklist

- [x] Self-review completed
- [x] Documentation updated (I don't think anything is needed)
- [ ] Appropriate logging added — N/A, doesn't affect any existing logging


will-lamerton:
Thanks for this PR @cschlick - feel free to add yourself as a contributor to our website via a PR which I will approve :)

PR #718, opened 2026-07-30
feat: implement image upload and preview functionality in the chat interface

akramcodez:
## Description
This PR introduces support for image uploads and pasting in the Nanocoder VS Code extension chat panel, allowing users to easily send multimodal messages (text + images) to the AI assistant. 

### Screen shots
image
image

## Key Changes
- **UI Updates (`chat-panel.html`, `chat-panel.css`, `chat-panel.js`)**: 
  - Added a `+` icon button to open a file dialog for selecting images.
  - Added clipboard event listeners to allow pasting images directly into the chat (`Ctrl+V`).
  - Implemented an image preview area above the chat input where users can see and remove attached images before sending.
  - Added an expandable image modal that opens when a user clicks on an image thumbnail in the chat.
- **Webview to Extension Communication (`chat-webview-provider.ts`, `webview-protocol.ts`)**: 
  - The `submitMessage` payload now includes an `images` array containing base64 data and MIME types.
- **ACP Integration (`acp-client.ts`, `acp-agent.ts`)**: 
  - Modified the `prompt` method to accept images and structure them as standard ACP `image` content blocks.
  - Handled the processing of these multimodal prompts on the backend agent side.

## Testing Instructions
1. Build the extension and start the nanocoder backend.
2. Open the Nanocoder chat panel in VS Code.
3. Try pasting an image using `Ctrl+V` or clicking the `+` icon to upload one.
4. Verify the image preview appears above the input box and can be removed.
5. Submit a prompt with the image and verify that the AI responds appropriately to the visual content.
6. Click on the sent image in the chat history to verify the modal expands correctly.

*(Note: Restoring images from historical sessions upon reload is deferred to a future PR).*


will-lamerton:
Hey @akramcodez - nice work wiring the image preview, paste handler, and modal. The UI additions in `chat-panel.html` and `chat-panel.js` feel polished, and the protocol message shape (`WebviewMessageSubmitMessage.images`) is reasonable. The webview stops forwarding the `image` content to the server cleanly via `acpClient.prompt(text, images)`. But a few blockers and loose ends need to be addressed before this can be merged. See below:

**Required changes**

- **Dead branch in `replaySessionHistory`** - `source/acp/acp-agent.ts:550` adds an `Array.isArray(message.content)` branch, but `Message.content` is typed as `string` in `source/types/core.ts:32` and the only writer at `source/acp/acp-agent.ts:290-295` stores the image block in a separate `images: ImageAttachment[]` field, never inside `content`. So the new `for (const part of message.content)` replay is unreachable for the live code path. Either delete the dead branch, or change `AcpAgent.prompt` to actually store content as an array of `ContentBlock` parts (and update `Message.content` to `string | ContentBlock[]` plus the downstream converter in `source/ai-sdk-client/converters`) - but that is the larger refactor you almost certainly want to avoid in this PR. Pick the delete option.

- **Field-name mismatch in `acp-agent.ts`** - the new branch reads `part.mimeType` (lines 562, 566), but `ImageAttachment` in `source/types/core.ts:26` defines the field as `mediaType`. The same `mimeType` vs `mediaType` mismatch exists in the inbound webview handler at `plugins/vscode/media/chat-panel.js:587`. Since the SDK uses `mimeType` on the wire and the internal type uses `mediaType`, the mapping must be explicit (e.g. `mimeType: part.mimeType ?? part.mediaType`) - otherwise images in the replayed/echo code path will be reconstructed with `undefined` MIME and the `` will render broken.

- **Optimistic user message + server echo = duplicate image** - `plugins/vscode/media/chat-panel.js:314` immediately appends the user message with its images, and then `source/acp/acp-agent.ts`’s `replaySessionHistory` (and the new image branch) fires a `user_message_chunk` with the same image, which the webview handler at `chat-panel.js:584-590` re-appends. The user will see two copies of the same image when an image-only prompt is sent. Either skip the optimistic append when images are present, or have the webview de-dup `user_message_chunk` echoes against the optimistic message. The fastest fix is to drop the optimistic append whenever `imagesToSubmit` is non-empty.

- **Add tests for the new code path** - add at least one test covering `acpContentToUserMessage` with a mixed text+image prompt (the existing `source/acp/acp-content.spec.ts:40` covers the parser, but please add one that verifies `AcpAgent.prompt` routes the images through to the conversation), and add a unit test for `chat-webview-provider._handlePrompt` invoking `acpClient.prompt` with the `images` argument. As is, the new JS path is uncovered, and the AC/backend branches aren’t exercised either.

- Missing changeset. Please add one.

- Currently Ava unit tests are failing as is semgrep.

**Suggestions**

- `plugins/vscode/media/chat-panel.js:179-205` - The `addImageBtn` click handler kicks off `imageUpload.click()`, but the button doesn't disable while the `FileReader` is in flight, and pasting 50 large images from the clipboard will decode all of them in parallel with no size cap. Consider a max file size (e.g. 10MB) and a max attach count, and disable the button while reads are pending.

- `plugins/vscode/media/chat-panel.js:237-241` - The accept attribute is `image/*`, but the server (`SUPPORTED_IMAGE_MEDIA_TYPES` in `source/acp/acp-content.ts:23`) only accepts png/jpeg/gif/webp. Surface a friendly error in the webview (toast) when the agent strips the image rather than silently dropping it.

- `plugins/vscode/media/chat-panel.html:9` - The `https:` source in `img-src` is a wildcard. If you only need it for `` tags that the markdown renderer might emit, you can scope it to the same `cspSource` mechanism used for `style-src`, but leaving it broad is acceptable for a chat UI.

- `plugins/vscode/media/chat-panel.js:447` - `appendMessage(message.content, 'agent')` is now called with `content` (a string) but the new signature is `appendMessage(content, role, images = undefined)`. That still works, but worth noting that `'agent'` messages will never receive images server-side — fine, just verify nothing is passing them.

- `source/acp/acp-agent.ts:556-573` - Whenever the dead branch is removed, also delete the `part.type === 'image' && part.data` branch and the now-redundant `part.mimeType` references; the single existing `user_message_chunk` for text is sufficient given current data shape.

- `plugins/vscode/src/acp-client.ts:278-289` - The `promptData` `any[]` could be `ContentBlock[]` from `@agentclientprotocol/sdk` to keep the types tight.

akramcodez:
Thanks @will-lamerton for the detailed review! I've addressed the required feedback:

- removed the dead replay branch and kept Message.content as string
- fixed the duplicate image behavior after validating the prompt flow
- added the requested backend and webview tests
- tightened the ACP typing, added the changeset, and implemented the suggested upload constraints/error handling

One small difference from the suggested duplicate fix: I initially tried skipping the optimistic append, but during testing the user's image message disappeared because the backend doesn't emit a user_message_chunk during the active prompt flow. I kept the optimistic append and removed the obsolete replay echo path instead, which resolves the duplication while preserving immediate feedback.

I'd appreciate another look when you have a chance. Thanks!

will-lamerton:
Thanks for the quick follow-up, @akramcodez. I re-fetched the branch (now at `5c71f7a1`) and re-verified each prior concern against the latest code with the tests run locally. The server-side cleanup is genuinely good — the dead branch is gone, the media-type mapping is explicit and safe, and the new ACP tests pass (38/38). Two items still need attention before this is ready, plus one process concern about the new test.

**Required changes**

- **Duplicate image rendering on resume/reconnect — still not addressed.** The follow-up commit did not touch the render path. The webview still optimistically appends the user message *with* images at `plugins/vscode/media/chat-panel.js:360` (`appendMessage(text, 'user', imagesToSubmit)`), and `handleAcpUpdate` still unconditionally re-appends any `user_message_chunk` at `chat-panel.js:631-636`. For the live prompt this is currently benign — I confirmed `user_message_chunk` is only emitted by `replaySessionHistory` (`source/acp/acp-agent.ts:546`), not during a live turn. But on `loadSession`/resume, the server replays the user's text via that chunk and the webview renders a second bubble while the optimistic image bubble remains, so the thread duplicates. Now that the dead `Array.isArray` image-echo branch is (correctly) removed, the cleanest fix is to de-dup on the client: track that the just-submitted message is pending and skip the first matching `user_message_chunk` text, or suppress the optimistic append once the session is known to be resume-capable. As-is, resume-after-image-upload shows the message twice.

- **The new vscode test is never run.** `plugins/vscode/src/chat-webview-provider.spec.ts` exists and type-checks, but the `plugins/vscode` package has no test runner at all — no `ava`/`vitest`/`mocha`/`jest` in `devDependencies`, and no `test` script in `package.json` (only `build`/`watch`/`lint`). So the coverage added for `_handlePrompt` → `acpClient.prompt` images is dead weight that will silently rot. Either wire up a runner + `test` script (and add it to CI), or drop the spec. I verified `npx tsc --noEmit` passes in both the root and `plugins/vscode`, so it at least compiles — but a test that can't execute isn't coverage.

- **Fix merge conflicts**

**Suggestions** (non-blocking)

- The friendly-error path for unsupported types is now duplicated in two places: the client pre-filters via `SUPPORTED_TYPES` (`chat-panel.js:212-217`) and the server notes omissions (`acp-content.ts:68-72`). Consider letting the server message be the single source of truth so the lists can't drift (e.g. client supports GIF but a future server doesn't).
- CSP `img-src` still includes a broad `https:` wildcard (`plugins/vscode/media/chat-panel.html:9`). Since images are rendered as `data:` URLs, you could drop `https:` from `img-src` to tighten it.
- `openImageModal` now guards `src.startsWith('data:image/')` (`chat-panel.js:301-308`) — good. Same guard could be applied to the `appendMessage` image path for defense in depth, though `mimeType` is already whitelist-validated upstream.

PR #720, opened 2026-07-31
feat: add context attachment functionality with UI for file/folder chips and drag-and-drop support

akramcodez:
### Overview
This PR resolves several UI/UX issues in the VS Code extension's chat panel, specifically addressing limitations with drag-and-drop file attachments and aggressive webview caching.

### Screen Shots
image

### Key Changes

1. **Native Attach Button**
   - **Problem:** Dragging files from the VS Code explorer into the webview iframe was being intercepted by VS Code's outer "Editor Drop" widget, blocking the webview's internal drop listeners.
   - **Solution:** Added a native `Attach` (paperclip) button next to the send button that delegates to `vscode.window.showOpenDialog`. This allows users to safely and natively attach multiple files and folders to their chat context.
   - Files changed: `chat-panel.html`, `chat-panel.js`, `webview-protocol.ts`, `chat-webview-provider.ts`.

2. **Webview Cache Busting**
   - **Problem:** Updating `chat-panel.js` or `chat-panel.css` required users to completely restart VS Code because the webview aggressively cached the `media/` URIs.
   - **Solution:** Injected a `?t=timestamp` query parameter into the resolved webview URIs in `chat-webview-provider.ts` so that reloading the webview accurately serves fresh assets.

3. **Dropdown Label Fix**
   - **Problem:** A previous UI refactor expected dropdown options to be objects (`{ label, value }`), but the extension host passes arrays of strings, resulting in blank text on the Provider/Model/Mode dropdown triggers.
   - **Solution:** Reverted the `setOptions` logic in `chat-panel.js` to correctly parse strings and automatically extract the display basenames (e.g., stripping paths).


will-lamerton:
Hey @akramcodez - this is really neat. Thanks for tackling it - the drag-and-drop interception problem and the stale-asset caching are real pain points, and the native attach button is a sensible workaround. I've reviewed all six files, and the extension host compiles cleanly (`tsc --noEmit` passes) with the new protocol messages wired up correctly. A few items below, one of which I think needs fixing before this ships.

**Required changes**

- `plugins/vscode/media/chat-panel.js:316` - The drop handler's URI parsing will produce wrong paths on Windows and mishandle spaces. `u.replace(/^file:\/\//, '')` turns `file:///C:/Users/foo/file.txt` into `/C:/Users/foo/file.txt` (leading slash breaks `fs.statSync` at `plugins/vscode/src/chat-webview-provider.ts:165`, so the drop silently no-ops - the exact failure mode the comment at line 168 describes). It also strips `file://` from UNC paths (`file://server/share/...`) leaving `server/share/...` instead of `\\server\share`. And `text/plain` fallback lines are run through the same `file://` strip + `decodeURIComponent`, so a dropped plain-text path containing `%20` gets corrupted. Suggest: only decode when the line actually starts with `file://`, strip with `/^file:\/\/\/?/`, and on Windows (`navigator.platform`/`navigator.userAgentData`) drop the leading slash before a drive letter. Also filter out comment lines (`#...`) which are legal in `text/uri-list`.

- Missing changeset. Please add one.

- Failing tests and semgrep - please take a look and hopefully make them green.

- UI improvement - can we have better tagged files UI? Currently this works, but doesn't look particularly polished:

Screenshot 2026-08-02 at 21 43 19


**Suggestions**

- `plugins/vscode/media/chat-panel.js:274-277` - `chip.innerHTML` interpolates `item.path` (into `title`) and `item.name` directly. These originate from `showOpenDialog`/drop data, so it's local and low-risk, but a filename containing `"` or `<` would break markup or inject into the webview DOM. `textContent`/`setAttribute` for the name and title would be safer and barely more code.
- `plugins/vscode/src/chat-webview-provider.ts:293` -`new Date().getTime()` cache-busts on every webview creation, so restores of a persisted sidebar view re-download JS/CSS even when unchanged. `Date.now()` is the idiom, but better still would be a build/version-derived value (e.g. extension version from `package.json`) so unchanged assets stay cached while upgrades bust reliably.
- `plugins/vscode/src/chat-webview-provider.ts:162-163` and `183-184` -`require('fs')`/`require('path')` inside the handlers works since esbuild bundles them, but the rest of the codebase (e.g. `acp-process-manager.ts:2-5`) uses top-level `import * as fs from 'fs'`. Hoisting these to imports (or at least module scope) would match convention and avoid re-resolving per dialog selection. (`_getHtmlForWebview` at line 286 already does the same `require` dance, so this is pre-existing style — feel free to treat as follow-up.)
- `plugins/vscode/media/chat-panel.js:115-116` - Stray blank lines left after removing `setValue`/`toggle`; minor tidy-up.

akramcodez:
Thanks @will-lamerton for the review! I've addressed the required feedback:

- fixed Windows/UNC URI parsing and text/uri-list handling
- resolved the CodeQL XSS issue by replacing innerHTML with safe DOM construction
- switched cache busting to a version-based key and hoisted fs/path imports
- added the changeset and polished the attachment chip UI

I also fixed the backend context injection path and improved chat history rendering for attached files/folders. I'd appreciate another look when you have a chance. Thanks!

image
image

will-lamerton:
Thanks for the updates @akramcodez - several of the previous points are genuinely addressed (details below), but I found two blockers, one of which will prevent the extension from compiling at all.

**Required changes**

- `plugins/vscode/src/chat-webview-provider.ts:2-3` and `5-6` - The PR adds a second `import * as fs from 'fs'; import * as path from 'path';` pair, but the base file already imports both at lines 5-6. The head file now has both pairs, which is a hard `TS2300: Duplicate identifier 'fs' / 'path'` compile error (verified locally with `tsc`). Delete the duplicated pair you added at lines 2-3. (CI's "Type Checks" only covers the root `source/` tsconfig and the VS Code build uses esbuild without typechecking, which is why this passed green - it still breaks the extension build for anyone running `tsc` on the plugin or in editor tooling.)
- `plugins/vscode/media/chat-panel.js:401` and `412` - The previous review's XSS concern was fixed for the composer chips (`renderChips` now correctly uses `textContent`/`setAttribute` at `chat-panel.js:274-277` - good), but the same pattern was reintroduced in `appendMessage`: both the `@[file]` and `<context path="...">` replacements interpolate `path` and `name` straight into `chipsHtml`, which is then assigned to `chipsContainer.innerHTML` at `chat-panel.js:429`. A filename containing `"` or `` breaks out of `data-path`/`title` and executes in the webview DOM (these are unsandboxed webviews). This is almost certainly what the open CodeQL "Client-side cross-site scripting" alert on this file is flagging. Build these chips with `document.createElement` + `textContent`/`setAttribute` like `renderChips` does, or escape the values before interpolation.
- Fix merge conflicts

**Suggestions**

- `plugins/vscode/media/chat-panel.js:22` — trailing whitespace on the blank line after `this.onChange = onChange;` (and a few other touched lines); a quick format pass would clean these up.
- `plugins/vscode/media/chat-panel.js:337` — `navigator.platform` is deprecated; `navigator.userAgent.includes('Windows')` is the more portable fallback if `userAgentData` is absent.

will-lamerton:
Hey @akramcodez - can you tackle the merge conflicts and then all good to go! :)

PR #732, opened 2026-08-01
feat(voice):  Add push-to-talk voice pipeline and VoiceStatusBar integration (PR 3/6)

RONAK-AI647:
## Description
This is PR 3 of 6 for Realtime Voice Mode (see #622 for the full plan). PR1 gave 
us the settings and UI. PR2 gave us the local recording/speech tools as a standalone plugin. This PR connects them together for the first time - this is the first PR where voice mode actually does something real.

## Type of Change

- [x] New feature

## Changes

- Hold `Ctrl+T` to talk: records your voice, turns it into text, sends it into 
  the chat exactly like you'd typed it, then reads the AI's reply back out loud.
- `/voice` now plays a short "Voice mode activated / deactivated" sound when you 
  toggle it, not just a text message.
- The little voice status box now shows real states (Idle / Listening / 
  Processing / Speaking) instead of always saying "Idle."
- If the voice plugin isn't installed, or recording/speech fails for any reason, 
  the app shows a clear error and keeps working normally -  it never crashes or 
  gets stuck.

## Important note

`Ctrl+T` push-to-talk here is meant as an internal way to prove the whole voice 
pipeline actually works end-to-end. It's not the final experience - the real goal 
is fully hands-free voice (no button needed), which is coming in PR4.

### DIFF
--- a/source/app/sections/interactive-app.tsx
+++ b/source/app/sections/interactive-app.tsx
@@ -16,6 +16,7 @@ import {useTerminalRows} from '@/hooks/useTerminalWidth';
 import {useTheme} from '@/hooks/useTheme';
 import {UIStateProvider} from '@/hooks/useUIState';
 import type {useUserMessageQueue} from '@/hooks/useUserMessageQueue';
+import {useVoice} from '@/hooks/useVoice';
 import type {useVSCodeServer} from '@/hooks/useVSCodeServer';
 import type {ImageAttachment} from '@/types/core';
 import type {RestoredInputDraft, SubmittedInputDraft} from '@/types/hooks';
@@ -82,6 +83,12 @@ export function InteractiveApp({
 		React.useState<SubmittedInputDraft | null>(null);
 	const [restoredDraft, setRestoredDraft] =
 		React.useState<RestoredInputDraft | null>(null);
+
+	const {state: voiceState, startStopRecording} = useVoice({
+		handleUserSubmit,
+		messages: appState.messages,
+		addToChatQueue: appState.addToChatQueue,
+	});
 
 	const handleToggleCompactDisplay = () => {
 		const expanding = appState.compactToolDisplay;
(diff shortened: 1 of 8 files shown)


RONAK-AI647:
Question about the piper/whisper dependency setup, plus an idea:

Right now piper and whisper.cpp are optional - voice mode toggles fine without  them, but to actually record and speak, someone has to manually install sox + whisper.cpp + piper themselves in a separate terminal first. That's real friction, especially for anyone non-technical who just wants to try voice mode.

The engineering reason they're optional in the first place: piper/whisper.cpp aren't small npm packages, they're real native tools with model files that can be hundreds of MB. If we made them required, every single Nanocoder install would get 
heavier -  even for the ~95% of people who'll never touch voice mode. That felt like the wrong tradeoff for a project built around staying lightweight.

I though of this Idea : keep them optional at install time, but the first time someone actually tries /voice or Ctrl+T and the tools aren't found, show a simple "Voice needs piper + whisper (~X MB) - install now? Y/N" right there in the app. If they say yes, Nanocoder installs everything for their OS automatically, with progress shown, no separate terminal or manual steps needed.

This keeps the default install light, but removes basically all the friction for anyone who actually wants voice. It'd need its own PR though -  this is a real, separate feature (OS detection, install logic, progress UI, error handling), not something to squeeze into PR3 at the last minute. Where do you think this should sit -  maybe as its own PR after PR6, or folded in as part of PR6's scope? what you think. @will-lamerton @akramcodez

RONAK-AI647:
Looked at how Gemini CLI handles a similar situation (their optional ripgrep  dependency, and keytar for credentials) - they use optionalDependencies with  silent fallback, no user prompt at all, and everything stays within npm (no  pip/Python anywhere, even for native tools). Our current optional-dependency  setup already follows that same pattern. The Y/N install prompt idea would go a  step further than what they do - since voice genuinely can't work at all without  the tools (unlike their case, where it just degrades to something slower), I  think a clear, one-time prompt is worth the extra step rather than silently  failing.

will-lamerton:
Hey @RONAK-AI647 - looking great and like your solution :)

PR #741, opened 2026-08-02
Fix update checks, cache presence, and network error classification

### LINKED ISSUE #737: [Bug] isNetworkError Checks Node Errno Codes on Wrong Property (error.name vs error.code)
## Description

The `isNetworkError` function checks `error.name` for Node.js system error codes such as `ECONNREFUSED`, `ENOTFOUND`, `ECONNRESET`, and `ETIMEDOUT`. In Node.js, these codes live on `error.code`, not `error.name`. The `name` property of a system error is always `"Error"`. As a result, `isNetworkError` never returns `true` for actual Node network failures - only for the loose message-string matches further down the condition.

The same problem affects `ETIMEDOUT` in `isTimeoutError`.

## Environment

- **OS**: Any
- **Node version**: Any (v18+)
- **Nanocoder version**: Latest
- **Provider**: Any network-facing provider (OpenRouter, OpenAI-compatible)
- **Model**: Any

## Steps to Reproduce

1. Configure a provider with an unreachable base URL.
2. Send a prompt and wait for the connection to be refused.
3. Inspect the `ErrorInfo` object returned by `createErrorInfo()`.
4. Observe `isNetworkError` is `false` even though the error is `ECONNREFUSED`.

## Expected Behavior

`isNetworkError` should return `true` when Node throws a system error with `error.code === 'ECONNREFUSED'` (or `ENOTFOUND`, `ECONNRESET`).

## Actual Behavior

`isNetworkError` returns `false` because it checks `error.name` which is always `"Error"` for system errors, not the errno code.

## Logs/Screenshots

N/A

## Additional Context

- Fix: replace `error.name === 'ECONNREFUSED'` with `(error as NodeJS.ErrnoException).code === 'ECONNREFUSED'` (and equivalents for `ENOTFOUND`, `ECONNRESET`, `ETIMEDOUT`).
- Note: This issue is intentionally kept small and is suitable for first-time contributors. If you're interested in working on it, please leave a comment below and ask to be assigned before opening a pull request.

### LINKED ISSUE #738: [Bug] BoundedMap.has() Returns false for Keys Storing undefined
## Description

`BoundedMap.has()` is implemented as `return this.get(key) !== undefined`. `get()` returns `undefined` for two distinct reasons: (1) the key does not exist, or (2) the key exists but its stored value is `undefined`. This means
`has()` returns `false` for a key that was explicitly set to `undefined`, violating the standard `Map` contract where `has()` only reflects key existence, not value truthiness.

## Environment

- **OS**: Any
- **Node version**: Any
- **Nanocoder version**: Latest
- **Provider**: N/A
- **Model**: N/A

## Steps to Reproduce

1. Open `source/utils/bounded-map.ts`.
2. Create a `BoundedMap` and call `map.set('key', undefined)`.
3. Call `map.has('key')`.
4. Observe it returns `false` despite the key being present.

## Expected Behavior

`map.has('key')` should return `true` when the key was set, regardless of whether the value is `undefined`.

## Actual Behavior

`map.has('key')` returns `false` because `get()` returns `undefined` and the implementation treats that as "key not found".

## Logs/Screenshots

N/A

## Additional Context

- Fix: rewrite `has()` to check the internal `Map` directly and perform a manual TTL check, rather than delegating to `get()`.
- This matches how the native `Map.prototype.has()` behaves.
- Note: This issue is intentionally kept small and is suitable for first-time contributors. If you're interested in working on it, please leave a comment below and ask to be assigned before opening a pull request.

### LINKED ISSUE #739: [Bug] Update Check Timestamp Written Even When Version Fetch Failed
## Description

In `checkForUpdates()`, `updateLastCheckTime()` is called immediately after `fetchLatestVersion()` regardless of whether the fetch succeeded or returned `null`. When `fetchLatestVersion()` fails (e.g., offline, npm registry down), it returns `null` - but the "last checked" timestamp is still written to preferences. This causes the app to skip the next scheduled update check entirely, silently preventing users who were temporarily offline from ever
seeing an update notification.

## Environment

- **OS**: Any
- **Node version**: Any
- **Nanocoder version**: Latest
- **Provider**: Any
- **Model**: Any

## Steps to Reproduce

1. Disconnect from the internet.
2. Launch Nanocoder (or trigger an update check).
3. Observe that `preferences.lastUpdateCheck` is updated despite the fetch failing.
4. Reconnect to the internet.
5. Wait for the next update check interval - no update notification appears
   because the timestamp was already written.

## Expected Behavior

The "last checked" timestamp should only be updated when a valid version string is successfully returned from the npm registry.

## Actual Behavior

The timestamp is updated unconditionally, even when `fetchLatestVersion()` returns `null` due to a network error.

## Logs/Screenshots

N/A

## Additional Context

- Fix: wrap `updateLastCheckTime()` in an `if (latestVersion !== null)` check.
- Note: the existing `catch` block on line 176 already calls `updateLastCheckTime()` intentionally to prevent hammering on repeated crashes - that behavior is correct and should remain unchanged.
- Note: This issue is intentionally kept small and is suitable for first-time contributors. If you're interested in working on it, please leave a comment below and ask to be assigned before opening a pull request.

RealBhupesh:
## Summary

- Preserve `lastUpdateCheck` when the npm registry fetch fails.
- Make `BoundedMap.has()` report true for keys whose stored value is `undefined` while retaining TTL expiry behavior.
- Read Node.js errno codes from `error.code` for network and timeout classification.

## Issues

Closes #739  
Closes #738  
Closes #737

## Validation

- `pnpm exec ava source/utils/bounded-map.spec.ts source/utils/error-formatter.spec.ts source/utils/update-checker.spec.ts` - 73 tests passed
- `pnpm test:types`
- `pnpm test:format`
- `pnpm test:lint`
- `pnpm test:knip`
- `pnpm build`
- `pnpm exec ava source/cli-integration.spec.ts` - 7 tests passed

A full AVA run was also attempted. After building, the focused CLI integration tests passed; the full run still reports one unrelated settings-tabs assertion.

### DIFF
--- a/.changeset/fix-update-cache-network-guards.md
+++ b/.changeset/fix-update-cache-network-guards.md
@@ -1,0 +1,5 @@
+---
+"@nanocollective/nanocoder": patch
+---
+
+Fixed update checks incorrectly recording a successful check after a registry fetch failure, corrected `BoundedMap.has()` for entries whose value is `undefined`, and restored network-error classification for Node.js errno codes. Closes #739, #738, and #737.
--- a/source/utils/bounded-map.spec.ts
+++ b/source/utils/bounded-map.spec.ts
@@ -23,6 +23,15 @@ test('BoundedMap - has() method', t => {
 	t.false(map.has('nonexistent'));
 });
 
+test('BoundedMap - has() returns true for an existing undefined value', t => {
+	const map = new BoundedMap<string, undefined>();
+
+	map.set('key', undefined);
+
+	t.true(map.has('key'));
+	t.false(map.has('missing'));
+});
+
 test('BoundedMap - delete() method', t => {
 	const map = new BoundedMap<string, number>();
 
--- a/source/utils/bounded-map.ts
+++ b/source/utils/bounded-map.ts
@@ -90,7 +90,21 @@ export class BoundedMap<K, V> {
 	 * Check if key exists and is not expired
 	 */
 	has(key: K): boolean {
-		return this.get(key) !== undefined;
+		const entry = this.map.get(key);
+		if (!entry) {
+			return false;
+		}
+
+		// Check TTL if configured
+		if (this.ttl !== undefined && this.ttl > 0) {
+			const age = Date.now() - entry.timestamp;
+			if (age > this.ttl) {
+				this.map.delete(key);
+				return false;
+			}
+		}
+
+		return true;
 	}
 
 	/**
--- a/source/utils/error-formatter.spec.ts
+++ b/source/utils/error-formatter.spec.ts
@@ -1,5 +1,5 @@
 import test from 'ava';
-import {formatError} from './error-formatter';
+import {createErrorInfo, formatError} from './error-formatter';
 
 console.log(`\nerror-formatter.spec.ts`);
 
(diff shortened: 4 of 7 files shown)


will-lamerton:
Thanks for this PR @RealBhupesh  - feel free to add yourself as a contributor to our website via a PR which I will approve :)

PR #742, opened 2026-08-03
Fix wide terminal content width

### LINKED ISSUE #595: Hardcoded 120-column max width ignores large terminals

RealBhupesh:
## Summary

- Increase the bounded terminal content width from 120 to 200 columns.
- Reuse the centralized width constant in the responsive terminal hook.
- Update the regression test and add a patch changeset.

## Why

Terminals wider than 120 columns were unnecessarily constrained even though the layout can use more horizontal space. A 200-column cap covers the reported wide-terminal case while keeping a deliberate upper bound and the existing 40-column minimum.

Closes #595

## Validation

- `pnpm exec ava source/hooks/useTerminalWidth.spec.tsx`
- `pnpm exec biome check source/constants.ts source/hooks/useTerminalWidth.tsx source/hooks/useTerminalWidth.spec.tsx`
- `pnpm run test:types`


RealBhupesh:
CI follow-up: the required unit-test job was failing on a stale baseline assertion in `source/app/components/settings-tabs.spec.tsx`.

The current settings UI exposes `Appearance`, `Input`, `Behavior`, `Providers`, and `Advanced` after the settings consolidation, while the test still expected the removed `Display` tab. I updated the assertion to match the current tab contract. This is a shared baseline-test correction and does not change the feature scope of this PR.

The full local AVA suite now passes with 6,254 tests passed and 1 skipped. The fix has been pushed to this branch, and GitHub Actions is rerunning.

RealBhupesh:
Update: all required GitHub Actions checks are now passing, including unit tests and coverage. The PR is technically mergeable; the remaining requirement is maintainer review and approval.

PR #744, opened 2026-08-03
fix: preserve multibyte input in alternate-screen TUI

### LINKED ISSUE #532: [Feature] korean
export LANG=ko_KR.UTF-8
export LC_ALL=ko_KR.UTF-8

please check

RealBhupesh:
## Summary

- Decode alternate-screen stdin with a stateful UTF-8 decoder so multibyte characters are preserved across `data` chunk boundaries.
- Keep the existing mouse-sequence filtering and wheel-event handling unchanged after decoding.
- Add regression coverage that splits Korean input inside a character, plus the current settings-tab contract needed by the upstream test suite.

Closes #532

## Root cause

The alternate-screen input proxy converted every `Buffer` chunk with `Buffer#toString('utf8')`. When a terminal or IME split a multibyte character across chunks, the incomplete byte sequence was replaced before the next chunk arrived. The decoder now retains incomplete sequences and emits the character only after all bytes are available.

## Validation

- `pnpm exec ava source/utils/terminal-mouse.spec.ts source/app/components/settings-tabs.spec.tsx`
- `pnpm test:ava` (6,255 passed, 1 skipped)
- `pnpm test:format`
- `pnpm test:lint`
- `pnpm test:types`
- `pnpm build`

### DIFF
--- a/.changeset/fix-korean-terminal-input.md
+++ b/.changeset/fix-korean-terminal-input.md
@@ -1,0 +1,5 @@
+---
+"@nanocollective/nanocoder": patch
+---
+
+Fixed multibyte terminal input being corrupted when an alternate-screen stdin chunk split a UTF-8 character, which could affect Korean and other IME input.
--- a/source/app/components/settings-tabs.spec.tsx
+++ b/source/app/components/settings-tabs.spec.tsx
@@ -97,7 +97,7 @@ test.beforeEach(() => {
 	resetPreferencesCache();
 });
 
-test('renders all four category tabs with Appearance selected first', async t => {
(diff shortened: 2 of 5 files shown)


RealBhupesh:
Update: all required GitHub Actions checks are passing, including CodeQL, unit tests, coverage, type checks, linting, formatting, and the build. The PR is technically mergeable; the remaining step is maintainer review and approval.

will-lamerton:
Nice work @RealBhupesh :)

RealBhupesh:
> Nice work @RealBhupesh :)

Thank you so muchh ❤️

PR #773, opened 2026-08-03
fix: drop toLocaleString() thousands separators from LLM-facing tool output

### LINKED ISSUE #768: toLocaleString() thousands separators leak into LLM-facing tool results
## Problem

`toLocaleString()` is used when building strings that go into tool results, not just display output. Examples:

- `source/tools/read-file.tsx:53` `Size: ${size.toLocaleString()} bytes`
- `source/tools/read-file.tsx:66-67` line count and estimated tokens
- `source/tools/read-file.tsx:133-135` the metadata response
- `source/tools/list-directory.tsx:162` per-entry byte sizes

`12,345` costs more tokens than `12345` and carries no additional meaning for a model. Thousands separators are a human-readability affordance.

This is correct for the Ink display layer, which is why the pattern exists. It is just wrong on the path that feeds the LLM.

## Fix

Use plain numbers in any string returned from a tool `execute`, and keep `toLocaleString()` in the formatter components that render to the terminal.

Worth a quick grep for other instances beyond the ones listed, since the two paths are easy to confuse. The distinction to apply: anything inside a `Formatter` React component is display and should stay; anything inside an `execute` function is LLM payload and should not.

Mechanical and low-risk, but it does require care to only change the LLM-facing path and leave the display path alone. Good way to get familiar with how nanocoder separates tool execution from tool rendering.

MsfPablo:
## Summary

Fixes #768. `toLocaleString()` was leaking comma thousands-separators into strings returned to the model as tool results, not just into the terminal display. Comma separators cost extra tokens and carry no meaning for the model — this was flagged as good-first-issue material by the maintainer.

Fixed the LLM-facing (`execute`/`validator`) paths:
- `source/tools/read-file.tsx` — metadata output (size, lines, tokens), large-file metadata output, and the minified/binary-content validator error message
- `source/tools/list-directory.tsx` — per-entry byte size in the directory listing
- `source/utils/file-content-loader.ts` — binary-file and large-file metadata shown when a file is loaded via `@file` mention (this content is inlined into the chat message sent to the model)

Left `toLocaleString()` in place everywhere it's actually display-only, per the issue's own guidance (`Formatter` components vs `execute` functions):
- The Ink `<Text>` formatter components in `read-file.tsx`, `list-directory.tsx`, `search-file-contents.tsx`, `find-files.tsx` — these render to the terminal, never to the model
- `compact-handler.ts`, `context-max-handler.ts`, `auto-compact.ts` — checked these too since they also format token counts, but their formatted strings only go to `onAddToChatQueue` (local terminal display); the actual conversation array passed to the model (`setMessages(...)`) doesn't include them, so they're out of scope

## Test plan

- `pnpm run test:ava` on the three affected spec files — 84 tests pass
- One existing test (`read-file.spec.tsx`, minified-content validator) asserted the old comma-formatted output (`/15,000 characters/`); updated it to match the fixed output (`/15000 characters/`) since that was testing the exact behavior this issue asks to remove

### DIFF
--- a/.changeset/tolocalestring-llm-payload.md
+++ b/.changeset/tolocalestring-llm-payload.md
@@ -1,0 +1,5 @@
+---
+"@nanocollective/nanocoder": patch
+---
+
+Dropped `toLocaleString()` thousands-separators from strings returned to the model (`read_file`'s metadata output and validator error, `list_directory`'s per-entry size, and `@file`-mention metadata). Comma separators cost extra tokens without adding meaning for the model. Left them in place in the terminal display components, where they're actually useful.
--- a/source/tools/list-directory.tsx
+++ b/source/tools/list-directory.tsx
@@ -158,9 +158,7 @@ const executeListDirectory = async (
 							? '@'
 							: '';
 				const displayPath = recursive ? entry.relativePath : entry.name;
-				const sizeStr = entry.size
-					? ` (${entry.size.toLocaleString()} bytes)`
-					: '';
+				const sizeStr = entry.size ? ` (${entry.size} bytes)` : '';
 				output += `${displayPath}${suffix}${sizeStr}\n`;
 			}
 		}
--- a/source/tools/read-file.spec.tsx
+++ b/source/tools/read-file.spec.tsx
@@ -585,7 +585,7 @@ test.serial(
 			t.false(result.valid);
 			if (!result.valid) {
 				t.regex(result.error, /minified or binary content/);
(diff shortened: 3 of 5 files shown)


akramcodez (commented):
LGTM another great PR @MsfPablo

will-lamerton:
Thanks for this PR @MsfPablo  - feel free to add yourself as a contributor to our website via a PR which I will approve :)

PR #774, opened 2026-08-03
fix: format search_file_contents results grep-style, one line per match

### LINKED ISSUE #766: search_file_contents wastes tokens on multi-line result formatting
## Problem

`source/tools/search-file-contents.tsx:83-88` formats each match across three lines:

```ts
for (const match of matches) {
    output += `${match.file}:${match.line}\n`;
    output += `  ${match.content}\n\n`;
}
```

Producing:

```
Found 12 matches:

source/foo.ts:42
  const x = 1

source/bar.ts:11
  const y = 2
```

The extra newline, the 2-space indent, and the blank line between matches are all formatting the model does not need.

## Fix

Standard grep format, one line per match:

```
source/foo.ts:42:const x = 1
source/bar.ts:11:const y = 2
```

Shorter, equally parseable, and it matches what models have seen most in training. Saves roughly 3-4 tokens per match, so about 100 tokens on a default 30-result search (`DEFAULT_SEARCH_RESULTS`, `source/constants.ts:48`).

Small on its own, but it is free and completely lossless.

Note `contextLines` is supported (`source/tools/search-file-contents.tsx:75`), so whatever format is chosen needs to still read clearly when context lines are present. Standard grep uses `-` as the separator for context lines rather than `:`, which would work here too.

Single function, one loop. Tests in `source/tools/search-file-contents.spec.tsx` will need their expected output updated. Run with `pnpm run test:ava source/tools/search-file-contents.spec.tsx`.

MsfPablo:
## Summary

Fixes #766. `executeSearchFileContents` formatted each match across three output lines (a `file:line` header, an indented content line, and a blank separator), which costs tokens the model doesn't need — the newline, the 2-space indent, and the blank line between matches are pure formatting overhead.

Switched to standard grep format: `file:line:content`, one line per match.

Per the issue's note, `contextLines` produces multi-line `content` (already prefixed per-line with its own line number by the search itself), so a single `file:line:content` line doesn't work for that case. For those matches, the header uses `-` instead of `:` as the separator (standard grep convention for marking a context block vs. an exact match line) and the content block keeps its own lines rather than being squeezed onto one:

```
# no context
src/foo.ts:42:const x = 1

# contextLines: 1
src/foo.ts:42-
41: const before = 1
42: const x = 1
43: const after = 1
```

## Test plan

- `pnpm run test:ava source/tools/search-file-contents.spec.tsx` — 56 tests pass, including the `contextLines` cases
- One existing test (`search_file_contents truncates very long matching lines to 300 chars`) located "the content line" and checked its raw length, assuming the old two-line-per-match layout where content had its own line. Updated it to strip the `file:line:` prefix before measuring, since that prefix now shares the line with content.

### DIFF
--- a/.changeset/search-results-grep-format.md
+++ b/.changeset/search-results-grep-format.md
@@ -1,0 +1,5 @@
+---
+"@nanocollective/nanocoder": patch
+---
+
+`search_file_contents` now formats results grep-style (`file:line:content`, one line per match) instead of spreading each match across three lines with a blank separator. Matches with `contextLines` still show the full multi-line context block, now with a `-` header separator matching grep's convention.
--- a/source/tools/search-file-contents.spec.tsx
+++ b/source/tools/search-file-contents.spec.tsx
@@ -763,10 +763,13 @@ test.serial(
 				// Count x's in result - should be 300 - 12 (searchTarget) = 288
 				const contentLine = result.split('\n').find(l => l.includes('searchTarget'));
 				t.truthy(contentLine, 'Should have content line');
-				// The trimmed content should be max 300 chars + ellipsis
-				const trimmedContent = contentLine!.trim();
+				// Content follows the `file:line:` prefix in grep-style output;
+				// strip it before checking the truncated length.
+				const contentMatch = contentLine!.match(/^.+?:\d+:(.*)$/);
+				t.truthy(contentMatch, 'Content line should be in file:line:content format');
+				const trimmedContent = contentMatch![1];
 				t.true(
-					trimmedContent.length <= 302, // 300 + ellipsis (1 char) + possible whitespace
+					trimmedContent.length <= 301, // 300 + ellipsis (1 char)
 					`Content should be truncated, got ${trimmedContent.length} chars`,
 				);
 			} finally {
--- a/source/tools/search-file-contents.tsx
+++ b/source/tools/search-file-contents.tsx
@@ -79,13 +79,20 @@ const executeSearchFileContents = async (
 			return `No matches found for "${args.query}"`;
 		}
 
-		// Format results with clear file:line format
+		// Format results grep-style: one line per match (`file:line:content`).
+		// Matches with context lines have multi-line content already prefixed
+		// per-line by the search itself, so those use `-` as the header
+		// separator (standard grep convention for context) and keep the block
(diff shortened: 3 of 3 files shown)


akramcodez:
@MsfPablo Thanks for the PR! The grep-style formatting looks good, and I like the - vs : distinction for contextLines. I noticed one small thing though: the results are still joined with \n\n, which leaves a blank line between every match. Since the goal of this issue is to reduce formatting overhead, could you change that to \n (or only add separators where they're actually needed for multi-line context)? Once that's updated and the tests still pass, this looks good to me.

will-lamerton (approved):
Verified locally: checked out the branch, 56 tests pass.

The `contextLines` handling is the part that needed care and you got it right. I confirmed against \`source/utils/file-search.ts:305-308\` that context content really is built as \`${lineNum}: ${line}\` per line and joined with newlines, and that non-context content is \`currentLine.trim()\` at \`:297\` so it can never contain one. That makes \`match.content.includes("\n")\` a sound discriminator, and \`-\` as the context header separator is the correct grep convention. Nice to see the code read rather than the issue text pattern-matched.

The spec change is a legitimate adaptation rather than a weakened assertion: the old test measured the whole line because content used to have a line to itself, so stripping the \`file:line:\` prefix before measuring is right. Tightening 302 to 301 while you were in there is a good catch.

Two follow-ups, neither blocking, and I will pick them up myself:

1. `.join("\n\n")` keeps a blank line between every match, which #766 also flagged. Measured on a real 30-match search: main 3045 chars, this PR 2985, single-newline 2956. So this lands about two thirds of the available saving. Single newline between plain matches with a blank line only around multi-line context blocks is probably the end state.

2. Minor edge case: when a context block collapses to a single line it falls into the plain branch and renders with the wrong separator and a doubled line number. Reproduced on a single-line file with `contextLines: 3`, giving `single.ts:1:1: const onlyLine = 1;`. Also reachable if a first context line exceeds `MAX_CONTEXT_CONTENT_LENGTH`. Cosmetic and rare.

Also worth correcting my own issue text: I estimated 3-4 tokens per match there, and the real figure is closer to 2 characters per match. The win is real and free but smaller than I wrote. That is on me, not on this PR.

Thanks for picking this one up.

PR #776, opened 2026-08-03
fix: stop echoing full file contents back after write_file

### LINKED ISSUE #762: write_file echoes back the entire file the model just wrote
## Problem

`source/tools/file-ops/write-file.tsx:48-64` reads the file back after writing and returns **every line of it**, line-number prefixed, as the tool result:

```ts
// Generate full file contents to show the model the current file state
let fileContext = '\n\nFile contents after write:\n';
for (let i = 0; i < lines.length; i++) {
    const lineNumStr = String(i + 1).padStart(4, ' ');
    const line = lines[i] || '';
    fileContext += `${lineNumStr}: ${line}\n`;
}
```

This is pure duplication. The model just sent that exact content as the tool call arguments. We pay for the same bytes twice in the same conversation, and the second copy is about 6 characters per line more expensive than the first because of the line-number prefix.

There is no cap, so the cost scales linearly with file size.

## Impact

Measured on files in this repo, the echo alone costs roughly:

- `source/app/App.tsx` (791 lines): ~8,500 tokens
- `source/hooks/useAppState.tsx` (447 lines): ~4,500 tokens
- `source/tools/tool-manager.ts` (414 lines): ~3,650 tokens

And because tool results stay in the conversation, that cost is re-sent on every subsequent step of the agent loop.

## Fix

Replace the echo with a confirmation line:

```ts
return `File ${action} successfully (${lineCount} lines).`;
```

This is losslessly removing 100% of the echo. The model cannot learn anything from the echo that it did not already know, since it authored the content.

Worth keeping the read-back itself (it verifies the write succeeded) and just not returning the content.

Small, isolated change in one function. Existing tests are in `source/tools/file-ops/write-file.spec.tsx` and will need their expected-result assertions updated. Run with `pnpm run test:ava source/tools/file-ops/write-file.spec.tsx`.

MsfPablo:
## Summary

Fixes #762. `executeWriteFile` read the file back after writing and appended every line, number-prefixed, to the tool result. That's pure duplication: the model just sent that exact content as the tool call arguments, so returning it again costs tokens (with an extra ~6 chars/line for the prefix) without adding information the model didn't already have. There's no cap, so it scales linearly with file size — the issue measured ~8,500 tokens on `source/app/App.tsx` (791 lines) in this repo — and since tool results stay in the conversation, that cost is re-sent on every subsequent agent step.

Kept the read-back itself (per the issue: it verifies the write succeeded, and its line/char/token counts already feed the existing confirmation message), just stopped appending the line-numbered content dump.

## Test plan

- `pnpm run test:ava source/tools/file-ops/write-file.spec.tsx` — 43 tests pass
- Updated the three existing tests that asserted the echoed content:
  - `write_file: create new file` / `write_file: overwrite existing file` — dropped the `'File contents after write:'` assertion
  - `write_file: returns file contents after write` → renamed to `write_file: reports accurate stats without echoing content back`, now asserts the line count is still accurate while the raw content and the `'File contents after write:'` marker are absent

### DIFF
--- a/.changeset/write-file-echo.md
+++ b/.changeset/write-file-echo.md
@@ -1,0 +1,5 @@
+---
+"@nanocollective/nanocoder": patch
+---
+
+`write_file` no longer echoes the full file contents back after writing. The model already authored that content as the tool call arguments, so returning it again was pure duplication that scaled with file size and got re-sent on every later step of the agent loop. The confirmation message (line/char/token counts) is unchanged.
--- a/source/tools/file-ops/write-file.spec.tsx
+++ b/source/tools/file-ops/write-file.spec.tsx
@@ -115,7 +115,6 @@ test('write_file: create new file', async t => {
 	const content = await readFile(filePath, 'utf-8');
 	t.is(content, 'Hello World\n');
 	t.true(result.includes('File written successfully'));
-	t.true(result.includes('File contents after write:'));
 });
 
 test('write_file: overwrite existing file', async t => {
@@ -129,7 +128,6 @@ test('write_file: overwrite existing file', async t => {
 	const content = await readFile(filePath, 'utf-8');
 	t.is(content, 'New content\n');
 	t.true(result.includes('File overwritten successfully'));
-	t.true(result.includes('File contents after write:'));
 });
 
 test('write_file: write empty file', async t => {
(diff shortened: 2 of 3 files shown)


akramcodez (approved):
LGTM great work @MsfPablo

will-lamerton:
Nice work @MsfPablo :)

Issue #769, opened 2026-08-03
No global cap on tool result size


will-lamerton:
## Problem

There is no global ceiling on how large a tool result can be. Only two tools truncate at all:

- `source/tools/execute-bash.tsx:50-54` (head-only, 2000 chars)
- `source/custom-tools/handler.ts:100-101` (head-only, same constant)

Everything else is unbounded, including:

- `read_file` with an explicit range covering a huge file
- `string_replace` and `write_file`, which return the whole post-edit file
- `git_diff` (`source/tools/git/git-diff.tsx:71`), which returns full diff content plus a `[Total: N lines]` footer
- `list_directory` recursive listings
- MCP tool results, which we do not control at all

A single large diff or a recursive listing of a big tree can dominate the context window, and because tool results persist in the conversation, the cost is re-sent on every subsequent step.

## Proposal

Add one choke point that caps any tool result before it enters the conversation, with head-plus-tail elision and a clear marker telling the model what was dropped and how to get it.

Making it structural rather than per-tool means unbounded output becomes impossible by construction, and it covers MCP and custom tools that we do not own.

Design questions worth settling on the issue:

- Where the choke point lives. `ToolManager` is the natural place since it already wraps every tool including MCP and custom ones.
- Whether the cap should be a fixed character count or a share of the model's context window. The latter is better but needs the context size plumbed through.
- Whether some tools should be exempt or get a higher ceiling.
- Whether the elided content should be recoverable, for example written to a temp file the model can read a range from.

## Related

Several individual uncapped tools have their own issues. This one is the structural fix that would make those individually less critical, but it is more invasive, so the per-tool fixes are probably worth landing first.

will-lamerton:
Part of #772.

RealBhupesh:
I'd like to take this issue.

Plan:
- add a shared head-and-tail truncation helper with a bounded character budget and an actionable elision marker;
- apply it at the central tool execution boundary so built-in, MCP, and file-based custom tools are covered;
- cover the interactive message-builder path as well, since streaming Bash and subagent results can bypass the normal handler;
- add focused regression tests and a patch changeset;
- run the focused tests plus the repository validation checks before opening a ready-for-review PR.

I'll keep structured tool payloads intact and cap only the LLM-facing text content.

RealBhupesh:
Implemented and submitted as ready-for-review PR #786: https://github.com/Nano-Collective/nanocoder/pull/786

What changed:
[...]

PR #777, opened 2026-08-03
fix: truncate LLM-facing command output head+tail instead of head-only

### LINKED ISSUE #764: Bash output truncation is head-only, discarding the tail where build and test errors live
## Problem

`source/tools/execute-bash.tsx:50-54` truncates bash output for the LLM by keeping the **first** 2000 characters and discarding everything after:

```ts
const llmContext =
    fullOutput.length > TRUNCATION_OUTPUT_LIMIT
        ? fullOutput.substring(0, TRUNCATION_OUTPUT_LIMIT) +
            '\n... [Output truncated. Use more specific commands to see full output]'
        : fullOutput;
```

Compiler and test-runner output puts the important part at the **end**:

- `tsc` prints the error list, then the count
- AVA prints passing output, then the failure summary
- most build tools print the exit status and error digest last

So head-truncating a `pnpm run test:types` or `pnpm run test:ava` run throws away exactly the part the model needs. It then has to re-run with a narrower command to recover information we already had in hand, which costs an extra step and an extra full round-trip.

## Fix

Head plus tail with an elision marker. Same byte budget, much higher information density:

```
<first ~800 chars>
... [N characters elided] ...
<last ~1200 chars>
```

Tail should get the larger share, since that is where the actionable content usually is.

Also worth revisiting whether `TRUNCATION_OUTPUT_LIMIT` (2000 chars, `source/constants.ts:80`) is the right budget for build output. That is roughly 500 tokens, which is tight for a failing typecheck.

Note the same constant is used by `source/custom-tools/handler.ts:100-101` with the same head-only behaviour, and by `source/components/bash-progress.tsx:88` for display (where head-only is arguably fine). A shared head-plus-tail helper would cover the first two.

Self-contained change to one function plus a small shared helper. Tests in `source/tools/execute-bash.spec.tsx`.

MsfPablo:
## Summary

Fixes #764. `execute-bash.tsx` and `custom-tools/handler.ts` both truncated long output for the model by keeping only the first `TRUNCATION_OUTPUT_LIMIT` (2000) characters. Compiler and test-runner output puts the actionable part at the end — `tsc` prints its error list then the count, AVA prints passing output then the failure summary, most build tools print the exit digest last — so head-only truncation throws away exactly what the model needs from a `pnpm run test:types` or `pnpm run test:ava` run. It then has to re-run with a narrower command to recover information already in hand.

Added a shared helper, `source/utils/truncate-output.ts`, that keeps both the head and the tail within the same byte budget (40/60 split, tail-weighted since that's where the actionable content usually is), with a marker showing how many characters were elided in between. Used it at both LLM-facing call sites:

- `formatBashResultForLLM` in `execute-bash.tsx`
- `truncate()` in `custom-tools/handler.ts`

Left `bash-progress.tsx`'s head-only truncation alone — per the issue, that one only feeds a display-only token-count estimate for the live progress UI, not something the model sees.

## Test plan

- New `source/utils/truncate-output.spec.ts` — 5 tests covering the under/at/over-limit cases, that both head and tail are kept, and the 40/60 split
- `pnpm run test:ava source/tools/execute-bash.spec.tsx source/custom-tools/handler.spec.ts` — 67 tests pass unmodified, including the existing truncation-length and truncation-marker assertions

### DIFF
--- a/.changeset/bash-output-truncation.md
+++ b/.changeset/bash-output-truncation.md
@@ -1,0 +1,5 @@
+---
+"@nanocollective/nanocoder": patch
+---
+
+`execute_bash` and custom tools now truncate long output by keeping both the head and the tail (tail-weighted) instead of only the head, so the actionable part of compiler/test-runner output (error list, failure summary, exit status) — which usually lands at the end — isn't discarded.
--- a/source/custom-tools/handler.ts
+++ b/source/custom-tools/handler.ts
@@ -5,6 +5,7 @@ import {TRUNCATION_OUTPUT_LIMIT} from '@/constants';
 import {renderBody} from '@/custom-tools/template';
 import type {CustomToolMetadata} from '@/types/custom-tools';
 import type {ToolHandler} from '@/types/index';
+import {truncateHeadAndTail} from '@/utils/truncate-output';
 
 /**
  * Build a `ToolHandler` that renders the script body and runs it under the
@@ -97,8 +98,7 @@ export function runScript(
 }
 
 function truncate(text: string): string {
-	if (text.length <= TRUNCATION_OUTPUT_LIMIT) return text;
-	return text.slice(0, TRUNCATION_OUTPUT_LIMIT) + '\n... [Output truncated]';
+	return truncateHeadAndTail(text, TRUNCATION_OUTPUT_LIMIT);
 }
 
 /**
--- a/source/tools/execute-bash.tsx
+++ b/source/tools/execute-bash.tsx
@@ -9,6 +9,7 @@ import {useTheme} from '@/hooks/useTheme';
 import {type BashExecutionState, bashExecutor} from '@/services/bash-executor';
 import type {NanocoderToolExport} from '@/types/core';
 import {jsonSchema, tool} from '@/types/core';
+import {truncateHeadAndTail} from '@/utils/truncate-output';
 
 /**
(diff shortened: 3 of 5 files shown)


akramcodez (approved):
LGTM great work @MsfPablo

luisangelrod (commented):
Nice extraction and the tail-weighted split matches the failure-output use case well. One contract detail is worth clarifying before this helper is reused by the global-cap work in #769: `limit` currently limits only the retained source characters, not the returned string. For a 10,000-character input with `limit = 2_000`, the result is 2,052 characters after adding the marker.

That is consistent with the two old call sites (they also appended a marker beyond `TRUNCATION_OUTPUT_LIMIT`), so I do not see it as a blocker for this PR. But the generic helper name/signature makes a hard output ceiling easy to assume. I would either document that `limit` is the retained-content budget, or make the marker count toward the limit and add a `result.length <= limit` test. The latter would make this helper safe to reuse for #769 without quietly exceeding the configured cap.

I also traced both call sites and verified the 40/60 calculation directly. I could not complete the repository test command locally because the current checkout's frozen install reports a pre-existing `patchedDependencies`/lockfile configuration mismatch, so this feedback is based on the diff plus a direct execution of the helper logic rather than claiming a full suite run.

will-lamerton:
Nice work @MsfPablo :)

PR #778, opened 2026-08-03
feat: make list_directory file sizes opt-in via showSizes

### LINKED ISSUE #767: list_directory emits byte sizes on every entry by default
## Problem

`source/tools/list-directory.tsx:161-164` appends a byte size to every file entry:

```ts
const sizeStr = entry.size
    ? ` (${entry.size.toLocaleString()} bytes)`
    : '';
```

So a listing looks like:

```
useAppState.tsx (15,504 bytes)
tool-manager.ts (12,134 bytes)
App.tsx (29,252 bytes)
```

That is roughly 8-10 tokens per entry for information that is rarely load-bearing when the model is just orienting itself in a directory. On a 50-entry listing it is 400-500 tokens, and on a recursive listing it multiplies.

The information is already available on demand through `read_file` with `metadata_only: true`.

There is also an `lstat` call per file (`source/tools/list-directory.tsx:96-102`) purely to get the size, so dropping it by default is a small wall-clock win on large directories too.

## Fix

Make sizes opt-in behind a `showSizes` boolean on the tool schema, defaulting to false. Skip the `lstat` when it is false.

The tool description at `source/tools/list-directory.tsx:188` says "List directory contents with file sizes", so that needs updating to match.

Contained to one file. Add the flag to the `ListDirectoryArgs` interface and the `inputSchema`, gate the `lstat` and the size string on it, update the description. Tests in `source/tools/list-directory.spec.tsx`.

MsfPablo:
## Summary

Fixes #767. `list_directory` appended a byte size to every file entry unconditionally — roughly 8-10 tokens per entry for information that's rarely load-bearing when the model is just orienting itself in a directory (400-500 tokens on a 50-entry listing, more on recursive listings). It also cost an `lstat` syscall per file purely to get that size.

Added a `showSizes` boolean to `ListDirectoryArgs` and the tool schema, defaulting to `false`:
- Gated the `lstat` call on `showSizes` (skipped entirely when not requested)
- The existing `entry.size ? ... : ''` size-string logic already handles `size` being `undefined` correctly, so no change needed there
- Updated the tool description, which claimed "List directory contents with file sizes" — now mentions `showSizes=true` as opt-in, and points to `read_file`'s `metadata_only` for a single file's size

## Test plan

- `pnpm run test:ava source/tools/list-directory.spec.tsx` — 30 tests pass
- Updated `list_directory shows file sizes` → `list_directory hides file sizes by default` (asserts no `bytes` in output) and added `list_directory shows file sizes when showSizes=true` to cover the opt-in path explicitly

### DIFF
--- a/.changeset/list-directory-opt-in-sizes.md
+++ b/.changeset/list-directory-opt-in-sizes.md
@@ -1,0 +1,5 @@
+---
+"@nanocollective/nanocoder": minor
+---
+
+`list_directory` no longer includes file sizes by default — they cost an `lstat` syscall per file plus output tokens for information that's rarely needed just to orient in a directory. Pass `showSizes=true` to opt back in, or use `read_file` with `metadata_only=true` for a single file's size.
--- a/source/tools/list-directory.spec.tsx
+++ b/source/tools/list-directory.spec.tsx
@@ -209,7 +209,7 @@ test.serial('list_directory shows files and directories', async t => {
 	}
 });
 
-test.serial('list_directory shows file sizes', async t => {
+test.serial('list_directory hides file sizes by default', async t => {
 	t.timeout(10000);
 	const originalCwd = process.cwd();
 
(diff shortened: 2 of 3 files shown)


akramcodez (commented):
LGTM great work @MsfPablo :)

will-lamerton:
Nice work @MsfPablo  :)

PR #782, opened 2026-08-04
fix: preview very large files before ranged reads

### LINKED ISSUE #760: read_file forces a metadata round-trip on every file over 300 lines
## Context

A third-party benchmark ([harness efficiency, not quality](https://nqawhc.github.io/articles/harness-efficiency-not-quality/)) tested nanocoder 1.29.0 against Pi, OpenCode and Claude Code with the model held constant. Quality was statistically identical across all four harnesses, but nanocoder averaged 37.4 tool calls per task against OpenCode's 16.6, at the same grade. The author's specific criticism was that nanocoder was "re-reading the same files more than any other harness".

That observation is not model behaviour. It is a forced round-trip in `read_file`.

## Problem

`source/tools/read-file.tsx:122-160`: when `read_file` is called with no line range on a file longer than `FILE_READ_METADATA_THRESHOLD_LINES` (300, `source/constants.ts:54`), it returns **zero content**. It returns metadata plus an instruction blob telling the model to call again with `start_line`/`end_line`.

`FILE_READ_CHUNK_SIZE_LINES` is 250 (`source/constants.ts:56`), so reading one 800-line file costs 1 metadata call plus 4 chunk calls, all against the same path.

Compounding this, `source/tools/read-file.tsx:137-157` emits copy-paste-ready chunk calls in the response:

```
[Large file - Choose one approach:]
[2. Progressive read: Read file in chunks (recommended chunk size: 200-300 lines)]
   - read_file({path: "...", start_line: 1, end_line: 250})
   - read_file({path: "...", start_line: 251, end_line: 500})
   ... and N more chunks to complete the file
```

We are not merely permitting the expensive path, we are prompting it with ready-made calls. `source/app/prompts/sections/file-editing.md:5` reinforces the same instruction.

## Scale

Measured against this repo's own tree: 90 of 439 non-spec source files (20%) exceed 300 lines, and 31 exceed 500. In a larger codebase the hit rate is plausibly higher.

## Proposed fix

A read should always return usable content. Truncate with a footer instead of returning nothing:

- Raise `FILE_READ_METADATA_THRESHOLD_LINES` substantially (1500-2000 lines feels right, open to discussion).
- On overflow, return the first N lines of actual content plus `[truncated at line N of M, call read_file with start_line to continue]`.
- Drop the generated chunk-call examples.

This is strictly better than today. The common case collapses to one call, and the rare genuinely-huge-file case degrades to today's two calls rather than five.

## Related

Blocked-adjacent to the `markFileSeen` bug on the same code path, which stacks a third round-trip on top of this one.

RealBhupesh:
## Summary

Fixes #760.

- return files up to 1,500 lines in full
- preview the first 250 lines of larger files with a precise continuation hint
- remove generated multi-chunk boilerplate
- retain `metadata_only=true` for explicit metadata requests
- update prompt guidance, formatter behavior, and focused tests

The branch is rebased onto current `main`, merge conflicts are resolved, and an unrelated settings-tabs test edit has been removed.

## Validation

- feature-specific large-file preview cases pass
- formatter assertions are ANSI-safe
- `pnpm run test:format`
- `pnpm run test:lint`
- `git diff --check`

### DIFF
--- a/.changeset/tough-poems-read.md
+++ b/.changeset/tough-poems-read.md
@@ -1,0 +1,5 @@
+---
+"@nanocollective/nanocoder": patch
+---
+
+Return a useful preview before requiring ranged reads for very large files
--- a/source/app/prompts/sections/file-editing.md
+++ b/source/app/prompts/sections/file-editing.md
@@ -2,7 +2,7 @@
 
 **Reading files**:
 - Always read a file before editing it. Never blindly suggest edits.
-- Large files (>300 lines) return metadata first — use `start_line`/`end_line` for specific sections.
+- Very large files (>1500 lines) return a 250-line preview with a continuation hint — use `start_line`/`end_line` for specific sections or to continue the preview.
 - Use `metadata_only=true` to check file size without reading content.
 
 **Editing tools** (always read first):
--- a/source/constants.ts
+++ b/source/constants.ts
@@ -51,9 +51,10 @@ export const DEFAULT_TERMINAL_WIDTH = 200;
 export const DEFAULT_TERMINAL_COLUMNS = 80;
 
 // === FILE READING ===
-export const FILE_READ_METADATA_THRESHOLD_LINES = 300;
-export const FILE_READ_CHUNKING_HINT_THRESHOLD_LINES = 500;
-export const FILE_READ_CHUNK_SIZE_LINES = 250;
+// Files at or below this size are returned in full. Larger files get a bounded
+// preview so the first read still gives the model useful context.
+export const FILE_READ_PREVIEW_THRESHOLD_LINES = 1500;
+export const FILE_READ_PREVIEW_LINES = 250;
 
 // === @-FILE MENTION INLINING (prompt-processor) ===
 // Files at or under this many lines are inlined in full when @-mentioned.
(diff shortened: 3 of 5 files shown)


RealBhupesh:
Follow-up: the first CI run exposed two test-only issues. The read_file formatter assertion assumed unstyled terminal output, so it now strips ANSI styling before checking the rendered line range. The settings-tabs spec was still asserting the former four-tab contract even though the current component exposes five categories; that expectation now matches Appearance, Input, Behavior, Providers, and Advanced. Both affected suites pass locally, and the full GitHub check suite is rerunning on this commit.

will-lamerton:
Hey @RealBhupesh! Can you resolve the merge conflicts? Thanks :)

RealBhupesh:
Merge conflicts are resolved and the branch is rebased onto current `main`. I also removed the unrelated stale settings-tabs edit from this PR, while retaining the ANSI-safe assertion that belongs to the read-file formatter coverage.

The feature-specific read-file cases pass locally. The current red type/build/test jobs are from the ACP SDK regression already present on `main` after #822; #830 is the focused prerequisite repair and passes the full local gate.

PR #783, opened 2026-08-04
feat: add copy-to-clipboard toolbar for message bubbles

### LINKED ISSUE #746: [Feature] Universal Copy Buttons on Chat Bubbles in VS Code Extension
## Description
Add a quick "Copy" button to the end of both user prompt bubbles and AI response bubbles in the VS Code extension's chat panel.

## Use Case
Developers frequently need to extract code snippets, terminal commands, or previous prompts from the chat. Manually highlighting text in a narrow VS Code sidebar webview is tedious and error-prone.

## Proposed Solution
Inject a subtle clipboard icon at the bottom right of every message block in the `messages-container`. When clicked, it should copy the raw text/markdown to the clipboard and display a brief "Copied!" checkmark animation.

## Alternatives Considered
- Relying on native VS Code selection and keyboard shortcuts (Ctrl+C), but this provides poor UX in narrow sidebars.

## Additional Context

- **UI:** Add an absolutely positioned SVG icon to the message div during the DOM injection phase (likely within `plugins/vscode/media/chat-panel.js`).
- **Logic:** Use the standard browser `navigator.clipboard.writeText()` API.

## Contribution Guidelines

arpan7sarkar:
## Description

Adds a "Copy" button to both user prompt bubbles and AI response bubbles in the VS Code extension's chat panel (closes #746). Hovering over a message bubble reveals a subtle clipboard icon (bottom-right for user messages, bottom-left for agent messages); clicking it copies the raw markdown text via navigator.clipboard.writeText() and briefly swaps the icon for a checkmark to confirm the copy.

[Screencast_20260803_234750.webm](https://github.com/user-attachments/assets/b7070cdd-45e8-41e1-b8ae-f202ef37761a)

For streamed agent responses, the copy button always reads the latest in-progress text (via a dataset.rawText kept in sync on each chunk) rather than a stale snapshot, so copying mid-stream works correctly.

Type of Change

- [x] New feature
- [ ] Bug fix
- [ ] Breaking change
- [ ] Documentation update

Changeset

- [x] Added a changeset (pnpm changeset) describing this change for the changelog

nanocoder-vscode is listed in .changeset/config.json's ignore array, so no changeset is needed for this package.

Testing

Automated Tests

- [ ] New features include passing
- [x] All existing tests pass (pnpm test:format, pnpm test:lint)

chat-panel.js is a plain DOM script with no existing test harness/spec coverage in this extension, so no automated test was added for the UI behavior. Verified pnpm run test:format and pnpm run
test:lint both pass.

Manual Testing

- [x ] Tested with Ollama
- [ x] Tested with OpenRouter
- [ x] Tested with OpenAI-compatible API
- [ ] Tested MCP integration 

Built and installed the extension locally (pnpm run build:vscode) and confirmed it activates
correctly and the copy button rendechat panel. Full end-to-endmessage-send testing against a live provider is still pending on my end.

Checklist

- [x] Code follows project style gu
- [x] Self-review completed
- [x] No breaking changes (or clear
- [x] Appropriate logging added using structured logging (see CONTRIBUTING.md#logging) — n/a, no new logging paths introduced


arpan7sarkar:
@akramcodez  Check out this pr and tell me if any changes are needed , thanks

akramcodez:
@arpan7sarkar Thanks for the contribution! I really like the overall implementation, especially the use of `dataset.rawText` to keep the copy button in sync with streamed responses. That's a clean solution for the mid-stream copy problem and makes sure users always copy the latest response rather than a stale snapshot.

I went through the issue, the implementation, and the user flow, and I have a few comments before I'd be comfortable approving this.

### 1. Missing Clipboard API Guard

**What is wrong:**

The copy button click handler assumes `navigator.clipboard` is always available.

**Why is it wrong:**

In some VS Code webview environments (or older runtimes), `navigator.clipboard` can be unavailable. Calling `navigator.clipboard.writeText(...)` in that case throws synchronously before the Promise is created, meaning the current `.catch()` block never runs and the UI won't enter the error state.

**Where is it wrong:**

`plugins/vscode/media/chat-panel.js` inside `createCopyToolbar()`.

**What should be done:**

Guard against `navigator.clipboard` being unavailable and wrap the whole operation in a `try/catch` (preferably using `async/await`) so the button always shows the correct success or failure state.

### 2. Accessibility & Button Semantics

**What is wrong:**

The icon-only button doesn't specify `type="button"` and doesn't have an accessible name.

**Why is it wrong:**

Without an explicit button type, it could behave as a submit button if this UI is ever placed inside a form. Since the button only contains an SVG icon, screen readers also have no way of announcing its purpose.

**Where is it wrong:**

`plugins/vscode/media/chat-panel.js` inside `createCopyToolbar()`.

**What should be done:**

Set:

* `btn.type = 'button'`
* `btn.setAttribute('aria-label', 'Copy message')`

to improve accessibility and future-proof the component.


### 3. Missing Test Coverage

**What is wrong:**

The new functionality currently relies entirely on manual verification.

**Why is it wrong:**

This PR introduces two important behaviors:

* Copying static user/assistant messages.
* Copying in-progress streamed responses using `dataset.rawText`.

Without automated coverage, future refactors could accidentally break the streamed-copy logic without anyone noticing.

**What should be done:**

I'd recommend adding tests covering:

* Copying a normal message.
* Copying an in-progress streamed response (ensuring the latest `dataset.rawText` is copied).
* Clipboard failure path (button enters the error state).

### 4. UI Suggestion (Non-blocking)

This isn't a blocker, but I think the current placement of the copy button looks a little disconnected from the message bubble. It feels more like a floating action than metadata associated with that specific message.

Since Nanocoder already persists conversations, I think we could make better use of the bottom area of each message by displaying both the copy action and the local timestamp.

For example:

**User message**

```text
┌──────────────┐
│ Hello        │
└──────────────┘
    2:34 PM  📋 
```

**Assistant message**

```text
┌─────────────────────────────┐
│ Hello! I am Nanocoder...    │
└─────────────────────────────┘
📋  2:34 PM 
```

This keeps the layout balanced:

* **User messages:** Copy button → Timestamp
* **Assistant messages:** Timestamp → Copy button

I think this has a few advantages:

* The copy action feels naturally attached to the message instead of floating beside it.
* Users can immediately see when each prompt and response was sent.
* Since conversations are already saved in history, users would still see the local timestamp when reopening previous chats, providing useful context about when a conversation happened.
* It also leaves room for future message actions without changing the overall layout.

Overall, I think this is a really nice feature and the streaming solution is well thought out. Once the clipboard guard, accessibility improvements, and some automated coverage are added, I'd be happy to approve it. The timestamp suggestion is non-blocking, but I think it would make the overall chat experience feel much more polished.

arpan7sarkar:
Hey @akramcodez  I looked into the best way to add these tests and wanted to flag something before just doing it.

chat-panel.js is a single top-level IIFE with no exports and no module system, and there's currently no test setup for this file at all (no DOM environment, nothing wired up). To actually test the copy-to-clipboard behavior (normal copy, streamed copy reading the live dataset.rawText, and the clipboard-failure state) against the real code, I need some way to run it in a fake DOM from Node.

I looked into jsdom for this — it's the standard tool for testing browser DOM code from Node, dev-only (never ships in the extension bundle or CLI), and it's what most projects reach for instead of hand-rolling a fake DOM. The alternative is writing our own minimal DOM stub with Node's built-in vm module, which avoids a new dependency but means we maintain that fake DOM ourselves and it'll quietly drift out of sync as chat-panel.js grows.

Do I have the go-ahead to add jsdom as a devDependency for this, or is there a different testing approach you'd prefer for this package?

arpan7sarkar:
I have added the timestamp chat messages.

image

akramcodez:
@arpan7sarkar Great work! Just a couple of small suggestions:

- Could you remove the unnecessary code comments?
- The timestamp and copy button look slightly misaligned. Could you align them so the bottom metadata row looks cleaner and more polished?

arpan7sarkar:
Hey @akramcodez  , 
I have fixed the allignment issue 
image

and removed unnecessary commits now you can merge it thanks.

will-lamerton:
Thanks for this PR - feel free to add yourself as a contributor to our website via a PR which I will approve :)

arpan7sarkar:
Hi , @will-lamerton  I have added the pr to add me as contributor here https://github.com/Nano-Collective/organisation/pull/89#issue-5071081345 , 
I also have added another pr do check it out  , https://github.com/Nano-Collective/nanocoder/pull/789

PR #784, opened 2026-08-04
fix: bound string_replace result context

### LINKED ISSUE #763: string_replace returns the entire file after every edit, uncapped
## Problem

`source/tools/file-ops/string-replace.tsx:83-89` appends the **entire post-edit file**, line-number prefixed, to the tool result:

```ts
const newLines = newContent.split('\n');
let fileContext = '\n\nUpdated file contents:\n';
for (let i = 0; i < newLines.length; i++) {
    const lineNumStr = String(i + 1).padStart(4, ' ');
    const line = newLines[i] || '';
    fileContext += `${lineNumStr}: ${line}\n`;
}
```

There is no cap. A three-line edit to a large file returns the whole file.

## Impact

Measured on files in this repo, per edit:

- `source/app/App.tsx` (791 lines): ~8,500 tokens
- `source/hooks/useAppState.tsx` (447 lines): ~4,500 tokens
- `source/tools/tool-manager.ts` (414 lines): ~3,650 tokens

Every edit, every time. Tool results persist in the conversation, so each edit's copy is re-sent on all subsequent steps of the agent loop. A task with several edits to a large file compounds badly.

## Fix

Return a window around the edit rather than the whole file. The tool already computes `startLine` and `newEndLine` (`source/tools/file-ops/string-replace.tsx:68-81`), so the range is available.

Suggested: 20 lines either side of the changed range, with a marker showing it is a window. That preserves the genuinely useful part (letting the model confirm surrounding context survived the edit) while dropping the rest.

Open question worth discussing on the issue: whether even the window is needed, or whether the confirmation line alone is enough. `string_replace` already fails loudly on a non-unique or absent match, so the model has strong feedback without the echo.

## Related

Same pattern as the `write_file` echo, but this one is less trivially removable, since unlike `write_file` the model did not author the full resulting file.

RealBhupesh:
## Summary

- Bound successful `string_replace` results to a 20-line context window on each side of the edited range.
- Added explicit line ranges and omitted-range markers so the model can still orient itself without receiving the whole file.
- Corrected line-number detection for inline matches and added large-file and boundary regression tests.
- Added a patch changeset.

## Validation

- `pnpm run test:ava source/tools/file-ops/string-replace.spec.tsx` (42 passed)
- `pnpm run test:format`
- `pnpm run test:lint`
- `pnpm run test:types`
- `pnpm run test:knip`
- `pnpm run test:audit`

The full AVA suite has one unrelated existing failure in `source/app/components/settings-tabs.spec.tsx:109`, where the test expects `Display` but the rendered output does not include it.

Fixes #763

### DIFF
--- a/.changeset/bounded-string-replace-results.md
+++ b/.changeset/bounded-string-replace-results.md
@@ -1,0 +1,5 @@
+---
+"@nanocollective/nanocoder": patch
+---
+
+Bound `string_replace` results to a context window around the edited range.
--- a/source/tools/file-ops/string-replace.spec.tsx
+++ b/source/tools/file-ops/string-replace.spec.tsx
@@ -338,6 +338,71 @@ test('string_replace: works with large replacements', async t => {
 	t.true(result.includes('Successfully replaced'));
 });
 
+test('string_replace: returns only a bounded context window around the edit', async t => {
+	const filePath = await createTestFile(
+		'large-context.txt',
+		Array.from({length: 100}, (_, i) => `line ${i + 1}`).join('\n'),
+	);
+
+	const result = await executeStringReplace({
+		path: filePath,
+		old_str: 'line 50',
+		new_str: 'changed line 50',
+	});
+
+	t.regex(result, /Updated file context \(lines 30-70 of 100\)/);
+	t.true(result.includes('[... lines 1-29 omitted ...]'));
+	t.true(result.includes('  30: line 30'));
+	t.true(result.includes('  50: changed line 50'));
+	t.true(result.includes('  70: line 70'));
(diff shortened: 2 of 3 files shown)


RealBhupesh:
CI validation update: all checks pass except the repository-wide Unit Tests & Coverage Analysis job. It reports the existing `source/app/components/settings-tabs.spec.tsx:109` assertion because `Display` is absent from the rendered output. The failure reproduces locally on this branch and is outside the three changed files; focused `string-replace` coverage remains green with 42 passing tests.

will-lamerton:
Nice work @RealBhupesh :)

Issue #772, opened 2026-08-03
Tracking: harness efficiency (step count and tool result tokens)


will-lamerton:
Tracking issue for the efficiency work prompted by a third-party benchmark.

## Background

[Harness efficiency, not quality](https://nqawhc.github.io/articles/harness-efficiency-not-quality/) tested nanocoder 1.29.0 against Pi, OpenCode and Claude Code with the model held constant (DeepSeek V4 Flash) across eight bug-fix tasks in a large codebase. Nanocoder got 64 runs, 8 per task, so it is the best-sampled harness in the study.

Results:

- Pi: 4 tools, 1,340 fixed overhead, 28.0 avg steps, 14,775 output tokens, 2.1m, grade 2.34
- OpenCode: 10 tools, 7,197 fixed overhead, 16.6 avg steps, 17,463 output tokens, 3.1m, grade 2.07
- Claude Code: 27 tools, 23,132 fixed overhead, 37.9 avg steps, 58,370 output tokens, 8.0m, grade 2.42
- Nanocoder: 15 tools, 6,121 fixed overhead, 37.4 avg steps, 55,844 output tokens, 5.2m, grade 2.25

Quality was statistically flat: every confidence interval overlaps every other. The author's conclusion is that harness choice is an efficiency decision, not a quality one.

The good news is our system prompt is the second leanest of the four, about a quarter of Claude Code's. That is the part we control most directly and we have done it well.

The bad news is all our cost sits in the agent loop. 37.4 steps and 55.8K output tokens to reach the same grade OpenCode reaches in 16.6 steps and 17.5K. Since the system prompt is re-sent every step, a lean prompt multiplied by 37 steps loses to a fat prompt multiplied by 17.

The author's specific criticism was that nanocoder was "re-reading the same files more than any other harness". An audit found that is not model behaviour, it is a designed-in forced round-trip.

## Steps and latency

These are the ones that move step count, which is what the benchmark actually measured.

- [x] #760 read_file forces a metadata round-trip on every file over 300 lines
- [x] #761 read_file metadata path skips markFileSeen, forcing an extra read before edits

## Tokens

These compound, because tool results persist in the conversation and are re-sent on every subsequent step.

- [x] #762 write_file echoes back the entire file the model just wrote  — landed via #776
- [x] #763 string_replace returns the entire file after every edit, uncapped  — landed via #784
- [x] #764 Bash output truncation is head-only, discarding the tail where build and test errors live  — landed via #777
[...]

PR #785, opened 2026-08-04
fix: summarize oversized git diffs

### LINKED ISSUE #770: git_diff truncation is head-only, dropping the end of large diffs
> **Rewritten 2026-08-09.** The original framing of this issue was wrong. It claimed `git_diff` returned the full diff with no ceiling. There has been a 500-line cap since before the #772 work started. The real defect is the shape of that truncation, described below.

## Problem

`source/tools/git/git-diff.tsx:69` caps the diff, and `source/tools/git/utils.ts:654` implements the cap:

```ts
const truncatedContent = lines.slice(0, maxLines).join('\n');
```

Truncation is head-only, at 500 lines. That is the same flaw #764 fixed for `execute_bash`, and it bites harder here.

A diff is ordered by path, so the first 500 lines are the first file or two alphabetically, not the important ones. Cutting a large diff at the head gives the model `source/ai-sdk-client/...` in full and silently drops everything after it. For the common case of orienting in a branch's changes, that is close to the least useful 500 lines available.

Because tool results persist in the conversation, those 500 lines are also re-sent on every subsequent step of the agent loop.

Secondary, cosmetic: when truncation fires, the result carries two footers. `utils.ts:673` appends `... [Diff truncated: showing 500 of N lines]`, then `git-diff.tsx:71` appends `[Total: N lines]` on top of it. One of them should go.

## Fix

Preferred, and specific to this tool: when the diff exceeds the cap, return the `--stat` summary instead of a head slice, plus a note that the model should request specific paths with `git_diff({paths: [...]})`. A file-level summary is the right granularity for orienting in a large diff, and it is dramatically cheaper than 500 lines of one file. The model can then pull the hunks it actually needs.

Fallback if that is too large a change: head-plus-tail elision, matching what #764 did for bash, so at least both ends of the diff survive.

Either way, drop the duplicate footer, and reconsider whether 500 lines is the right ceiling for something re-sent on every step.

`args.stat` mode correctly skips truncation already, since stat output is compact. Leave that path alone.

## Related

Partly covered by a global tool-result cap (#769) if that lands, but the `--stat` fallback is smarter than anything a generic cap can do, so this is worth handling specifically either way.

RealBhupesh:
Closes #770
Part of #772

## Summary

- use a bounded `--stat-count=20 --stat` summary for oversized multi-file diffs
- return bounded head-and-tail patch content when an oversized diff is already scoped to one file
- keep recovery guidance actionable and preserve useful context from both ends
- add utility and integration regression tests plus a patch changeset

## Validation

- focused git utility and git-diff suites: 37 passed
- `pnpm run test:format`
- `pnpm run test:lint`
- `git diff --check`

### DIFF
--- a/.changeset/quiet-git-diffs.md
+++ b/.changeset/quiet-git-diffs.md
@@ -2,4 +2,4 @@
 "@nanocollective/nanocoder": patch
 ---
 
-Summarized oversized `git_diff` results with a diffstat and guidance to request a specific file for details.
+Bound oversized multi-file `git_diff` results to a 20-entry diffstat while preserving the total file count, and kept file-scoped results as bounded head-and-tail patches.
--- a/.changeset/quiet-git-diffs.md
+++ b/.changeset/quiet-git-diffs.md
@@ -1,0 +1,5 @@
+---
+"@nanocollective/nanocoder": patch
+---
+
+Summarized oversized `git_diff` results with a diffstat and guidance to request a specific file for details.
--- a/source/tools/git/git-diff.spec.tsx
+++ b/source/tools/git/git-diff.spec.tsx
@@ -2,6 +2,10 @@
  * Git Diff Tool Tests
  */
 
+import {execSync} from 'node:child_process';
+import {mkdtempSync, rmSync, writeFileSync} from 'node:fs';
+import {tmpdir} from 'node:os';
+import {join} from 'node:path';
 import React from 'react';
 import test from 'ava';
 import {render} from 'ink-testing-library';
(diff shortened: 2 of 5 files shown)


RealBhupesh:
CI update: all required checks pass except Unit Tests & Coverage Analysis. The full `npx c8 ava` job reproduces the existing failure at `source/app/components/settings-tabs.spec.tsx:109` (`Display` is not found in the rendered category tabs). This is outside the five files changed by the PR and matches the local `pnpm run test:all` result recorded in the PR description. CodeQL, build, type, format, lint, audit, Semgrep, unused-dependency, and changeset checks are green.

will-lamerton:
Hey @RealBhupesh - thanks for this one, and for the `truncateDiff` rewrite in particular. Moving it from head-only to head-and-tail is the right call and matches what we settled on in #764. I checked and it's only consumed by `git-diff.tsx`, so nothing else shifts underneath it. The `buildGitDiffArgs` extraction is a genuine tidy-up too.

There's a couple blockers to resolve first before merge. The summary tells the model to use the `file` parameter to get detail, but doing exactly that returns the same message:

```
STEP 1: git_diff({})
 a.ts | 601 +++++...
 1 file changed, 600 insertions(+), 1 deletion(-)

Diff is too large to return in full (606 lines). Use the file parameter to request a specific path for detailed output.

STEP 2: git_diff({file: "a.ts"})
 a.ts | 601 +++++...
 1 file changed, 600 insertions(+), 1 deletion(-)

Diff is too large to return in full (606 lines). Use the file parameter to request a specific path for detailed output.
```

The file-scoped diff is still over `MAX_DIFF_LINES`, so it gets summarised again. The net effect is that there's now no way to see the patch for any single file whose diff exceeds 500 lines, and the recovery path we recommend is a no-op.

That makes it a regression rather than a missed optimisation. Before this change a 606-line diff returned its first 500 lines: head-only and imperfect, but real content. Now it returns none, and costs a wasted round-trip to discover that. Step count is the metric the efficiency work is chasing, so a loop that burns a step and yields nothing is the thing we can least afford.

The fix should be small. Only substitute the stat when there is more than one file to summarise. Once the diff is already scoped to a single path, fall through to `truncateDiff` and return the bounded head-and-tail content, which is exactly what your change to that function now makes good.

One other thing while you're in here: the stat is itself uncapped, so at the moment we're swapping one unbounded output for a smaller unbounded one. A 400-file change produces 401 lines and around 9,200 characters, all returned. `git diff --stat-count=20 --stat` caps the file list, elides the remainder with `...` and keeps the `400 files changed` summary line, which gets you the bound for one flag.

Last thing, unrelated to the code: this branch is 53 commits behind `main`, which is where that `settings-tabs.spec.tsx` failure is coming from. It passes on current `main`, so a rebase should clear it and let the diff be validated against the tree it'll actually land in. Same applies to what you had on #784.

Happy to take another look once that's sorted :)

RealBhupesh:
Thanks for the review. The rebased update now does both requested things:

- a file-scoped oversized diff returns bounded head-and-tail patch content instead of another stat summary, so the suggested recovery step is real
- multi-file summaries use `--stat-count=20 --stat`, bounding the file list while retaining the total-files summary

The focused git-diff suites pass (37 tests), and format/lint checks pass locally.

The current Type Checks, Verify Build, and Unit Tests failures are inherited from `main` after #822. I opened #830 to add the ACP SDK's required `providerId`; that prerequisite passes the full local project gate.

will-lamerton:
Nice work @RealBhupesh :)

PR #788, opened 2026-08-04
fix: clarify refusal message when only metadata was read (#761)

### LINKED ISSUE #761: read_file metadata path skips markFileSeen, forcing an extra read before edits
> **Resolution note, added 2026-08-09.** Resolved via the first of the two options in the Fix section below, not the one the title implies. Commit `638a0ea9` clarified the read-before-edit refusal messages in `diff_edit`, `string_replace` and `write_file` so they state that files over 300 lines need a ranged `read_file` call. The metadata path at `source/tools/read-file.tsx:160` still returns early **without** calling `markFileSeen`, and that is deliberate: marking a file seen when the model has only received metadata would defeat the guard and permit a blind edit. Do not "fix" that early return. The remaining round-trip cost is tracked in #760, which is the right place to remove it, by having reads return content in the first place.

## Problem

`source/tools/read-file.tsx:159` returns the metadata response for large files and exits early, without calling `markFileSeen`. That call only happens on the content path at `source/tools/read-file.tsx:173`.

The read-before-edit guard in `source/utils/read-tracker.ts` is therefore never satisfied by a metadata read. A model that calls `read_file` on a file over 300 lines, gets metadata, and then tries to edit will be refused and forced to read again.

This stacks an extra round-trip on top of the forced metadata round-trip that `read_file` already imposes on files over 300 lines.

## Reproduction

1. Call `read_file` on any file longer than 300 lines with no `start_line`/`end_line`.
2. Attempt a `string_replace` on that file.
3. The edit is refused for not having read the file, despite `read_file` having been called on it.

## Fix

Judgement call on which way to resolve it:

- If the metadata response is kept, it should **not** mark the file seen, because the model genuinely has not seen the contents and a blind edit is exactly what the tracker exists to prevent. In that case the fix is to make the refusal message explicit that a ranged read is required, so the model recovers in one step rather than guessing.
- If the linked `read_file` change lands and reads always return content, this resolves itself, and the early return should mark seen for whatever content was returned.

Worth deciding alongside the parent issue rather than in isolation.

The relevant files are small and self-contained: `source/utils/read-tracker.ts` is 39 lines and well commented on the intent. Tests live in `source/tools/read-file.spec.tsx`.

DeepamJha:
Closes #761

## Description
`read_file`'s metadata-only response (files over 300 lines with no
line range) correctly does not call `markFileSeen`, since the model
hasn't actually seen the file's contents. However, the refusal
messages in `diff_edit`, `string_replace`, and `write_file` (which
check `hasSeenFile`) didn't explain *why* the edit was refused after
a metadata read, or what to do about it.

Per the maintainer's noted judgment call in the issue: kept the
metadata-response behavior as-is (not marking seen), and instead
made all three refusal messages explicit that files over 300 lines
require specifying `start_line`/`end_line` on `read_file` to read
actual content — so the model (or contributor) can recover in one
step instead of guessing why the edit was blocked.

Updated all three call sites for consistency, not just the one
implied by the issue's reproduction steps, since all three tools
share the same `hasSeenFile` guard and refusal pattern.

## Type of Change
- [x] Bug fix

## Changeset
- [x] Added a changeset (`pnpm changeset`)

### Automated Tests
- [x] All existing tests pass (`bash scripts/test.sh`)

Verified against `main` directly: both show the same 63 pre-existing
failures unrelated to this change (coverage threshold issues in
`fetch-models.ts` and others); this change introduces none.

### Manual Testing
Reviewed the change against the reported repro steps in #761 —
message wording verified for accuracy in all three files.

## Checklist
- [x] Self-review completed
- [x] No breaking changes

### DIFF
--- a/.changeset/wacky-tables-shout.md
+++ b/.changeset/wacky-tables-shout.md
@@ -1,0 +1,5 @@
+---
+"@nanocollective/nanocoder": patch
+---
+
+Clarify read-before-edit refusal messages to specify that files over 300 lines need a ranged read
--- a/source/tools/file-ops/diff-edit.tsx
+++ b/source/tools/file-ops/diff-edit.tsx
@@ -339,7 +339,7 @@ const diffEditValidator = async (
 	if (!hasSeenFile(absPath)) {
 		return {
 			valid: false,
-			error: `You must read "${args.path}" before editing it. Call read_file on it first, then retry diff_edit with SEARCH blocks copied from the file.`,
+			error: `You must read "${args.path}" before editing it. Call read_file on it first — if the file is over 300 lines, specify start_line and end_line to read its actual content, not just metadata — then retry diff_edit with SEARCH blocks copied from the file.`,
 		};
 	}
 
--- a/source/tools/file-ops/string-replace.tsx
+++ b/source/tools/file-ops/string-replace.tsx
@@ -214,7 +214,7 @@ const stringReplaceValidator = async (
 	if (!hasSeenFile(absPath)) {
 		return {
 			valid: false,
-			error: `You must read "${path}" before editing it. Call read_file on it first, then retry string_replace with old_str copied exactly from the file.`,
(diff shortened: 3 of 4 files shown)


akramcodez (approved):
LGTM such a great work @DeepamJha just maintain the pr content :)

Issue #758, opened 2026-08-03
[Feature] Instant Stream Cancellation (`Esc` / `Ctrl+C`)


akramcodez:
## Description
Ensure that pressing `Escape` in the VS Code extension (or `Ctrl+C` in the CLI) instantly aborts the current LLM network request.

## Use Case
If the AI starts generating an answer and the user immediately realizes they made a typo in their prompt (or the AI is going down the wrong path), they want to stop it instantly to save tokens/time.

## Proposed Solution
Bind the `Escape` key in the webview to trigger the `cancel` message. Ensure the backend `NanocoderAcpClient` uses an `AbortController` to physically terminate the HTTP request to the LLM provider, rather than just hiding the output.

## Alternatives Considered
- Only providing a mouse-clickable "Stop" button, but power users heavily rely on keyboard shortcuts for cancellation.

## Additional Context
- [x] I have searched existing issues to ensure this is not a duplicate
- [x] This feature aligns with the project's goals (local-first AI assistance)

## Implementation Notes (optional)
- **VS Code Extension:** Add a keydown listener on `document` or the `chat-input` textarea to catch `Escape`.
- **Backend:** Ensure the Vercel AI SDK `streamText` call is passed an `AbortSignal`.

## Contribution Guidelines
- If you would like to work on this issue, please **comment below to ask for assignment** first. 
- Before opening a Pull Request, please briefly **share your implementation plan** in this thread so we can align on the approach!

arpan7sarkar:
Hey @akramcodez  I am working on this issue please assign me it .

akramcodez:
Hey @arpan7sarkar, thanks for your interest! Feel free to share a short implementation plan first, and if everything looks good, I'll assign it to you.

arpan7sarkar:
Hey @akramcodez , here's my plan for this one:

Frontend (chat-panel.js in the VS Code webview)
- Move the Escape listener from the textarea to document, so it works even if focus is somewhere else (like a tool card or the response area).
- Keep it guarded by the existing isProcessing flag so it does nothing when idle.
- On Escape, just call the same cancel path the Stop button already uses (postMessage({type: 'cancel'}) + setProcessing(false)).
- Also added a nanocoder.cancel command so it shows up in the Command Palette too.

Backend
- Checked the ACP session + chat handler code turns out the AbortController is already wired into the streamText call, and cancel already aborts the real request, not just the UI. So nothing to change here, just confirmed it actually works.

[...]

PR #789, opened 2026-08-04
feat(vscode): cancel in-flight requests with Escape.

arpan7sarkar:
## Description

Implements #758 

Pressing Escape in the VS Code extension's chat panel now cancels an in-flight LLM request, mirroring the existing Stop button. The listener is bound on the webview's document (not just the chat input) so it fires even when focus has moved to a tool card, dropdown, or the streaming response area, and it's guarded by isProcessing so it's a no-op when idle. Also added a nanocoder.cancel command for the Command Palette.

While wiring this up, found that cancelling mid-stream surfaced as a Nanocoder prompt failed: RequestError: Internal error toast — AcpAgent.prompt was throwing a generic error on abort, and the ACP JSON-RPC layer flattened it into an opaque failure. Fixed that too: cancellation is now recognized explicitly, reported inline in the chat as "Cancelled by user", and resolved with stopReason: 'cancelled' instead of rejecting. NanocoderAcpClient also tracks an in-flight cancel client-side and suppresses the toast for it, so this holds even against a CLI build that hasn't picked up the agent-side change. Genuine errors still show the toast as before.

## Type of Change

- [x] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

## Changeset

### Automated Tests

- [x] All existing tests pass (pnpm test:all completes successfully)

### Manual Testing

- [ ] Tested with Ollama
- [ ] Tested with OpenRouter

## Checklist

- [x] Self-review completed


arpan7sarkar:
Hey @akramcodez  do check it out and let me know if any changes are needed .

arpan7sarkar:
Hey @will-lamerton  , @Avtrkrb  , @akramcodez , Do check this functionlity out , its perfectly working let me now if any change is needed , else merge it 
Thanks

akramcodez (dismissed):
LGTM @arpan7sarkar, amazing work on this!

arpan7sarkar:
Thanks @akramcodez , now waiting for another review :)

will-lamerton:
Hey @arpan7sarkar - nice PR. A few small things before merge:

**Blocking:** `cancel()` doesn't flush `pendingPermissions`. Repro: send a message, let a tool hit a permission prompt, press Escape. The agent resolves via its abort race, but the client's resolver stays in the map, so `hasPendingPermissions()` stays true and `_handlePrompt` rejects every later message with "Please approve or deny the pending tool" until reload. Resolve each pending entry with `{outcome: {outcome: 'cancelled'}}` (the agent already handles that shape) and clear the map, plus dismiss the stale permission card. Pre-existing on the Stop button path, but binding Escape globally turns it from rare into routine.

**Please also:**
- Add an `acp-client.spec.ts` case for the new `cancelRequested` backstop. That's the half of the fix that has to survive a version-skewed CLI, and the spec file already exists.
- Add `t.false(session.turnActive)` to the new agent test, matching the sibling test above it. The `finally` still running on the new `return` path is the non-obvious bit worth pinning.
- Extract a `requestCancel()` and call it from both the Escape handler and the Stop button click, so the two copies can't drift.

**Non-blocking:** `'Operation was cancelled'` is now the fourth site coupled to that string (thrown in three places in `chat-handler.ts`, compared in `message-helpers.tsx:18`). Worth an `isCancellationError()` helper, but happy to take that as a follow-up rather than hold this PR.

arpan7sarkar:
Hey @will-lamerton  , Thanks for giving me the broad picture and telling me the things that I am missing , I will add all this then raised a PR thanks

akramcodez:
Hey @arpan7sarkar, could you please resolve the merge conflicts when you get a chance? Thanks!

arpan7sarkar:
Hello @akramcodez  today i will update the fixex that @will-lamerton  told me then after that i will commmit that in the mean time i was working on #814  this pr here i actully find a big ci issue that i have fixed and mentioned there you can have a look into that

arpan7sarkar:
Hello @akramcodez  @will-lamerton  , Have implemnted all the requirments and also fixed merge conflicts have a look here :-)

PR #790, opened 2026-08-04
fix(tools): stop prefixing edit tool responses with line numbers

### LINKED ISSUE #765: Edit tools return line-number prefixes while read_file strips them
## Problem

The same file is returned to the model in two different representations depending on which tool produced it.

`read_file` deliberately strips line numbers (`source/tools/read-file.tsx:175-176`):

```ts
// Return content without line numbers for clean content-based editing
return linesToReturn.join('\n');
```

But `string_replace` (`source/tools/file-ops/string-replace.tsx:86-88`) and `write_file` (`source/tools/file-ops/write-file.tsx:58-60`) both return it with a `%4d: ` prefix on every line.

Two costs:

1. The prefix is about 6 characters per line for information the model rarely needs, given both our editors are content-based rather than line-based. On a 500-line file that is roughly 750 tokens of pure formatting.
2. Seeing the same file in two representations within one session is a plausible independent trigger for re-reading to disambiguate, which is the exact behaviour flagged in the benchmark writeup.

## Fix

Drop the prefixes wherever the content is not being cited by line number. Since `read_file` already establishes the no-line-numbers convention and that is the tool the model sees most, the edit tools should match it.

## Related

Largely moot for `write_file` and `string_replace` if the echo-removal issues land, since there will be no echoed content left to prefix. Worth keeping open as a convention decision, and it still applies to any future tool that returns file content.

Pixie-19:
## Description

Fixes #765.

This PR removes line-number prefixes from the responses returned by the `string_replace` and `write_file` tools.

Previously these tools returned file contents like:

    1: import ...
    2: const x = ...

while `read_file` returned raw file contents without numbered prefixes. This inconsistency could cause follow-up edits to fail because `string_replace` expects the exact file contents.

The tools now return the raw file contents, making their output consistent with `read_file`.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Changeset

### Automated Tests

- [x] Existing TypeScript compilation passes (`npm run test:types`)
- [x] Biome lint and format checks run successfully (schema version warning only)

### Manual Testing

- [ ] Tested with Ollama
- [ ] Tested with OpenRouter

## Checklist

- [x] Self-review completed
- [x] No breaking changes

### DIFF
--- a/.changeset/ten-clouds-show.md
+++ b/.changeset/ten-clouds-show.md
@@ -1,0 +1,5 @@
+---
+"@nanocollective/nanocoder": patch
+---
+
+Fixed string_replace and write_file tool responses to return raw file contents instead of line-number-prefixed output, making them consistent with read_file and preventing follow-up edit mismatches. Thanks to @Pixie-19. Closes #765.
--- a/source/tools/file-ops/string-replace.tsx
+++ b/source/tools/file-ops/string-replace.tsx
@@ -80,13 +80,7 @@ const executeStringReplace = async (
 	const endLine = startLine + oldStrLines.length - 1;
 	const newEndLine = startLine + newStrLines.length - 1;
 
-	const newLines = newContent.split('\n');
-	let fileContext = '\n\nUpdated file contents:\n';
-	for (let i = 0; i < newLines.length; i++) {
-		const lineNumStr = String(i + 1).padStart(4, ' ');
-		const line = newLines[i] || '';
-		fileContext += `${lineNumStr}: ${line}\n`;
-	}
+	const fileContext = `\n\nUpdated file contents:\n${newContent}`;
 
 	const rangeDesc =
 		startLine === endLine
--- a/source/tools/file-ops/write-file.tsx
+++ b/source/tools/file-ops/write-file.tsx
@@ -47,18 +47,10 @@ const executeWriteFile = async (args: {
 
 	// Read back to verify and show actual content
 	const actualContent = await readFile(absPath, 'utf-8');
-	const lines = actualContent.split('\n');
-	const lineCount = lines.length;
+	const fileContext = `\n\nFile contents after write:\n${actualContent}`;
+	const lineCount = actualContent.split('\n').length;
 	const charCount = actualContent.length;
 	const estimatedTokens = calculateTokens(actualContent);
-
-	// Generate full file contents to show the model the current file state
-	let fileContext = '\n\nFile contents after write:\n';
-	for (let i = 0; i < lines.length; i++) {
-		const lineNumStr = String(i + 1).padStart(4, ' ');
-		const line = lines[i] || '';
-		fileContext += `${lineNumStr}: ${line}\n`;
-	}
 
 	const action = fileExists ? 'overwritten' : 'written';
 	return `File ${action} successfully (${lineCount} lines, ${charCount} characters, ~${estimatedTokens} tokens).${fileContext}`;


will-lamerton:
Thanks for this @Pixie-19, and sorry it's taken until now to get back to you.

Your read of the problem was correct, and the patch is exactly the minimal change #765 asked for. Unfortunately it's raced a few other PRs against the same tracking issue (#772) and main has moved underneath it, so I'm going to close this one rather than ask you to rebase into something that no longer applies.

Two things changed:

**`write_file`** now returns no file contents at all. #776 landed in 2a72da53 and removed the echo entirely, on the grounds that the model just sent us that exact content as the tool call arguments, so returning it is pure duplication. Rebasing this PR would reintroduce the echo, and it would fail the regression test added alongside it (`source/tools/file-ops/write-file.spec.tsx:193` asserts the result does not contain `File contents after write:`). Worth noting for future PRs here: `pnpm run test:ava source/tools/file-ops/write-file.spec.tsx` would have surfaced that, the checklist in the description only covered types and lint.

**`string_replace`** no longer echoes the whole file either. #784 landed in d8ca638e and bounds the result to a 20-line window either side of the edit, with a `lines X-Y of N` header and `[... lines A-B omitted ...]` markers. That deliberately kept the line numbers, because in a partial view they orient the model against the header and the elision markers in a way they didn't when the whole file was returned. #798 has since applied the same treatment to `diff_edit`, so all three edit tools now share one bounded format, and I've closed #765 recording that as the convention.

If you'd like another go at something in this area, #769 (no global cap on tool result size) and #770 (git_diff returns the full diff with no ceiling) are both unclaimed and structurally similar to what you did here. #769 in particular is the one that would make this class of bug impossible by construction rather than tool by tool.

Genuinely appreciate the contribution, and apologies for the collision. That's on our side for not splitting #772 into claimable sub-issues before people started work.

Pixie-19:
> Thanks for this @Pixie-19, and sorry it's taken until now to get back to you.
> 
> Your read of the problem was correct, and the patch is exactly the minimal change #765 asked for. Unfortunately it's raced a few other PRs against the same tracking issue (#772) and main has moved underneath it, so I'm going to close this one rather than ask you to rebase into something that no longer applies.
> 
> Two things changed:
> 
> **`write_file`** now returns no file contents at all. #776 landed in [2a72da5](https://github.com/Nano-Collective/nanocoder/commit/2a72da53d86654957062dd1d8c33467f869ed472) and removed the echo entirely, on the grounds that the model just sent us that exact content as the tool call arguments, so returning it is pure duplication. Rebasing this PR would reintroduce the echo, and it would fail the regression test added alongside it (`source/tools/file-ops/write-file.spec.tsx:193` asserts the result does not contain `File contents after write:`). Worth noting for future PRs here: `pnpm run test:ava source/tools/file-ops/write-file.spec.tsx` would have surfaced that, the checklist in the description only covered types and lint.
> 
> **`string_replace`** no longer echoes the whole file either. #784 landed in [d8ca638](https://github.com/Nano-Collective/nanocoder/commit/d8ca638ec5cd1673fa84546d7847a8d67a20f3d5) and bounds the result to a 20-line window either side of the edit, with a `lines X-Y of N` header and `[... lines A-B omitted ...]` markers. That deliberately kept the line numbers, because in a partial view they orient the model against the header and the elision markers in a way they didn't when the whole file was returned. #798 has since applied the same treatment to `diff_edit`, so all three edit tools now share one bounded format, and I've closed #765 recording that as the convention.
> 
> If you'd like another go at something in this area, #769 (no global cap on tool result size) and #770 (git_diff returns the full diff with no ceiling) are both unclaimed and structurally similar to what you did here. #769 in particular is the one that would make this class of bug impossible by construction rather than tool by tool.
> 
> Genuinely appreciate the contribution, and apologies for the collision. That's on our side for not splitting #772 into claimable sub-issues before people started work.

Thanks for the detailed explanation! I understand closing this PR, especially with how #776, #784, and #798 established a unified format.
I appreciate the tip on running pnpm run test:ava for targeted spec tests I'll incorporate that in my future work.
Since #769 is already closed, I'd love to take on any other available issues or upcoming tasks you think would be a good fit
Thanks again for the review and guidance.

PR #796, opened 2026-08-05
feat: show per-response token usage and estimated cost indicator

### LINKED ISSUE #756: [Feature] Token Usage & Cost Indicator per Response
## Description
Display a small, subtle token count (and estimated cost, if applicable) at the bottom of every AI response.

## Use Case
When using paid APIs (like OpenRouter or OpenAI), developers are highly sensitive to token usage. A small indicator helps them build an intuition for how much context they are sending and when they need to clear their session to save money.

## Proposed Solution
Append a tiny, grayed-out string like `(Tokens: 4.2k | ~$0.01)` at the bottom of the AI's chat bubble (in the extension) or terminal output (in the CLI).

## Alternatives Considered
- Only showing it in a separate "Usage" tab, but placing it inline with the response is much more helpful for immediate feedback.

## Additional Context

- Most LLM API responses (via Vercel AI SDK) include a `usage` object in the response payload.
- In the VS Code extension, this can be injected directly into the DOM below the `messages-container`.

## Contribution Guidelines

lakshay-2411:
## Description
     
  This PR adds a per-response token usage and estimated cost indicator to both the CLI and the VS Code extension. Closes #756

  When using paid APIs (OpenRouter, OpenAI, etc.), developers want immediate feedback on how many tokens each response consumed and roughly what it cost, so they can build an intuition for context size and know when to clear their session.

  **CLI:**
  - The gray footer under each assistant message (previously a client-side estimate like `~120 tokens`) now shows the provider-reported usage plus estimated cost,
  e.g. `Tokens: 4.2k | ~$0.01`.
  - Cost is computed from models.dev pricing via the existing `getModelPricing()` helper (cache-backed, best-effort). The cost segment is omitted for local/free models (e.g. Ollama), and sub-cent costs render as `<$0.01`.
  - When a provider reports no usage at all, the footer falls back to the previous client-side estimate — behavior never regresses.
  - New `source/usage/format.ts` (compact token/cost formatters) and `source/usage/response-usage.ts` (usage + cost payload builder, with injectable pricing lookup for tests).

  **VS Code extension:**
  - The ACP conversation loop now accumulates provider-reported usage across all model calls in a turn and returns it on the `PromptResponse` via the ACP SDK's experimental `usage` field, with the estimated cost in `_meta['nanocoder/usage']`.
  - The extension forwards both through the existing `prompt_response` webview update, and the chat panel renders the same subtle gray indicator under each finished response.
  - Rebuilt `assets/nanocoder-vscode.vsix` accordingly.

  **Design decisions** (from the approach discussed in #756 — happy to adjust):
  - CLI shows the indicator per API call (each rendered assistant message); the extension shows it per turn (ACP's usage field is turn-scoped).
  - Always-on, no config toggle — the footer already existed; this upgrades its accuracy.

  ## Type of Change

  - [ ] Bug fix
  - [x] New feature
  - [ ] Breaking change
  - [ ] Documentation update

  ## Changeset

  ## Testing

  ### Automated Tests

  New tests: `source/usage/format.spec.ts` (compact token/cost formatting, invalid inputs), `source/usage/response-usage.spec.ts` (cost computation, lump-sum totals, unavailable/failing pricing), component tests for the footer (real usage, no pricing, fallback to estimate), and two ACP tests verifying usage accumulation onto the `PromptResponse` (and its absence when the provider reports none).

  ### Manual Testing

  - [ ] Tested with Ollama
  - [ ] Tested with OpenRouter

  ## Checklist

  - [x] Self-review completed


will-lamerton:
Hey @lakshay-2411 - thanks for this, the feature is well-scoped and for the most part looking great! Requesting changes on two things, plus the merge conflicts.

**Merge conflicts.** The branch is currently conflicting with `main` in `plugins/vscode/src/acp-client.ts` and `plugins/vscode/src/chat-webview-provider.ts`. Could you rebase onto latest `main` and rebuild the vsix afterwards so the committed artifact matches the resolved source?

**1. ACP zero-fills input/output, silently zeroing cost for lump-sum providers.** `recordUsage` (`source/acp/acp-conversation.ts:63`) coerces missing `inputTokens`/`outputTokens` to `0` while still accumulating `totalTokens`. `buildResponseUsage` then sees both as finite and takes the input/output branch (`source/usage/response-usage.ts:43`) instead of the lump-sum averaging branch at line 50, so cost comes out `0`. I verified against the real code:

```
turnUsage {inputTokens: 0, outputTokens: 0, totalTokens: 1_000_000} -> cost = 0
direct     {totalTokens: 1_000_000}                                 -> cost = 9
```

Cost `0` then fails the `cost > 0` check in the webview, so any provider reporting only `total_tokens` gets no cost segment in the GUI with no signal that pricing was skipped. The same coercion undercounts when a provider reports input but not output. Suggest keeping the accumulators as `number | undefined` and only filling a field once a finite value arrives, so a total-only turn reaches `buildResponseUsage` as `{totalTokens: n}`. Worth a third ACP test with a total-only mock.

**2. The CLI now awaits a possibly-networked pricing lookup inside the streaming-to-static swap.** `conversation-loop.tsx:496` awaits `buildResponseUsage` after `setStreamingContent('')` at line 460 has already wiped the live message off screen, but before `addToChatQueue` appends the static one. `getModelPricing` has no in-process memo: `getModelsData` does a disk read plus `JSON.parse` every call, and `findModelById` falling through to `findModelByName` does it twice. The models.dev payload is 3.46MB, ~13ms per read+parse, so roughly 26ms of blocking parse per assistant message on the warm path.

The cold path is worse. With no cache or one older than 7 days, `fetchModelsData` runs with a 10s headers timeout and its failure path reads the cache again. Offline, that's a blank screen where the reply used to be, up to ten seconds, on every response, including for Ollama users whose models aren't on models.dev anyway. Suggest memoizing the parsed database (and the negative lookup) at module level in `models-dev-client.ts`, which also speeds up `/usage`, and not blocking the queue push on cost.

**Smaller things, not blocking:**

- Cost ignores cache-read/write tokens. `ModelInfo['cost']` only carries `input`/`output`, so with prompt caching the estimate is materially over-stated, and this footer is far more visible than `/usage`. A note in the changeset would help.
- Resumed sessions regress to the old estimate. `session-history-renderer.tsx:186` and `subagent-view.tsx:61,91` render `AssistantMessage` without `usage`, and usage isn't persisted, so `/resume` shows `~120 tokens` where the live session showed `Tokens: 4.2k | ~$0.01`. Fine as a follow-up.
- `acp-conversation.ts:150` returns a bare `{stopReason: 'end_turn'}` while the other eight returns go through `withTurnUsage`. Harmless today since it precedes any model call, but it reads as an oversight.
- Widening `prompt()` to `Promise<unknown>` forces the cast in `chat-webview-provider.ts:230`. Returning `PromptResponse | undefined` gets `usage` typed for free and drops the `any`.
- `source/usage/` isn't in the CLAUDE.md key-directories list.
- The manual testing boxes are all unchecked. Given this touches the render path on every response, at least one Ollama run and one paid-API run would be reassuring, especially for item 2.

lakshay-2411:
@will-lamerton please review

will-lamerton:
Looks good @lakshay-2411 :)

Thanks for this PR - feel free to add yourself as a contributor to our website via a PR which I will approve :)

PR #797, opened 2026-08-05
Hands-free voice mode (VAD) + install-on-demand for voice dependencies (PR4 of 6)

RONAK-AI647:
## What this does
Add fetaure for #622 

Two things in this PR:

1. **No more manual terminal installs.** The first time you try voice mode and 
   the required tools (sox, whisper, piper) aren't installed, Nanocoder now asks 
   you right there in the app - "Install now? Y/N" - and installs everything 
   automatically if you say yes. No separate terminal, no manual steps.

2. **Real hands-free voice.** You can now switch to `/voice hands-free` and just 
   talk - no need to hold a button. Nanocoder listens continuously and 
   automatically detects when you start and stop speaking.

## Known limitation

The hands-free listening uses a simple volume-based detection method rather than 
a full AI speech detector. This keeps it lightweight and working across every 
platform, but it means loud background noise could occasionally be picked up as 
speech. This is a reasonable first version - a smarter version can come later if 
needed.

## Test 1: Install prompt
1. Make sure sox/whisper/piper are NOT installed.
2. Run /voice or Ctrl+T.
3. Confirm you see: "Voice mode requires sox, whisper, piper (~150 MB). Install now?"
4. Press Esc/No - confirm it cancels cleanly and doesn't ask again this session.
5. Restart the app, trigger voice again - confirm it DOES ask again (session-only memory).

## Test 2: Real install
1. Trigger the prompt again, select Yes.
2. Confirm you see actual progress updates, not just a frozen "Installing..." screen.
3. Once done, run the same command-line check we used before 
   (e.g. `which sox`, `which whisper-cli`, `which piper`) and confirm all three 
   are genuinely present - not just sox.

## Test 3: Hands-free mode
1. Run /voice hands-free.
2. Speak without pressing anything - confirm the status bar shows Listening 
   automatically.
3. Stop speaking - confirm it transitions to Processing, then Speaking (reads 
   the reply), then back to Listening automatically for the next turn.

## Test 4: Push-to-talk still works
1. Run /voice push-to-talk.
2. Confirm Ctrl+T still works exactly as it did in PR3.

### DIFF
--- a/plugins/voice/src/dependencies.spec.ts
+++ b/plugins/voice/src/dependencies.spec.ts
@@ -1,0 +1,76 @@
+import test from 'ava';
+import { checkDependenciesInstalled, installDependencies } from './dependencies.js';
+
+test('checkDependenciesInstalled returns true when all commands succeed', async (t) => {
+	const mockChecker = async () => true;
+	const result = await checkDependenciesInstalled(mockChecker);
+	t.true(result.installed);
+	t.is(result.missing.length, 0);
+	t.true(result.details.sox);
+	t.true(result.details.whisper);
+	t.true(result.details.piper);
+});
+
+test('checkDependenciesInstalled identifies missing dependencies correctly', async (t) => {
+	const mockChecker = async (cmd: string) => {
+		if (cmd.includes('sox') || cmd.includes('rec')) return true;
+		return false;
+	};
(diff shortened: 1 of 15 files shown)


will-lamerton:
Nice work @RONAK-AI647 :)

PR #801, opened 2026-08-06
fix(mcp): don't leave a server connected when tool discovery fails

### LINKED ISSUE #787: MCP server remains connected when tool discovery fails
## Description

When an MCP transport connects successfully but the initial `client.listTools()` call fails, `MCPClientManager.connectToServer()` rejects as expected. However, the server has already been registered in the internal client, transport, and configuration maps.

This leaves the failed server visible as connected for the rest of the session.

## Reproduction

I reproduced this at commit `a4d054bb5a875df7bc54accc194d52f4acf29112` using a local stdio MCP server that:

1. completes MCP initialization successfully;
2. intentionally returns an error for `tools/list`.

After `connectToServer()` rejects:

- `getConnectedServers()` still includes the server;
- `isServerConnected()` returns `true`;
- `getServerInfo()` reports `connected: true` with zero tools;
- the stdio child process remains alive until an explicit disconnect.

## Suspected cause

In `source/mcp/mcp-client.ts`, the client, transport, and server configuration are added to their maps before the initial `listTools()` call. The failure path logs and rethrows the error, but does not close the client or roll back the partial state.

## Proposed implementation plan

I would like to submit a focused PR that:

1. registers the server only after connection and initial tool discovery both succeed;
2. performs best-effort client cleanup and removes any partial state if initialization fails;
3. adds a deterministic AVA regression test for the failed `listTools()` path;
4. includes the required patch changeset.

Expected regression assertions would verify that the connection rejects, no server remains registered, connected status is false, and the client/transport is closed.

Would this approach be acceptable, or is there an existing lifecycle design I should follow?

serhiizghama:
Closes #787.

`connectToServer()` registered the client, transport, and config in its internal maps right after the handshake but *before* calling `listTools()`. So when a server connects fine but its initial `tools/list` fails, the rejection propagates as expected — but the server is left registered: `isServerConnected()` returns true, `getServerInfo()` reports `connected: true` with zero tools, and the stdio child stays alive for the rest of the session.

Moved the map registration to after `listTools()` succeeds, so a server only becomes visible once both connection and tool discovery are done. The catch now also closes the partially-established client best-effort, so its transport / child process doesn't leak when discovery fails.

Added two AVA tests around the connect lifecycle — one for the failed `listTools()` path (rejects, nothing registered, client closed) and one confirming the success path still registers. They use a small `createClient()` seam so the test can inject a client without standing up a real transport.

### DIFF
--- a/.changeset/fix-mcp-failed-tool-discovery.md
+++ b/.changeset/fix-mcp-failed-tool-discovery.md
@@ -1,0 +1,5 @@
+---
+"@nanocollective/nanocoder": patch
+---
+
+Fixed an MCP server staying visible as connected when its initial `tools/list` call failed. `connectToServer()` now registers the client, transport, and config only after tool discovery succeeds, and closes the partially-established client on failure so its transport/child process doesn't leak.
--- a/source/mcp/mcp-client.spec.ts
+++ b/source/mcp/mcp-client.spec.ts
@@ -1065,3 +1065,79 @@ test('callTool sanitizes object arguments to strings when schema expects a strin
 		t.is(typeof capturedArgs.path, 'string', 'Path should remain a string');
 	}
 });
+
+// ============================================================================
+// Regression Tests for connectToServer lifecycle (failed tool discovery)
+// ============================================================================
+
+// Subclass exposing the createClient() seam so we can drive the connect path
+// with a fake client and no real transport/network.
+class SeamMCPClient extends MCPClient {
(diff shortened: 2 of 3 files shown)


akramcodez (approved):
LGTM thanks @serhiizghama

will-lamerton:
Thanks for this PR @serhiizghama - feel free to add yourself as a contributor to our website via a PR which I will approve :)

PR #814, opened 2026-08-07
[Feature] Add session renaming functionality and improve session metadata

### LINKED ISSUE #749: [Feature] Chat Session Management Polish in VS Code Extension
## Description
Improve the visibility and UX of the Chat History and Session management features within the VS Code extension.

## Use Case
Users often switch between different coding tasks (e.g., fixing a bug, then writing a new feature). Without robust session management visible in the VS Code extension, context pollution occurs, and the AI gets confused by previous instructions.

## Proposed Solution
Expose the existing `history-view` more prominently. Add a toggle button in the chat panel header to switch between the active chat and the history list. Allow users to rename or quickly delete sessions from this view.

## Alternatives Considered
- Opening history in a separate VS Code panel, which disjoints the user experience.

## Additional Context

- **UI:** Unhide the `history-view` div through a cleaner toggle mechanism in `plugins/vscode/media/chat-panel.html`. Add edit/delete icons to the generated list items.
- **Backend:** The `listSessions`, `resumeSession`, and `deleteSession` cases in `plugins/vscode/src/chat-webview-provider.ts` are already implemented; this is purely a frontend webview UX enhancement.

## Contribution Guidelines

arpan7sarkar:
## Description
Fixes #749 

https://github.com/user-attachments/assets/87fe50ab-20c7-406c-b372-2e408c13780c

Improve session management & history UX in the VS Code extension.
- Session renaming: Sessions can now be renamed directly from the History view via a renameSession ACP extension method (extMethod) on the CLI's ACP agent, backed by the session manager — no full resume required.
- History view navigation: Creating a new chat (+) or resuming a session from the History list now returns to the active chat view instead of leaving the panel stuck on the session list.

## Type of Change

- [x] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

## Changeset

### Automated Tests

### Manual Testing

- [ ] Tested with Ollama
- [ ] Tested with OpenRouter

## Checklist

- [x] Self-review completed


arpan7sarkar:
Hey @Avtrkrb  , @akramcodez  , @will-lamerton  do check this out its all done :-)

will-lamerton:
Hey @arpan7sarkar - thanks for this, the `extMethod` approach is the right call and the `titleManuallySet` threading is clean. A few things before merge:

**Blocking:** `case 'renameSession'` in `chat-webview-provider.ts` is missing its `break;`, so it falls through into `requestPathInfo`. That's not just cosmetic - `npx tsc --noEmit -p plugins/vscode/tsconfig.json` fails with three new `TS2339: Property 'path' does not exist` errors that aren't on `main`, and at runtime every rename hits `fs.statSync(undefined)` and gets swallowed by the bare `catch {}`. (Worth knowing: `test:all` doesn't cover the plugin and `build:ext` is esbuild with no type checking, so CI can't catch this today.)

**Please also fix:** `updatedAt` maps to `lastAccessedAt`, which is only ever written by `createSession` and `loadSession`. Autosave never bumps it, so the new "Today / Yesterday / Last 7 Days" buckets and the "5m ago" label show created-or-resumed time, not last activity. A session created three days ago and chatted in a minute ago renders as "3d ago". Either bump `lastAccessedAt` on autosave or add a real `updatedAt` field.

**Two smaller ones:**
- The checklist says tests were added but the diff has none. `session-manager.spec.ts` and `acp-agent.spec.ts` already exist, and `renameSession` (empty title, over-length, missing session, flag round-trip) plus `extMethod` (unknown method, non-string params) would be easy to cover there.
- The comment at `useSessionAutosave.ts:153` says `/rename` in the CLI sets this flag. It doesn't - `onRenameSession` is wired to `setSessionName`, which is display-only state for the status line and never reaches disk. Either wire it through `sessionManager.renameSession` or reword the comment.

Nice-to-haves, not blockers: `input.maxLength = 100` so an over-length title fails locally instead of round-tripping to an error toast, and sorting within date groups by `updatedAt` rather than reversed creation order.

Thanks :)

arpan7sarkar:
Hey @willlamerton, added all the changes you asked for.

While testing I also ran into two test bugs in this repo, unrelated to this PR, that were failing CI runs for others too. Fixed both:

1. styled-select-input.spec.tsx was failing in CI only. CI sets FORCE_COLOR=1, so Ink prints real color codes around the text, and the test's regex expects plain text starting with >. Fixed by stripping ANSI codes before checking, same as a few other test files already do.
2. user-input.spec.tsx had 18 tests that rendered a component and never unmounted it. Since tests in the same file run at the same time, all those leftover components kept running in the background for the rest of the file, which can slow things down enough to time out later tests on a busy CI runner. Added the missing cleanup to all 18.

Both are pre-existing on main, not caused by this PR, just happened to show up while I was testing this branch. Can you take a look?

will-lamerton:
Great work @arpan7sarkar :)

Issue #804, opened 2026-08-06
[Feature] Visual "Thinking" Loading Indicator for Chat UI


akramcodez:
## Description

Add a visual "thinking" or "processing" indicator to the chat interface (VS Code extension and Web UI) that appears immediately after the user submits a prompt, while waiting for the AI to return the first token of the response.

## Use Case

Currently, when a user sends a prompt, there is no immediate visual feedback in the chat history while the backend is processing. If the model takes several seconds to compute (which is common for complex queries or local models), the interface appears blank or unresponsive. A loading state provides immediate reassurance that the request was received and is being processed by the AI.

## Proposed Solution

Introduce a loading animation in the chat window where the AI's response will eventually appear. This could be:
- An animating "NC" (Nanocoder) logo with a gradient or pulsing effect.
- An ASCII-style animation (e.g., bouncing dots, spinner) to match the terminal aesthetic.

The indicator should appear as soon as the message is sent, and be removed instantly when the first text token or tool call is received from the backend.

## Additional Context

- [x] I have searched existing issues to ensure this is not a duplicate
- [x] This feature aligns with the project's goals (local-first AI assistance)

Kartik-2239:
hehy @akramcodez can i do this?

akramcodez:
sure @Kartik-2239 go for it :)

PR #817, opened 2026-08-07
Added visual loading indicator for vscode extension

Kartik-2239:
… loader for extension

## Description

This PR adds a small visual loading indicator for the chat ui in vscode extension in the ui (#804 ), it adds a wrapper div and uses css animation. It appears on sending a message and disappears right before the first chunk of response or stopping or error.

[video.mov](https://github.com/user-attachments/assets/9990cb67-66af-4522-8cd2-546bddb1ed86)

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

## Changeset

### Automated Tests

It only uses the DOM so no tests added.

### Manual Testing

- [x] Tested with Ollama
- [x] Tested with OpenRouter

## Checklist

- [x] Self-review completed


akramcodez:
Hey @Kartik-2239, thanks for the this! It looks good, but we're aiming for something a bit more unique that feels like NanoCoder. As mentioned in the issue, we'd prefer a loading animation built around the NC logo or branding rather than a comman one. That would make the experience feel more recognizable and consistent with NanoCoder's identity.

Kartik-2239:
Oh alright i was actually a little confused by it because i didn't know how exactly to implement a loading animation around the logo, should i use an svg? or something like that?
nevermind i got it

Kartik-2239:
Hey @akramcodez,
I made some changes and tried to make it use the given icon.svg with a pulse effect.
I would like to get your feedback if I should make any more changes.
Thanks

https://github.com/user-attachments/assets/3c8415a5-d8ff-4b2a-a042-2f238645c009

akramcodez:
Thanks, @Kartik-2239! Everything looks perfect. Could you just add a changeset? Other than that, everything looks good to me.

Kartik-2239:
Done, thanks

akramcodez (dismissed):
LGTM again great work @Kartik-2239

will-lamerton:
Hey @Kartik-2239 - thanks for this, the loader is a nice addition and the video looks great. A few things to fix before merge:

**1. CSP regression.** The new `img-src {{cspSource}}` is a second `img-src` directive, and CSP ignores every duplicate after the first, so the original `'self' data: https: vscode-webview-resource:` list is dropped. The chat panel renders attachments from data URIs (`chat-panel.js:359`, `:387`, `:642`), so image previews and the modal break. Please merge `{{cspSource}}` into the single existing `img-src` rather than adding a second one.

**2. The loader is only removed in `appendChunk`.** That covers the first agent text chunk, but not the stop button (`:245`), `prompt_response`/`done` with no text (`:1003`), errors delivered via `appendMessage`, or `case 'clear'` (which leaves `visualLoader` pointing at a detached node). Easiest fix is a `stopVisualLoader()` helper called from `setProcessing(false)` as well.

**3. Double loaders when queueing.** `startVisualLoader()` is called outside the `if (!isProcessing)` guard at `:470`, so sending a second message mid-turn appends another wrapper and orphans the first. Either move it inside the guard or clear any existing loader first.

**Suggestion:** consider inlining the icon into the `ICONS` map like every other icon in this file instead of fetching `icon.svg` at runtime. That removes the need for the `connect-src` change (and most of the CSP edit), the duplicated fetch, and the race between the top-level fetch and the fallback inside `startVisualLoader`.

Smaller notes:
- `media/chat-panel.css` is a checked-in Tailwind build and doesn't contain `animate-pulse`, `gap-1.5`, or the `[&_svg]:w-6/h-6` variants yet, so the pulse is invisible until someone runs `pnpm run build`. Worth regenerating it in this PR.
- The loader is appended to `messagesContainer`, so tool cards and thought boxes land below it and it ends up stranded mid-transcript until the first text chunk.
- `startVisualLoader();` at `:470` is space-indented, and there's a `}else{` plus a missing semicolon in the new function. Biome skips `plugins/`, so these need fixing by hand.
- Changeset is `minor`; comparable VS Code UI features have shipped as `patch` (see `add-vscode-copy-button.md`). Also small typo: "nancoder".

Kartik-2239:
Hey @will-lamerton, thanks for going through the changes. 
I have made the necessary changes accordingly.
- regenerated the css
- removed connect-src and img-src (duplicate)
- starting and stopping of visualLoader according to the given review fixes.
- and fixed the changeset

akramcodez:
Hey @Kartik-2239, this PR still has conflicts. Can you fix them?

Kartik-2239:
Hey @akramcodez i fixed the conflicts. It was the css since it was regenerated, i didn't see it before sorry for that

akramcodez:
Thanks for the updates @Kartik-2239! Almost everything is looking perfect.

There's just one tiny detail left from Will's review: in startVisualLoader(), the loader wrapper is still being appended directly to messagesContainer. This causes tool cards and thought boxes to render below the loader when they arrive before the text, leaving the loader stranded mid-transcript.

Could you update it so it gets appended to the active turn element instead (or similar) so it stays at the bottom? Let me know if you need any help with that!

Kartik-2239:
Hey @akramcodez,
The original issue said the spinner should disappear once first chunk is printed or a tool call is made.
And i implemented this, it doesn't look that good.

I did this by appending the same loader (saved as `visualLoader`) at the end of `handleAcpUpdate` so it pushes the spinner at the end after every update.

https://github.com/user-attachments/assets/18505f01-5cfe-4856-b2b6-38edbcb611a3

will-lamerton:
Hey @Kartik-2239 - love this. I think the only thing I don't like is that the NC logo stays showing as text and tools stream. I think as you said, once the first token streams, the loading indicator should disappear. What do you think?

will-lamerton:
Looking great. Thanks for this work @Kartik-2239 - feel free to add yourself as a contributor to our website via a PR which I will approve :)

PR #820, opened 2026-08-07
fix: normalize Atlas Cloud GPT-5.6 model IDs

### LINKED ISSUE #803: [Bug] "invalid request params" (400) when using Atlas Cloud API with GPT models
## Description

When configuring Nanocoder to use the Atlas Cloud API (`https://api.atlascloud.ai/v1`) with GPT models (e.g., `gpt-5.6-sol`), the client fails with a `400 Bad Request` error. The Atlas Cloud API returns `{"code":400,"msg":"invalid request params"}`.

This indicates that Nanocoder's OpenAI-compatible payload format is sending a parameter that the Atlas Cloud endpoint rejects for these GPT models (possibly related to system prompts, top_p, or tool definitions).

## Environment

- **OS**: linux
- **Nanocoder version**: 1.28.1
- **Provider**: OpenAI-compatible API (Atlas Cloud)
- **Model**: `gpt-5.6-sol`

## Steps to Reproduce

1. Configure Nanocoder to use Atlas Cloud as an OpenAI-compatible provider.
2. Start a session using a GPT model via VS Code or CLI: `nanocoder --provider openai-compatible --model gpt-5.6-sol`
3. Enter any prompt and press Enter.
4. Observe the immediate API rejection.

## Expected Behavior

Nanocoder should successfully map the payload to a format that Atlas Cloud accepts for GPT models, allowing the chat completion to proceed normally.

## Actual Behavior

The request is rejected by the server with a 400 Bad Request error complaining about invalid request parameters.

## Logs/Screenshots

```
Error: Bad request: URL: https://api.atlascloud.ai/v1/chat/completions
Response body: {"code":400,"msg":"invalid request params"}
```

Image

## Additional Context

RealBhupesh:
## Summary

- change the Atlas Cloud wizard default to the provider-qualified `openai/gpt-5.6-sol` model ID
- normalize legacy unqualified GPT-5.6 Atlas configs only at the OpenAI-compatible request boundary
- preserve qualified IDs, unrelated models, and other providers
- add template, normalization, and provider-wiring regression coverage plus a patch changeset

## Root-cause status

Atlas documents the model as `openai/gpt-5.6-sol`, while NanoCoder's wizard previously stored `gpt-5.6-sol`. The template mismatch is confirmed in the repository and fixed here. An authenticated two-request Atlas API comparison is still pending, so this PR does not claim the endpoint rejection is independently confirmed yet.

Closes #803

## Validation

- focused template, model-ID, and AI SDK client suites: 51 passed
- regression mutation: the provider-wiring test fails when the raw model is restored
- format and diff checks pass
- full branch CI is temporarily blocked by the ACP SDK regression on current `main`; prerequisite fix: #830

### DIFF
--- a/.changeset/bright-models-fix.md
+++ b/.changeset/bright-models-fix.md
@@ -1,0 +1,5 @@
+---
+'@nanocollective/nanocoder': patch
+---
+
+Fixed the Atlas Cloud wizard to store provider-qualified GPT-5.6 model IDs, while preserving compatibility with existing shorthand configurations. Thanks to @RealBhupesh. Closes #803.
--- a/source/ai-sdk-client/ai-sdk-client.spec.ts
+++ b/source/ai-sdk-client/ai-sdk-client.spec.ts
@@ -1,6 +1,7 @@
 import test from 'ava';
 import type {AIProviderConfig} from '@/types/index';
 import {AISDKClient} from './ai-sdk-client.js';
+import type {TaggedProvider} from './providers/provider-factory.js';
 
 test('AISDKClient constructor initializes with config', t => {
 	const config: AIProviderConfig = {
@@ -114,4 +115,33 @@ test('AISDKClient handles config without models', t => {
 
 	const client = new AISDKClient(config);
 	t.is(client.getCurrentModel(), '');
+});
+
+test('AISDKClient sends the normalized Atlas model to its provider', async t => {
+	const config: AIProviderConfig = {
+		name: 'Atlas Cloud',
+		type: 'openai',
+		models: ['gpt-5.6-sol'],
+		config: {
+			baseURL: 'https://api.atlascloud.ai/v1',
+			apiKey: 'test-key',
+		},
+	};
+	const client = new AISDKClient(config);
+	let receivedModel: string | undefined;
+	const provider = ((model: string): never => {
+		receivedModel = model;
+		throw new Error('model captured');
+	}) as unknown as Extract<
+		TaggedProvider,
+		{kind: 'openai-compatible'}
+	>['provider'];
+	(client as unknown as {provider: TaggedProvider}).provider = {
+		kind: 'openai-compatible',
+		provider,
+	};
+
+	await t.throwsAsync(client.chat([], {}, {}), {message: 'model captured'});
+
+	t.is(receivedModel, 'openai/gpt-5.6-sol');
 });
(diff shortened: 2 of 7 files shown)


will-lamerton:
Hey @RealBhupesh - thanks for picking this up. Two things before I can merge:

1. Can you confirm the root cause with an actual repro? The reported error is a generic `invalid request params`, and the reporter's own guess was a payload param. A pair of curls against Atlas with the same body, one using `gpt-5.6-sol` and one using `openai/gpt-5.6-sol`, would settle it. Right now the diagnosis comes from the docs page rather than the API.

2. If it is confirmed, the first fix should be our own template. `source/wizards/templates/provider-templates.ts:549` ships `modelDefault: 'gpt-5.6-sol'` for atlas-cloud (asserted in `provider-templates.spec.ts:661`), so the wizard is what writes the broken id. Changing that default fixes new setups at the source, and it also means the stored config, the `/model` display, and the models.dev context lookup all use the id the provider actually accepts, rather than only the wire request.

Happy to keep the normalization on top as relief for configs already written, with a comment saying that's its purpose. A few smaller notes if it stays:

- `model.includes('/')` in `normalizeModelIdForRequest` is unreachable, the anchored regex can't match a string with a slash.
- Computing `requestModel` above the switch makes it look like it applies to codex/copilot too. Normalizing inside the `openai-compatible` arm keeps the diff to one line and reads clearer.
- The only production change outside the new file is the wiring in `ai-sdk-client.ts`, and nothing tests it. `ai-sdk-client.spec.ts` is unchanged despite being in the validation list. A test with a fake provider capturing the model argument would catch someone reintroducing `this.currentModel` there later.
- Worth noting the fix only covers three aliases on one family. Atlas hosts plenty of other prefixed models, so any other shorthand fails the same way with no hint. Not blocking, but the template fix generalizes better than the allowlist.

RealBhupesh:
Thanks for the detailed review. I rebased the branch and addressed the code-side items in `1e58e9f1`:

- changed the Atlas wizard default to `openai/gpt-5.6-sol` and updated its template test
- retained normalization only as compatibility relief for existing configs written with the old shorthand
- removed the unreachable slash check and moved normalization into the OpenAI-compatible branch
- added an integration test with a fake provider that captures the model actually passed to the SDK; the test fails if the wiring reverts to the raw configured value

I cannot honestly claim the authenticated Atlas repro yet because I do not have an Atlas credential in this environment. @akramcodez, could you confirm whether the same request body fails with `gpt-5.6-sol` and succeeds with `openai/gpt-5.6-sol`? I will keep the root-cause wording qualified until that live result is available.

The fresh CI failures are inherited from current `main` after the ACP SDK upgrade in #822. I isolated that baseline failure and opened #830 with a regression test and a full passing local gate.

PR #828, opened 2026-08-08
fix: truncate LLM-facing command output head+tail instead of head-only

### LINKED ISSUE #764: Bash output truncation is head-only, discarding the tail where build and test errors live

addyCooks:
## Description

Replace head-only output truncation with a shared `truncateOutputForLLM` helper that keeps both the head (~40%) and tail (~60%) of command output. Compilers and test runners print errors and failure summaries at the end, so head-only truncation discards exactly the part the model needs, forcing an extra round-trip. Same byte budget, much higher information density.

Both `source/tools/execute-bash.tsx` and `source/custom-tools/handler.ts` now use this shared helper instead of their own inline head-only truncation.

Fixes #764

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Changeset

### Automated Tests

### Manual Testing

- [ ] Tested with Ollama
- [ ] Tested with OpenRouter

## Checklist

- [x] Self-review completed


will-lamerton:
Hey @Aditya234892 - thanks so much for this PR however, it was already addressed in an earlier PR. Appreciate your work though and hopefully there's another issue you're happy to contribute to :)

PR #835, opened 2026-08-09
fix: announce rotated Copilot reasoning item ids before the SDK parse…

### LINKED ISSUE #719: [Bug] Cannot read properties of undefined (reading 'summaryParts') when using Github Copilot with GPT-5.3 Codex
## Description

Brief description of the bug

## Environment

- **OS**: Fedora Linux 44
- **Node version**: v24.18.0
- **Provider**: Github Copilot
- **Model**: gpt-5.3-codex

## Steps to Reproduce

1. Have this `agents.config.json`
```json
{
  "nanocoder": {
    "providers": [
      {
        "name": "GitHub Copilot",
        "models": [
          "gpt-5.3-codex"
        ],
        "baseUrl": "https://api.githubcopilot.com",
        "sdkProvider": "github-copilot"
      }
    ],
    "modeProviders": {
      "normal": {
        "provider": "GitHub Copilot",
        "model": "gpt-5.3-codex"
      },
      "plan": {
        "provider": "GitHub Copilot",
        "model": "gpt-5.3-codex"
      }
    }
  }
}
```
2. Login using `nanocoder copilot login`
3. Run `nanocoder`
4. Enter any prompt

## Expected Behavior

After some waiting I get the LLM's answer

## Actual Behavior

After some waiting, I'm getting `Cannot read properties of undefined (reading 'summaryParts')` error

## Logs/Screenshots

Image

## Additional Context

The error happens both with brew version and npm version. Both in normal terminal, and in zellij

  - There is [an issue about GPT-5](https://github.com/Nano-Collective/nanocoder/issues/576), but it is about GPT-5.5 (not codex), it is resolved, the fix is merged, but for me the problem is still ther

addyCooks:
Closes #719

## The bug

Copilot's Responses proxy emits reasoning summary events under an `item_id` that was
never announced via `response.output_item.added` — a rotated id, or no announcement at
all. Reproduced against 3.0.91:
response.output_item.added            item.id = rs_A
response.reasoning_summary_part.added item_id = rs_B   ← never announced
→ TypeError: Cannot read properties of undefined (reading 'summaryParts')

## The fix

The throw happens inside the SDK's own stream transform, so no middleware or model
wrapper can intercept it — it has to be corrected at the SSE layer. `createUndiciFetch`
in this same file already sets that precedent for the `data:  [DONE]` bug.

`createReasoningItemNormalizer()` is applied to Copilot's `text/event-stream` responses.
One rule: when a reasoning event references an unannounced `item_id`, emit a synthetic
`response.output_item.added` ahead of it so the parser always has state. This ships in
`dist`, so it holds regardless of which `@ai-sdk/openai` the user resolves.

### DIFF
--- a/.changeset/fix-copilot-reasoning-item-id-rotation.md
+++ b/.changeset/fix-copilot-reasoning-item-id-rotation.md
@@ -1,0 +1,5 @@
+---
+"@nanocollective/nanocoder": patch
+---
+
+Fixed `Cannot read properties of undefined (reading 'summaryParts')` when streaming from GitHub Copilot with reasoning models such as `gpt-5.3-codex`. Copilot's Responses API proxy emits reasoning summary events under item ids it never announced, which crashed the OpenAI Responses stream parser; the Copilot response stream now announces those reasoning items before the summary events reach the parser. Closes #719.
--- a/.changeset/fix-copilot-reasoning-item-id-rotation.md
+++ b/.changeset/fix-copilot-reasoning-item-id-rotation.md
@@ -2,4 +2,4 @@
 "@nanocollective/nanocoder": patch
 ---
 
-Fixed `Cannot read properties of undefined (reading 'summaryParts')` when streaming from GitHub Copilot with reasoning models such as `gpt-5.3-codex`. Copilot's Responses API proxy emits reasoning summary events under item ids it never announced, which crashed the OpenAI Responses stream parser; the Copilot response stream now announces those reasoning items before the summary events reach the parser. Closes #719.
+Fixed `Cannot read properties of undefined (reading 'summaryParts')` when streaming from GitHub Copilot with reasoning models such as `gpt-5.3-codex`. Copilot's Responses API proxy rotates the opaque reasoning item id mid-stream while `output_index` stays stable, so the OpenAI Responses parser looked up state that was never registered and the stream died. Copilot's response stream is now normalized before it reaches the parser: a rotated id is mapped back to the reasoning item already tracked at that `output_index`, and a reasoning item that was never announced is announced first. Closes #719.
--- a/source/ai-sdk-client/providers/provider-factory.spec.ts
+++ b/source/ai-sdk-client/providers/provider-factory.spec.ts
@@ -525,7 +525,7 @@ function eventSummary(sse: string): string[] {
 		});
 }
 
-test('createReasoningItemNormalizer announces reasoning items with a rotated item_id', async t => {
+test('createReasoningItemNormalizer maps a rotated item_id back to the item at that output_index', async t => {
 	const output = await runNormalizer([
 		sseEvent({
 			type: 'response.output_item.added',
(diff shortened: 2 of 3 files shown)


addyCooks:
cc @will-lamerton

will-lamerton:
Hey @addyCooks - thanks for this - the SSE-layer approach is right, and I checked that the response.reasoning_summary prefix covers all three activeReasoning dereference sites in @ai-sdk/openai@3.0.71, so the crash surface is fully handled.

One thing I'd like to sort before merging... rewrite() bails on !trimmed.startsWith('data:'), so if Copilot frames its stream the way the upstream Responses API does, with an event: <type> line ahead of each data: line, the normalizer is a silent no-op. I ran that exact input through the branch and got byte-identical passthrough: nothing announced, nothing rewritten, and all 5 tests still green because they're all data-only. Do you have a raw capture from your repro that shows the framing? I tried to confirm against live Copilot but /responses is gated on my account.

Related: separator is hardcoded to '\n\n', and \r\n\r\n contains no two consecutive \n, so a CRLF-framed stream never splits. I confirmed it buffers the entire response and flushes in one go at close, which would stall streaming output on top of skipping the fix. Splitting on /\r\n\r\n|\n\n/ and scanning the block for its data: line would make both cases safe regardless of what Copilot actually sends.

Two smaller ones:

- flush() enqueues the trailing buffer raw, so a final event with no trailing separator skips normalization. Wrapping it in rewrite() covers it.
- On the never-announced path, announce() runs even when output_index isn't a number, and JSON.stringify then drops the key. The SDK schema requires output_index, so the synthetic chunk fails validation rather than repairing anything. Worth guarding.

Happy to push these if it's easier.

addyCooks:
Hi @will-lamerton ,
I don't have raw bytes. I traced it off the SDK's parsed chunk stream, so I genuinely can't tell whether there's an event: line in front of each data, that detail is already gone by the point I was looking. Good catch.

After the normalizer, the SDK runs its own SSE parser, and that one already handles event: lines and every line-ending style. Mine is stricter than the thing consuming its output, which is wrong no matter what Copilot sends. I'll split on /\r\n\r\n|\n\n/ and find the data: line inside each block so both framings work.

One thing that makes this easier, the SDK's parser only reads the data: line and ignores the event: name, so the synthetic announcement can stay a plain data: block.

On output_index....it's required (z.number()), and I checked what happens when it's missing. The SDK sets finishReason = 'error' and emits an error part. So the synthetic chunk doesn't just fail to help, it breaks the response another way. 
I'm thinking to skip the rewrite entirely when the field isn't there rather than guess an index.

flush() through rewrite(), agreed.

I'll push these.
Adding tests for event: framing, CRLF, and a final event with no trailing separator the existing five all use plain data with \n\n, which is how this slipped through.

will-lamerton:
Nice job @addyCooks - looks good and merging! Quick one, I had @akramcodez mention to me about taking issues and submitting PRs without asking for assignment first. Although your contributions are great and much appreciated - if you wouldn't mind dropping a comment on the issue asking for assignment before beginning work that would be great :)

addyCooks:
Hey @will-lamerton,
moving forward I'll handle it that way
do let me know if there's anything specific to help with (maybe a few task?)
also really loving the codebase

PR #842, opened 2026-08-10
Feat/vscode context mentions

### LINKED ISSUE #747: [Feature] Context-Aware Mentions (`@` referencing) in VS Code Extension
## Description
Implement an `@` mention system inside the VS Code extension's chat input field, allowing users to explicitly attach specific context to their prompt.

## Use Case
When a codebase is large, the AI shouldn't automatically send the entire context. Users of the VS Code extension need a fast, intuitive way to tell the AI exactly which files, open editors, or folders to look at for a specific prompt.

## Proposed Solution
Typing `@` in the `chat-input` textarea should display a floating autocomplete dropdown. Users can search and select workspace files, folders, or terminal outputs. The selected context is appended to the payload sent to the AI.

## Alternatives Considered
- A separate "Attach Context" button that opens a native file picker, but this breaks the user's typing flow.

## Additional Context

- **UI:** Use a webview-side floating dropdown in `plugins/vscode/media/chat-panel.html` similar to the existing provider/model dropdowns.
- **Backend:** The extension needs a message listener in `plugins/vscode/src/chat-webview-provider.ts` to query the workspace for file paths and return them to the webview for autocomplete filtering.

## Contribution Guidelines

addyCooks:
Closes #747.

## Summary

Implements @ mention autocomplete in the VS Code extension's chat composer, allowing users to explicitly attach workspace files, folders, and open editors as context. Typing `@` displays a dropdown with fuzzy-matched suggestions; selecting an item attaches it as a chip that's expanded into the AI's context on submit.

### New Modules

- **`mention-search.ts`**
- **`mention-search.spec.ts`** 
- **`mention-utils.js`** 
- **`mention-utils.spec.ts`** 
- **`context-attachment.ts`**
- **`context-attachment.spec.ts`**

### Modified Files

- **`webview-protocol.ts`** 
- **`chat-webview-provider.ts`**
- **`chat-panel.js`**
- **`chat-panel.html`** 
- **`.changeset/vscode-context-mentions.md`** 

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

## Changeset

### Automated Tests

### Manual Testing

- [x] Tested with Ollama
- [ ] Tested with OpenRouter

###Checklist
- [x] Self-review completed


will-lamerton:
Hey @addyCooks - thanks for this, the design is solid - host-side search, the request-id staleness guard, mousedown over click, and folding the keys into the existing `keydown` listener are all the right calls.

A few things before merge:

**The description doesn't match the diff.** The body describes `mention-search.ts`, `mention-utils.js`, `context-attachment.ts` and 56 tests, but the final `refactor: fold @ mention support into the existing files` commit collapsed all of that back inline and took the specs with it. Six files changed, no specs. That leaves ~480 lines of ranking/glob/caret logic uncovered, and the "New features include passing tests" boxes ticked for tests that aren't there. The split you originally described was the better structure - could you restore it?

**The 100 KB read cap isn't in the diff either.** `_expandContextAttachments` is still a bare `readFileSync` with no size check. Your reasoning for adding it was right and matters more now that `@` makes hitting a lockfile one keystroke.

**Escape while streaming kills the agent turn.** The mention Escape branch calls `preventDefault()` but not `stopPropagation()`, so it bubbles to the `document` handler that cancels on Escape when `isProcessing`. Drafting a message mid-stream and dismissing the dropdown cancels the run.

**`files.exclude` isn't actually honoured.** Passing an explicit `exclude` glob to `findFiles` replaces the default excludes rather than adding to them (and `search.exclude` never applies to `findFiles`). So `.env` and friends show up in the dropdown, and the changeset currently claims the opposite in the public changelog.

**Stale `mentionToken`.** It's only assigned inside `requestMentions`, after the `mentionLastQuery === token.query` early return. With `@foo @foo`, moving the caret from the second mention to the first short-circuits the dedupe and leaves `start` pointing at the second one, so accepting adds the chip but leaves the literal `@foo` in the message. Assigning it in `syncMentionState` fixes it.

Smaller stuff: accepting mid-token leaves the suffix behind (`@src/foo` with the caret after `src` leaves `/foo`); the `absPath.toLowerCase()` dedupe key collides case-sensitive siblings; `@**` degrades to a full workspace scan that then filters to nothing; and the listbox has `role="option"` but no `aria-activedescendant` linking it to the textarea. The `styled-select-input.spec.tsx` change is a no-op on main (I ran it) and unrelated to this feature, so probably worth splitting out.

addyCooks:
Hey @will-lamerton,
Thank you for the detailed review again : )
I've addressed all 10 issues

The PR description has been updated to accurately reflect the current implementation.
All tests pass, and the feature is ready for re-review.

will-lamerton (approved):
Re-reviewed. All 10 points from my earlier review are addressed:

- Split restored as `mention-search.ts` / `mention-utils.js` / `context-attachment.ts` with 67 passing specs, and `package.json` now globs `plugins/**` into ava so they actually run.
- 100 KB read cap, binary sniff, explicit truncation marker, and a 200-entry folder cap in `readCappedFile` / `readCappedDirectory`.
- Escape mid-stream now calls `stopPropagation()`; verified against the bubble-phase `document` handler main has since gained.
- `_mentionExcludeGlob` folds `files.exclude` and `search.exclude` into the exclude list by hand, folder-scoped, skipping non-boolean `when` entries.
- `mentionToken` is assigned in `syncMentionState` ahead of the dedupe early return.
- `mentionTokenEnd` takes the whole token so a mid-token accept leaves no suffix.
- Dedupe keys on `caseInsensitiveFs` rather than folding case unconditionally.
- A glob-only query answers from open editors instead of scanning the workspace.
- `aria-activedescendant` wired up, plus `role=combobox` / `aria-expanded` / `aria-controls` and per-row ids.
- The `styled-select-input.spec.tsx` change landed on main independently, so it is a no-op here.

I pushed a merge of main into the branch to clear the conflicts in `chat-panel.js`: `addImageBtn` was removed by the upload consolidation, and `closeAllDropdowns` was extracted, so the mention block now sits after it. Also corrected two comments main had made stale - the Escape branch now names the handler it stops propagating to, and the `requestMentionCompletions` docblock no longer claims `findFiles` honours `files.exclude` on its own.

Verified locally at the merge commit: root and extension typechecks clean, biome check and lint clean, extension bundles, 67/67 specs pass.

Thanks for the thorough follow-up on this.

addyCooks:
Thanks for the thorough re-review, @will-lamerton! 
Glad I could address all the points and get everything verified. 
Really appreciate the detailed feedback throughout the review, 
it helped improve the implementation quite a bit.

PR #844, opened 2026-08-11
fix: preserve newer pending file reads

### LINKED ISSUE #840: [Bug] Race Condition in Concurrent File Cache Reads
### Description
The file cache (`source/utils/file-cache.ts`) implements request deduplication via a `pendingReads` map, ensuring that concurrent requests for the same file wait on a single disk read instead of duplicating the I/O. However, in the cache invalidation block (when `mtime` changes), the `finally { pendingReads.delete(absPath); }` cleanup block is executed by *every* awaiting request, rather than just the request that initiated the read.

Because microtasks in JavaScript can interleave, this creates a race condition:
1. Request A misses the cache, creates a pending read, and awaits it.
2. Request B also wants the same file, finds the pending read, and awaits it.
3. The read finishes. Request A resumes, gets the file content, and deletes the pending read in its `finally` block.
4. Request C comes in right after, misses the cache, and creates a *new* pending read.
5. Request B now resumes (from step 3), finishes awaiting, and executes its own `finally` block, **deleting Request C's new pending read**.
6. This breaks deduplication for Request C and any subsequent requests, which will now spawn duplicate I/O operations because the pending state was prematurely deleted.

### Environment
- **OS**: Any 
- **Node version**: v18+
- **Nanocoder version**: Latest (main branch)

### Steps to Reproduce
(This is a microtask race condition that is easiest to reproduce programmatically)
1. Write to a file to invalidate its cache.
2. Trigger two concurrent `getCachedFileContent` calls (A and B).
3. Chain a `.then()` to A that immediately triggers a third call (C).
4. Observe that C's `pendingReads` entry is deleted by B's `finally` block when B resolves.

### Expected Behavior
Only the request that *creates* the pending read should be responsible for cleaning it up from `pendingReads` in its `finally` block.

### Actual Behavior
All awaiting requests blindly call `pendingReads.delete(absPath)`, potentially deleting subsequent unrelated pending reads for the same path.

### Logs/Screenshots
*N/A*

### Additional Context
- Occurs specifically in: `source/utils/file-cache.ts` in the `mtime` invalidation block (`getCachedFileContent`).

mikemikimike:
## Summary

- Make pending-read cleanup owner-aware on the mtime invalidation path.
- Only remove a pending read when the map still points to the same Promise.
- Add a deterministic regression test that prevents an older read from clearing a replacement read.

Fixes #840

## Validation

- `pnpm run test:ava source/utils/file-cache.spec.ts` — 19 passed
- `pnpm test:format` — passed
- `pnpm test:types` — passed
- `pnpm test:lint` — passed
- `pnpm test:knip` — passed
- `pnpm audit --registry=https://registry.npmjs.org --audit-level=high --ignore-unfixable` — passed

The repository's `pnpm run test:all` wrapper cannot start on Windows because it invokes `./scripts/test.sh`. Running the equivalent commands individually reached existing full-suite failures/timeouts in `acp-content.spec.ts` and `ai-sdk-client` tests; the focused file-cache suite passes.

### DIFF
--- a/.changeset/quiet-rays-cache.md
+++ b/.changeset/quiet-rays-cache.md
@@ -1,0 +1,5 @@
+---
+"@nanocollective/nanocoder": patch
+---
+
+Prevent concurrent file-cache reads from clearing a newer pending read.
--- a/source/utils/file-cache.spec.ts
+++ b/source/utils/file-cache.spec.ts
@@ -1,5 +1,6 @@
 import test from 'ava';
-import {mkdtemp, rm, writeFile} from 'node:fs/promises';
+import fs, {mkdtemp, rm, writeFile} from 'node:fs/promises';
+import {mock} from 'node:test';
 import {tmpdir} from 'node:os';
 import {join} from 'node:path';
 import {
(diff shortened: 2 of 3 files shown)


will-lamerton:
Nice work @mikemikimike :)

PR #845, opened 2026-08-11
fix: don't clear a newer file-cache pending read on cleanup

### LINKED ISSUE #840: [Bug] Race Condition in Concurrent File Cache Reads

kishore280:
## Description

Closes #840. `pendingReads.delete(absPath)` ran unconditionally in every
awaiter's `finally` block. When a fresh read started for the same path
before an older one finished, the older read's cleanup deleted the newer
entry, breaking dedup and forcing an unrelated caller into an extra disk
read.

Cleanup now only clears the `pendingReads` entry when it's still the one
this call created, checked by identity rather than by key alone.

Note: #844 is already open for this issue. Same core fix, but without the
redundant `ownsPendingRead` flag (the identity check alone already proves
ownership) and with a regression test that hangs on the old code and
passes on the new one.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Changeset

### Automated Tests

### Manual Testing

- [ ] Tested with Ollama
- [ ] Tested with OpenRouter
- [x] Tested MCP integration (if applicable) — n/a, internal cache-layer change only

## Checklist

- [x] Self-review completed
- [x] Appropriate logging added using structured logging (see CONTRIBUTING.md#logging) — n/a, no logging in this path


will-lamerton:
Hi @kishore280 - thanks for the PR however, an earlier submitted PR solved this issue. Appreciate you taking it on and if there's another issue you'd like to be assigned to let me know!

PR #847, opened 2026-08-11
feature: show queued and running agent actions in the chat panel

### LINKED ISSUE #752: [Feature] Visual Agentic Action Planner in VS Code Extension
## Description
Display a visual roadmap or checklist in the chat panel representing the AI agent's current plan and running tools (e.g., "Reading file X", "Running command Y").

## Use Case
When using Nanocoder for complex tasks, the AI often executes multiple tool calls behind the scenes. Without a visual indicator, the user is left waiting, unsure if the AI is frozen, thinking, or actively reading files. A transparent "action planner" builds trust and makes the tool feel like a true autonomous agent.

## Proposed Solution
Inject a temporary, animated "status block" into the chat stream when the agent begins working. As `NanocoderAcpClient` receives updates about tool calls being executed, update this block to show a checklist of completed and currently executing steps.

## Alternatives Considered
- Printing raw JSON tool payloads to the chat, which is too noisy and technical for general users.

## Additional Context

- **Backend:** Update `chat-webview-provider.ts` to emit specific `acpUpdate` messages when a tool call starts and finishes (intercepting the ACP lifecycle).
- **UI:** In `chat-panel.js`, render an interactive UI block (perhaps an accordion) that dynamically populates with the agent's current actions.

## Contribution Guidelines

addyCooks:
## Description

Makes the VS Code chat panel show what the agent is doing while it works, instead of leaving the
user on a silent turn.

Closes #752

## Type of Change

- [x] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

## Changeset

### Automated Tests

### Manual Testing

- [x] Tested with Ollama
- [x] Tested with OpenRouter
- [x] Tested MCP integration (if applicable)

## Checklist

- [x] Self-review completed


will-lamerton:
Thanks for this @addyCooks, the direction is right and the `chat-webview-provider.ts` id fix is the key one. `onSessionUpdate` gets the full `{sessionId, update}` params, so `payload.toolCallId` was always undefined and diffs were being registered under the file path while the panel asked for them by tool call id. Routing edit cards on ACP `kind` instead of tool names is also the correct call, since the old list matched names we never emit and missed `string_replace` entirely. I ran the ACP specs on your branch and all 26 pass, plus the contract spec, and biome is clean.

A few things I'd like addressed before merge:

- Announced calls hang in `pending` forever when the turn returns early on `permission === 'cancelled'` (acp-conversation.ts:311-321). The abort branch handles this case, the permission-cancel path doesn't. VS Code masks it via the `setProcessing` sweep, but Zed will show stuck rows.
- The edit card is created with an active "Open Diff" button, but the announcement sends `content: []`, so nothing is registered with `DiffManager` until the second emit. Clicking in that window pops `Change <id> not found`, and in approval mode that window is as long as the user takes to approve.
- Queued edit cards read "Edited <file>" before the tool has run or been approved. Worth making the label status-aware.

Smaller things, happy for them to be follow-ups:

- Each announced call sends `tool_call` twice (announce, then again at line 275). We upsert by id so it's invisible in VS Code, but a client that appends on `tool_call` could double-render. Switching the second to `tool_call_update` would be cleaner.
- `buildToolCallMeta` now runs twice per call, and for `string_replace`/`write_file` the announce pass reads the whole file to build a diff it immediately discards. A `withDiff: false` option would skip that.
- `fetch_url` and `web_search` take `url`/`query` rather than `path`, so their titles never contain `": "` and those `TOOL_VERBS` entries can't fire. `string_replace`/`write_file` are dead entries too since `kind === 'edit'` routes them to the edit card.
- `ToolAggregator.cancelPending()` still sniffs `animate-spin`, the same pattern you replaced in `setProcessing`, so it wouldn't catch the new `pending` rows. It's uncalled today, so deleting it is fine.

Also worth a test on the new abort branch emitting `failed` for announced-but-unrun calls.

addyCooks:
Hi @will-lamerton, 
Thanks for the detailed review! I've addressed everything.
Should be good to go!

will-lamerton:
Good to go! Nice work @addyCooks

PR #848, opened 2026-08-11
feat(settings): sunset /setup-providers and /setup-mcp in favour of /…

### LINKED ISSUE #832: Sunset /setup-providers and /setup-mcp in favour of /settings
## Summary

`/setup-providers` and `/setup-mcp` predate `/settings`. Now that `/settings` has a proper tabbed UI with a Providers tab that already lists configured providers and MCP servers, having two separate top-level entry points into the same wizards is redundant and confusing. We should sunset the two `/setup-*` commands and make `/settings` the single place to configure providers and MCP servers.

## Current state

Both commands are "special commands": thin stubs in `source/commands/setup-providers.tsx` and `source/commands/setup-mcp.tsx` registered for slash-menu discovery, dispatched in `source/app/utils/app-util.ts` via `SPECIAL_COMMANDS.SETUP_PROVIDERS` / `SPECIAL_COMMANDS.SETUP_MCP`, which call `onEnterConfigWizardMode()` / `onEnterMcpWizardMode()`. Those set app modes `configWizard` / `mcpWizard` in `useModeHandlers`, rendered by `source/app/components/modal-selectors.tsx`.

`/settings` reaches the same wizards by a different route: `settings-tabs.tsx` -> Providers tab -> `settings-providers-list.tsx` / `settings-mcp-list.tsx`, each of which renders `ProviderWizard` / `McpWizard` inline once you pick an entry.

So we have two parallel plumbing paths into one pair of wizards, with slightly different behaviour on each:

- The settings path refreshes running MCP connections after save (`onMcpChanged`); the standalone `/setup-mcp` path goes through `handleMcpWizardComplete` in `useModeHandlers` instead.
- The settings list panels are list-first but every row currently opens the same "add or edit" wizard flow. Selecting a specific provider or server does not deep-link into editing that entry.

## Proposed work

1. Remove `/setup-providers` and `/setup-mcp` from the slash-command registry (`source/commands/lazy-registry.ts`) and delete the two stub command files.
2. Retire the `SETUP_PROVIDERS` / `SETUP_MCP` special-command entries and the `configWizard` / `mcpWizard` app modes, along with `onEnterConfigWizardMode` / `onEnterMcpWizardMode` in `source/types/app.ts`, `useAppHandlers`, and `useModeHandlers`, once nothing else needs them. Keep whatever the first-run / no-provider bootstrap path relies on.
3. Make the settings panels the real home for this. Concretely: selecting a provider or server row should edit that entry, with explicit add and delete affordances, instead of every row opening the same generic wizard.
4. Unify post-save behaviour so both providers and MCP servers reload the running session consistently (MCP reconnect, provider/model reselection) regardless of which panel triggered the save.
5. Consider splitting MCP out of the Providers tab into its own settings tab now that it is a first-class configuration surface.
6. Add short-term aliases or a deprecation notice so existing users typing `/setup-providers` get pointed at `/settings`, rather than a bare "unknown command".
7. Update the docs, which reference the commands in a number of places: `docs/features/commands.md`, `docs/features/index.md`, `docs/getting-started/index.md`, `docs/configuration/mcp-configuration.md`, and every per-provider page under `docs/configuration/providers/`. Also the in-app hint in `source/commands/mcp.tsx:78`.
8. Update the affected tests, including `source/app/utils/app-util.spec.ts`, which asserts on the setup command names.

## Open questions

- Hard removal or a deprecation period with aliases that forward to `/settings`?
- Should `/settings providers` / `/settings mcp` style subarguments exist so the old muscle memory has a direct one-line equivalent?

addyCooks:
## Description

Makes `/settings` the single entry point for provider and MCP configuration, retiring `/setup-providers` and `/setup-mcp`.

Closes #832

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Documentation update

## Changeset

### Automated Tests

### Manual Testing

- [x] Tested with Ollama
- [ ] Tested with OpenRouter

## Checklist

- [x] Self-review completed


will-lamerton:
Nice consolidation @addyCooks, and the retired-alias forwarding is a good touch. tsc, biome, knip and the affected specs all pass on the branch. One thing needs fixing before merge:

**`onProvidersChanged` is wired to `handleConfigWizardComplete`, which wipes the conversation.** That handler was written for first-run onboarding, so it does `setMessages([])` + `clearContext()` and re-resolves the client with a bare `createLLMClient()`. Two consequences when you edit a provider mid-session from `/settings`:

- The model's history is emptied, but the visible transcript stays on screen (it renders from the chat component queue, and this path doesn't bump `clearKey` like `/clear` does). So the user sees their conversation intact while the model has amnesia.
- Provider and model reset to the config default, discarding whatever they picked with `/model`.

`reloadMcpServers` is the pattern you want here: it was deliberately split out of the wizard handler so the settings panel could reuse it without the mode-exit side effects. Providers needs the same, `reloadAppConfig()` plus a client rebuild for the *current* provider/model, leaving `messages` alone.

Smaller things:

- `settings-providers-list.tsx:38-41` calls `onBack()` after `onProvidersChanged` has already closed settings, so it sets state on an unmounting component. Let the parent own the close.
- `returnFromLaunchedWizard` (`interactive-app.tsx:86`) calls `enterSettingsMode()` with no tab, so returning from Tune/IDE always drops you on Appearance rather than the tab you came from.
- Unknown tab args are silently swallowed: `/settings providrs` opens Appearance with no hint the argument was wrong.
- `context-max-handler.spec.ts:15` still passes the now-removed `onEnterConfigWizardMode`.
- `app-util.ts` takes a runtime import of `settings-tabs` just for `SETTINGS_TAB_IDS`, pulling every settings panel and wizard into its graph. No cycle and no `--plain`/ACP cost since only `useAppHandlers` imports it, but that array is pure data and would sit better in a constants module.

Good incidental catch on `source/commands/mcp.tsx` pointing at `/setup-providers`.

addyCooks:
@will-lamerton hey! thanks for the thorough review..
just pushed the fixes you called out, everything should be good now

lmk if anything needs adjusting!

will-lamerton:
Thanks @addyCooks - the five fixes themselves all look right, but the branch has lost the feature it's meant to ship.

`b15db36c` (the review-fix commit) was built on top of the `feat/vscode-context-mentions` lineage rather than the sunset work, and force-pushed to `feat/sunset-setup-commands` on the 14th - its parents are `ac5abf51` and `8b1ac88b`, which are #842's commits. The sunset commits I reviewed, `53e7cd6b` and `28fa5f10`, aren't in the branch history at all any more.

So on the current head: `source/commands/setup-providers.tsx` and `setup-mcp.tsx` still exist, `SPECIAL_COMMANDS.SETUP_PROVIDERS` still maps to `onEnterConfigWizardMode` in `app-util.ts`, and the `lazy-registry.ts` retirement, docs updates, `app-util.spec.ts` additions (+141 lines) and the `/settings` consolidation changeset are all gone. CI is green because it's testing a branch that no longer does the thing.

Now that #842 has merged, a rebase on main will clear the duplicated vscode diff and the conflict, but it won't bring the sunset work back. Could you cherry-pick `53e7cd6b` and `28fa5f10` onto current main, then replay `b15db36c` + `f202d25a` on top?

On the fixes themselves, checked against the current head:

- **Provider reload**: exactly what I was after. `reloadProviders` reloads config and rebuilds the client for the current provider/model with `messages` untouched, and the wiring through `interactive-app` → `ModalSelectors` → `SettingsSelector` → `SettingsProvidersListPanel` is clean.
- **Unmount fix**: good, parent owns the close now.
- **Tab preservation**: good. One thing to keep in mind: `onTabChange` only fires from `goToTab`, so any future path that sets the active tab needs to go through `updateActiveTab` too.
- **Tab validation**: good. Worth simplifying the double cast in `app-util.ts` to a small `isSettingsTabId()` type guard.
- **`settings-constants.ts`**: good, that's the split I wanted.
- **`context-max-handler.spec.ts:15`**: moot on this head since `onEnterConfigWizardMode` is still there, so it'll need doing again once the sunset commits are back.

Two smaller things for the re-push:

- The new code uses inline `import('@/app/components/settings-constants').SettingsTabId` type references in five files. Please use a top level `import type` instead, that's the convention everywhere else.
- None of the five fixes have tests. The tab arg validation in particular (valid tab, invalid tab, no tab) is easy to cover in `app-util.spec.ts`, and `reloadProviders` preserving messages is worth a case in `useModeHandlers.spec.tsx`.

addyCooks:
Sorry about that @will-lamerton you're right, and thanks for digging into the history rather than just bouncing it back. 
The bad force-push was mine and it cost you a full re-review of a branch that wasn't doing the thing any more.

Fixed. 53e7cd6b and 28fa5f10 were still reachable by SHA, so I recovered both, cherry-picked them onto current main, and replayed b15db36c + f202d25a on top. The vscode conflicts cleared with it, since the branch no longer duplicates #842.

Replaying onto the right base exposed two bugs that only existed because of the wrong-base authoring:

settings-constants.ts had no mcp tab id it was extracted from a base predating the sunset commit's MCP tab, so RETIRED_SETUP_COMMANDS['setup-mcp'] = 'mcp' wouldn't compile and /settings mcp would have been rejected as unknown. SETTINGS_TAB_IDS is now canonical and TABS derives from it via a Record<SettingsTabId, string>, so a missing label is a compile error rather than silent drift.
Duplicate onProvidersChanged in interactive-app.tsx the wiring you flagged in the first review survived alongside reloadProviders. Removed.
Your list: isSettingsTabId() replaces the double cast; all inline import(...) types are now top-level import type; onEnterConfigWizardMode dropped from context-max-handler.spec.ts; and on your updateActiveTab point, the raw setter is renamed setActiveTabState so a future path skipping onTabChange stands out.

Tests: app-util.spec.ts covers no arg / valid tab / case-insensitivity / unknown tab / error contents. useModeHandlers.spec.tsx asserts reloadProviders leaves setMessages, setActiveMode and setIsSettingsMode untouched. One behaviour change worth flagging — the sunset commit's own test asserted an unknown tab silently falls back to the default, the thing you objected to, so I flipped it to expect the error.

038aa6f2 also picks up the last #832 item, since this PR closes it: rows opened the same generic wizard with no delete affordance. The steps already had per-entry edit-or-delete, just unreachable — so there's an optional initialEditName threaded panel → wizard → step, keyed by name not index (the panel lists resolved config, the wizard loads one file). BaseConfigWizard untouched, prop optional, first-run onboarding unchanged. Happy to split it out if you'd rather keep this PR to the sunset work.

CI is red but not from this branch  28 failures, all chat-panel-thoughts / chat-panel-tool-cards:


TypeError: Cannot destructure property 'findMentionQuery' of
'globalThis.NanocoderMentionUtils' as it is undefined.
chat-panel-harness.ts:12 boots only chat-panel.js, but #842 moved those helpers into media/mention-utils.js, which the harness never loads. This branch touches no vscode files, and the same 28 fail on clean main. One-line harness fix say the word and I'll send it separately. (CodeQL is a SARIF upload outage, not a finding; a re-run should clear it.)

PR #866, opened 2026-08-12
add Explain Code / Generate Tests code lenses

### LINKED ISSUE #750: [Feature] Inline Editor Code Lenses for AI Actions in VS Code Extension
## Description
Add "Code Lenses" (clickable text above functions or classes) directly inside the VS Code text editor that trigger VS Code extension AI actions.

## Use Case
Users often see a piece of code in their editor and want to interact with the AI about it. Switching to the VS Code extension sidebar and copying/pasting the code is slow. 

## Proposed Solution
Display small code lenses like `[Explain Code]` or `[Generate Tests]` above function definitions. Clicking a lens programmatically opens the webview chat panel, populates the input, and sends the prompt with the selected code as context.

## Alternatives Considered
- Relying solely on right-click context menus, which are less discoverable than code lenses.

## Additional Context

- **Backend:** Implement a standard `vscode.CodeLensProvider` in the `plugins/vscode/src/` directory. The command attached to the lens should trigger an extension command that communicates with `plugins/vscode/src/chat-webview-provider.ts` to focus the webview and set the `chat-input` value.

## Contribution Guidelines

addyCooks:
Closes #750.

## Description

Adds inline `Explain Code` / `Generate Tests` code lenses above every function, method, constructor and class. Clicking one reveals the chat view and submits that symbol as a prompt instruction, a `file:startLine-endLine` locator, and the source fenced with the document's language. Previously, asking the agent about a specific function meant switching to the sidebar and pasting the code in.

Symbols come from `executeDocumentSymbolProvider`, so there is no per-language parsing. Each lens anchors on the symbol's `selectionRange` (the name) while the command carries `symbol.range` (the whole body). Lenses can be turned off with the `nanocoder.codeLens` setting.

Inlined source is capped at 200 lines / 8000 characters, whichever binds first, with a truncation marker, so `Generate Tests` on a large class cannot spend a whole context window on one turn. The `file:startLine-endLine` locator is emitted before the fence and survives truncation, so the agent can read the rest itself.

### Also fixed here

Three latent bugs in `ChatWebviewProvider` that a lens click makes much easier to hit:

- Sending a message while a tool approval was still pending left the composer spinning forever. The webview had already drawn the user bubble and flipped to the loading state, but the early return posted nothing, so no `prompt_response` ever ended the turn. Pre-existing.
- `_isWebviewReady` was not reset on re-resolve, so a queued prompt could post into a fresh shell that had not yet attached its message listener, and be dropped.
- The pending-prompt timer could outlive the view, and a late `onDidDispose` from an already-replaced view could null out the live one.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

## Changeset

### Automated Tests

18 specs across two files:

- `code-lens-provider.spec.ts` (10) - kind filtering plus the nested-children walk, the `codeLens: false` short-circuit, legacy symbols with no `selectionRange`, the prompt string shape, the truncation boundaries (line cap, char cap, and a single line that busts the char cap), and the command invoked with no lens arguments.
- `chat-webview-provider.spec.ts` (8) - a prompt rejected for a pending permission ends the turn, a real turn still ends exactly once, `dispose()` clears the pending timer, a prompt dropped by dispose is not delivered to a later view, a stale view disposal does not clear a newer view, disposing the live view does clear it, a queued prompt survives a re-reveal, and a re-resolved shell is not treated as ready until it says so.

Each fix was checked by reverting it individually and confirming the corresponding spec fails.

### Manual Testing

- [ ] Tested with Ollama
- [ ] Tested with OpenRouter

## Checklist

- [x] Self-review completed


will-lamerton:
Hey @addyCooks - nice work, this is a clean implementation. Using `executeDocumentSymbolProvider` instead of per-language parsing is the right call, and anchoring lenses on `selectionRange` while passing `symbol.range` as the payload is a good detail. The `_isWebviewReady` reset in `resolveWebviewView` is a real latent bug fix on its own.

A few things before merge:

1. **Stuck spinner on rejected prompts.** `_handlePrompt` early-returns when `hasPendingPermissions()` is true, but the webview has already appended the user bubble and called `setProcessing(true)`. Nothing posts `prompt_response`, so the loader spins until the user hits Escape. Pre-existing, but clicking a lens while an approval sits unattended in the sidebar makes it much easier to hit. One-liner in that guard: `this.postMessage({type: 'acpUpdate', update: {sessionUpdate: 'prompt_response'}})`.

2. **Tests.** `plugins/vscode/src/` already has AVA specs (`acp-client.spec.ts`, `acp-process-manager.spec.ts`), but `NanocoderCodeLensProvider` and `sendCodeLensPrompt` are unexported locals in `extension.ts` so they can't be reached from one. Moving them to `src/code-lens-provider.ts` matches the rest of `src/` and makes three cheap tests possible: kind filtering plus the nested `children` walk, the `codeLens: false` short-circuit, and the prompt string shape.

3. **Unbounded inlined source.** `document.getText(range)` has no cap, so `Generate Tests` on a large class inlines the whole thing. Worth a line/char cap with a truncation marker, or falling back to just the `file:start-end` locator past a threshold.

Smaller stuff: `_pendingPromptTimer` is never cleared on dispose, so the timeout warning can fire after the view is gone; and `onDidDispose` should guard with `if (this._view === webviewView)` before nulling. `Constructor` is also missing from `LENS_SYMBOL_KINDS` while every sibling method gets a lens.

addyCooks:
Thanks @will-lamerton! 
All six addressed in 95d7941, plus tests for the three that didn't have any.

One thing worth flagging: the sibling specs cited as precedent were never actually running, the root AVA glob only matched source/, and vscode isn't resolvable outside the extension host. Getting any spec here to run meant widening the glob to plugins/*/src/**/*.spec.ts and adding a vscode stub behind a test-only paths entry. That revives acp-client.spec.ts and acp-process-manager.spec.ts as a side effect. The stub is .vscodeignored and the bundle still builds with --external:vscode. Happy to split that into its own PR if you'd rather keep this one scoped to the lenses.

Unbounded source. Capped at 200 lines / 8000 chars, whichever binds first, with a truncation marker. The file:start-end locator is emitted before the fence and survives truncation.

Timer on dispose. ChatWebviewProvider implements Disposable and is registered with the extension's subscriptions. I deliberately did not clear it in onDidDispose — a disposal is usually a re-reveal in progress, and dropping the queued prompt there would reintroduce what 9a9d5bf fixed. onDidDispose guards on view identity as you suggested.

Constructor added to LENS_SYMBOL_KINDS.

will-lamerton:
Looking good to go @addyCooks! Can you fix the conflicts then I think we're good to merge :)

addyCooks:
Thanks @will-lamerton! Conflicts resolved, all 11 checks green.

The only interesting one was `chat-panel.js`: this branch moved the send into
`dispatchPrompt()`, while #842 added a `closeMention()` before the input clears.
Kept both and dropped the now-duplicate inline `postMessage` otherwise every
composer message would send twice. `package.json` and `webview-protocol.ts` were
straightforward unions.

Heads up on one unrelated commit: `fix(test): boot mention-utils.js in the chat
panel harness`. `main` is currently red on its own 28 chat-panel specs throw
`Cannot destructure property 'findMentionQuery' of 'globalThis.NanocoderMentionUtils'`.
The harness (#847) never loaded `mention-utils.js` (#842) into its VM sandbox, so
every spec that boots a panel dies before asserting. Neither PR's CI saw the
combination. 11-line fix.

Happy to split that out if you'd rather keep this PR to code lenses though it
stays red until that lands somewhere.

PR #869, opened 2026-08-12
fix: make the stop button end the turn it was pressed

### LINKED ISSUE #864: [Bug] Cancel button does not always stop ongoing execution
## Description

The **Cancel/Stop** button in the chat input does not always cancel the current execution.

Sometimes, after clicking the red stop button, Nanocoder continues executing the current request, tool calls, or pending operations instead of stopping completely.

## Steps to Reproduce

1. Start a request that triggers a longer agent execution or multiple tool calls.
2. While the request is still running, click the red **Stop/Cancel** button.
3. Observe the execution after cancellation.

## Expected Behavior

Clicking the **Cancel/Stop** button should immediately stop the entire current execution, including:

* Ongoing model generation
* Pending tool calls
* Subsequent tool calls queued by the agent
* Any remaining execution loop for the current request

The UI should return to an idle state so the user can submit a new request.

## Actual Behavior

The cancel button sometimes appears to be clicked successfully, but the current execution continues. Tool calls or other operations may continue running after cancellation.

In some cases, the UI also shows a warning such as:

> "Nanocoder: Please approve or deny the pending tool before sending a..."

This can leave the user unable to properly stop the current request or immediately start a new one.

## Logs/Screenshots

Image

## Additional Context

Cancellation should be treated as a hard stop for the current user request. Once the user clicks **Cancel**, no new model generation or tool execution should continue for that request.

addyCooks:
Closes #864.

## Description

Cancel failed in two independent places, so the fix has two halves plus a small UI follow-on:

- **`source/acp/acp-session.ts` / `acp-agent.ts`** — `cancel()` now only aborts; the controller is rotated by a new `beginTurn()` called synchronously at the top of `prompt()`, before it awaits `acpContentToUserMessage`. The loop captures `session.abortController` once at turn start, so previously any cancel landing during that file-reading window was silently dropped.
- **`plugins/vscode/src/acp-client.ts`** — new `cancelPendingPermissions()` resolves every outstanding `session/request_permission` with `{outcome: 'cancelled'}` and clears the map; called from `cancel()` and from `setConnection()` so a crashed-and-restarted agent process doesn't leave the client permanently blocked. This also ends the turn agent-side on its own, since `requestToolPermission` returns `cancelled`.
- **`plugins/vscode/media/chat-panel.js`** — the agent reports a cancelled permission as ACP status `failed` with `rawOutput: 'Cancelled by user'`, which `updateEditCard` did not treat as terminal, so the card kept its spinner and live Allow/Deny buttons. `failed` is now terminal and maps to the cancelled icon when the output says so, matching what `ToolAggregator` already did.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Changeset

### Automated Tests


will-lamerton:
Thanks for this @addyCooks - the agent-side fix is spot on. Moving controller rotation into `beginTurn()` is the right shape, and I confirmed the two new `acp-agent.spec.ts` tests genuinely fail against the old `cancel()` behaviour.

One thing to sort before merge: #789 landed on `main` a few hours ago and already added `_clearPendingPermissions()` (plus an `onPermissionsCancelled` callback that strips Allow/Deny off the cards). Could you rebase on `main`? `acp-client.spec.ts` conflicts outright, and `acp-client.ts` auto-merges cleanly into a broken state - `cancel()` ends up calling your `cancelPendingPermissions()` first, which empties the map, so main's `_clearPendingPermissions()` hits its `size === 0` early return and the webview never gets the `permissionsCancelled` message. No conflict marker to catch it.

Suggested shape after the rebase: drop `cancelPendingPermissions()` and use main's `_clearPendingPermissions()`, but keep your two genuinely new bits on top of it:

- Call it from `setConnection()` - main doesn't cover the crashed-and-restarted-agent case, so that's a real gap you found.
- Your ordering in `cancel()` is better than main's. Clearing above the `if (!this.connection || !this._sessionId) return;` guard is correct; today a cancel with no live connection leaves every resolver hanging.

The `updateEditCard` change is still worth keeping (main only fixed `ToolAggregator`), just align the predicate to main's `/aborterror|cancelled/i` so the two copies don't drift again. Minor: `'Denied by user'` also arrives as `failed`, so denied edits get the error icon - adding `denied` to that regex would fix it in both places.

Heads up on something pre-existing, not your problem to fix here: `plugins/vscode/src/*.spec.ts` isn't picked up by anything. Root ava globs `source/**/*.spec.ts` only, there's no test script in the extension package, and root tsconfig excludes it. So the two new `acp-client` tests won't actually run in CI. I'll open a separate issue for that.

addyCooks:
Thanks @will-lamerton! 
I've rebased on main and addressed all your feedback. 

Here's what changed:
### 1. Dropped `cancelPendingPermissions()`
### 2. Called from `setConnection()`
### 3. Fixed ordering in `cancel()`
### 4. Updated `updateEditCard` regex
### 5. Tests preserved

Ready for re-review when you have a chance!

PR #886, opened 2026-08-17
fix: use os.homedir() for models cache path on Windows

tusharui:
## Problem

On Windows, `process.env.HOME` is undefined. The models cache fallback path uses `process.env.HOME || '~'`, which resolves to the literal string `~/.cache` — not the user's actual home directory. This causes:

- Cache writes to fail silently (caught and logged as a warning)
- Models re-fetched from the network on every launch
- A stray `~` folder created in the current working directory

## Fix

Replace `process.env.HOME || '~'` with `os.homedir()` in `source/models/models-cache.ts`. `os.homedir()` returns the correct home directory on all platforms:

- Windows: `C:\Users\username`
- macOS: `/Users/username`
- Linux: `/home/username`

## Testing

All 9 existing tests in `models-cache.spec.ts` pass.

### DIFF
--- a/.changeset/fix-windows-models-cache-path.md
+++ b/.changeset/fix-windows-models-cache-path.md
@@ -1,0 +1,5 @@
+---
+"@nanocollective/nanocoder": patch
+---
+
+Fixed models cache path on Windows by using `os.homedir()` instead of `process.env.HOME`, which is undefined on Windows. The cache now writes to the correct location instead of creating a literal `~` folder.
--- a/scripts/fetch-models.js
+++ b/scripts/fetch-models.js
@@ -6,6 +6,7 @@
  */
 
 import * as fs from 'node:fs';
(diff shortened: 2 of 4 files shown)


akramcodez:
@tusharui Thanks for the fix! One small process request for future contributions: please create an issue first for any new bugs/features, then ask to be assigned before starting the implementation. This helps us keep ownership and scope clear.

For the additional issues you found here, please open separate issues for them first, and then feel free to pick them up once assigned:

* Duplicate cache-path logic in `scripts/fetch-models.js`
* Dead macOS branch in `DEFAULT_CACHE_DIR`

For this PR, please add the Windows-specific regression test mentioned above and rerun the tests. Everything else looks good!

akramcodez (approved):
Crazy work @tusharui, LGTM

PR #899, opened 2026-08-18
fix(messages): keep UI-only notices out of the model payload

### LINKED ISSUE #893: [Bug] OpenAI Shim Injects Synthetic Protocol Text
### Description
The OpenAI shim might inject synthetic `[Tool results received]` markers or similar protocol text that reaches the model as `assistant` content, which can corrupt the model's generation context.

### Environment
- **OS**: Any
- **Node version**: Node 22+
- **Provider**: OpenAI-compatible API
- **Model**: Any

### Steps to Reproduce
1. Connect via OpenAI-compatible provider.
2. Trigger a tool call and examine the raw payload sent back to the model in subsequent turns.
3. Observe synthetic protocol text.

### Expected Behavior
The shim should format tool results strictly according to the API spec without injecting synthetic markdown/text that the model interprets as its own past output.

### Actual Behavior
Synthetic markers are potentially being injected into the context window.

### Logs/Screenshots
N/A

### Additional Context
- [x] This issue occurs with the latest version of nanocoder and feat

addyCooks:
## Description

Cancellation notices, inline error banners, the non-interactive tool-approval notice, and the
VS Code replies to built-in slash commands (`/help`, `/copy`, `/model`, unrecognized commands)
were pushed into conversation history as `assistant` messages. They sit in the same `Message[]`
that becomes the provider payload, so the next request handed the model harness-authored
markdown as its own past output and because sessions persist `Message[]` verbatim, a resumed
session replayed it into context too. The built-in command replies never involved the model at
all, so both halves of that exchange are marked, not just the reply.

Mark those notices `displayOnly` and filter them in `convertToModelMessages`, the single point
where history becomes the provider payload. They still render in chat and still replay with
session history.

**Deliberately left alone:**

- `truncateToolResult`'s elision marker (`... [Output truncated: N characters total …] ...`) harness-authored, but it lives inside a **tool result**, the correct channel for telling the
  model its output was elided. Removing it would make truncation silent.
- Harness `user` messages (auto-nudges, auto-diagnostics, `[Active file: …]`) and synthetic
  `tool` results (cancellation placeholders, unknown-tool errors). These are the legitimate way
  a harness addresses a model, not text passed off as the model's own output.
- ACP's `cleanedContent` and `prepareStep` were audited and are already purely **subtractive** they strip tool-call syntax and empty messages, never add text.

Closes #893.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Changeset

### Automated Tests

### Manual Testing

- [x] Tested with Ollama
- [ ] Tested with OpenRouter

## Checklist

- [x] Self-review completed

**Assignment:** #893 currently has no assignee — worth asking a maintainer to assign it before
merge.

**Logging:** no new logging. The change adds no new failure path; the three notice sites keep
their existing `logger.error` calls, and the converter filter cannot fail.

**Breaking changes:** none. `displayOnly` is an optional field on `Message`; every existing
message and every persisted session omits it and behaves exactly as before.

### DIFF
--- a/.changeset/fix-display-only-notices.md
+++ b/.changeset/fix-display-only-notices.md
@@ -1,0 +1,5 @@
+---
+"@nanocollective/nanocoder": patch
+---
+
+Stopped nanocoder's own UI text from being sent to the model as its past output. Cancellation notices (`_Cancelled by user._`), inline error banners (`**Error:** ...`), the non-interactive "Tool approval required" notice, and the VS Code replies to built-in slash commands (`/help`, `/copy`, `/model`, unrecognized commands) were all pushed into conversation history as `assistant` messages, so on the next turn the provider received harness-authored markdown as if the model had written it — teaching it to imitate the chrome and, on a resumed session, to believe it had errored. These are now marked display-only: they still render in the chat and replay with session history, but they are filtered out before messages are converted to the provider payload. Closes #893.
--- a/source/acp/acp-agent.spec.ts
+++ b/source/acp/acp-agent.spec.ts
@@ -9,6 +9,7 @@ import {
 	setToolRegistryGetter,
 	setToolManagerGetter,
 } from '@/message-handler';
+import {convertToModelMessages} from '@/ai-sdk-client/converters/message-converter';
 import {sessionManager} from '@/session/session-manager';
 
 console.log('\nacp-agent.spec.ts');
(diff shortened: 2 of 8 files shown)


addyCooks:
@will-lamerton 
Heads up!
main's been red since yesterday, unrelated to any one PR's contents.

chat-panel.js reads globalThis.NanocoderMentionUtils, which media/mention-utils.js defines. The webview loads both, but chat-panel-harness.ts only ran chat-panel.js, so createPanel() throws on the destructure and takes out all 28 chat-panel tests. The harness didn't get updated when the mention helpers were split into their own script in #842.

One-line fix in the harness PR is up: #900. 
Suggest merging that first, since anything branched off current main shows the same 28 failures.

will-lamerton:
Great work @addyCooks :)

addyCooks:
Thanks @will-lamerton !
appreciate it : )

PR #901, opened 2026-08-18
Bug: Perf/message token cache

### LINKED ISSUE #894: [Bug] Performance: In-place BoundedMap Update Causes High Allocation
### Description
`useAppState.tsx:265-275` allocates a new `BoundedMap` on every miss by making a full copy of existing entries, which is inefficient and triggers unnecessary garbage collection.

### Environment
- **OS**: Any
- **Node version**: Node 22+
- **Provider**: Any
- **Model**: Any

### Steps to Reproduce
1. Run Nanocoder in interactive mode.
2. Perform operations that frequently miss the BoundedMap cache.
3. Profile memory allocations.

### Expected Behavior
The map should update in-place, perhaps by tracking the oldest entry via an insertion counter.

### Actual Behavior
A new `BoundedMap` is allocated on every miss.

### Logs/Screenshots
N/A

### Additional Context
- [x] This issue occurs with the latest version of nanocoder and feat

addyCooks:
## Description

Closes #894.

`getMessageTokens` rebuilt the entire cache on every miss: it allocated a fresh `BoundedMap`, copied up to 1000 entries into it (each copy also allocating a new entry wrapper), and stored the result back through `setMessageTokenCache`. Every uncached message cost a full copy of the cache and turned all of its previous entries into garbage.

Because that copy landed in state, it also re-rendered the whole app and handed `getMessageTokens` a new identity, invalidating the memos and effects keyed on it (context percentage, usage display). And since the write was deferred to a microtask, repeated calls for the same message inside a single render pass all missed `usage-display` walks every message twice per render each queueing another full copy.

This holds the cache in a ref and writes the entry straight into it. The map instance stays stable for the session, a miss costs one `Map.set`, the value is visible to the very next lookup, and a cache write no longer reaches React which is what makes the `queueMicrotask` deferral (added to avoid updating state during render) unnecessary.

Eviction is deliberately unchanged: `BoundedMap` already drops the oldest entry in O(1) using Map insertion order, so the insertion counter suggested in the issue would add bookkeeping without a win.

`setMessageTokenCache` is dropped from the returned setters nothing consumed it. `messageTokenCache` stays exposed.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Changeset

### Automated Tests

- [x] All existing tests pass (`pnpm test:all` completes successfully) — see note below

### Manual Testing

- [ ] Tested with Ollama
- [ ] Tested with OpenRouter

## Checklist

- [x] Self-review completed

### DIFF
--- a/.changeset/perf-message-token-cache.md
+++ b/.changeset/perf-message-token-cache.md
@@ -1,0 +1,5 @@
+---
+"@nanocollective/nanocoder": patch
+---
+
+Update the message token cache in place instead of copying every entry into a new map on each cache miss.
--- a/source/hooks/useAppState.spec.tsx
+++ b/source/hooks/useAppState.spec.tsx
@@ -22,14 +22,21 @@ console.log('\nuseAppState.spec.tsx');
 type AppStateHook = ReturnType<typeof useAppState>;
 
 let captured: AppStateHook | null = null;
+let renderCount = 0;
 
 function Probe({initialMode}: {initialMode?: DevelopmentMode}) {
+	renderCount++;
 	captured = useAppState(initialMode ?? 'normal');
 	return null;
 }
 
+function tokenCacheKey(message: Message, model = '') {
+	return (message.content || '') + message.role + model;
+}
+
 function setup(initialMode: DevelopmentMode = 'normal') {
 	captured = null;
+	renderCount = 0;
 	const instance = render(<Probe initialMode={initialMode} />);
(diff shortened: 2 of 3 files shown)


will-lamerton:
Nice work @addyCooks :)

addyCooks:
Thanks @will-lamerton : )

PR #903, opened 2026-08-19
feat(caching): Anthropic prompt caching with cache-aware cost reporting

### LINKED ISSUE #888: [Feature] Prompt / Context Caching
### Description
Implement prompt/context caching using provider-specific options (e.g., Anthropic's `cacheControl` breakpoints or OpenAI's automatic prefix caching). Since the system prompt, tool schemas, and conversation history are resent every turn, caching these prefixes will drastically cut latency and cost.

### Use Case
For multi-turn sessions with large system prompts and tool definitions, users currently pay token costs and latency penalties on every turn for static context.

### Proposed Solution
Modify `source/ai-sdk-client/chat/provider-options.ts` to set cache-control hints. Mark stable prefixes (system prompt + tool definitions) as cacheable.

### Alternatives Considered
None. This is a native feature of newer model APIs that just needs to be wired up.

### Additional Context

### Implementation Notes
This is the single highest-leverage change identified for reducing multi-turn latency and cost.

addyCooks:
## Description

Marks the stable prompt prefix with Anthropic cache breakpoints so multi-turn sessions
read it back out of cache instead of resending it at full price every turn.

`withCacheBreakpoints()` folds the system prompt into the message array and places two
breakpoints: the system entry, and the final message of the turn. The system breakpoint
also covers tool schemas, since tools precede system in Anthropic's prefix order. The
`system:` param is dropped when caching is active a bare string can't carry
`providerOptions`, and sending both would duplicate the prompt. Below Anthropic's
minimum cacheable length the prompt is still folded in but no breakpoints are emitted.

Anthropic-only: OpenAI and OpenRouter prefix-cache automatically, and local models have
no cache to address. `"promptCaching": false` on the provider config opts out.

Cost reporting is made cache-aware in the same PR without it, every cache hit would be
billed at the full input rate and overstate spend. `cacheReadTokens`/`cacheWriteTokens`
flow from `usage.inputTokenDetails` through `ApiUsage` → `ApiCallRecord` → `ResponseUsage`,
and `priceTokens()` becomes the single pricing function shared by the per-response
indicator and `/usage`, billing reads and writes at their own models.dev rates.

Closes #888.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

## Changeset

### Automated Tests

### Manual Testing

- [ ] Tested with Ollama
- [ ] Tested with OpenRouter

## Checklist

- [x] Self-review completed


will-lamerton:
Amazing work @addyCooks. Brilliant.

addyCooks:
Thanks a lot @will-lamerton ! 
really appreciate the kind words 🙌 
glad I could help!

PR #908, opened 2026-08-20
Feat/smart auto routing

### LINKED ISSUE #891: [Feature] Smart Auto-Routing (Simple vs Strong Models)
### Description
Implement a cascading model strategy that auto-routes trivial or tool-only turns (simple reads, formatting checks) to a small/fast/cheap model, escalating to the user's "main" strong model only when task complexity warrants it.

### Use Case
Saves cost and reduces latency on easy tasks (like `view_file`) without sacrificing quality on hard reasoning tasks.

### Proposed Solution
Create `source/ai-sdk-client/smart-router.ts` with a classifier based on prompt length, code-block count, etc. Add a `/smartroute` command to toggle this behavior.

### Alternatives Considered
Forcing users to manually swap models (which creates friction).

### Additional Context

atulya-srivastava:
Implements a **Smart Auto-Routing** cascading model strategy that dynamically routes user prompts between a lightweight model (for simple/trivial turns) and the primary "strong" model (for complex reasoning & code edits). 

Closes #891.

### Overview

When enabled:
- **Classifier Engine**: Evaluates prompt complexity using heuristic keyword matching, code block counts, and character length thresholds.
- **Slash Command**: Provides `/smartroute [on|off|simple <model>|threshold <low|medium|high>]` to toggle and configure auto-routing at runtime.
- **Per-Turn Model Swapping**: Temporarily switches the LLM client to the simple model for trivial turns, with a `try...finally` block guaranteeing model restoration for subsequent turns.
- **State Persistence**: Persists settings to `nanocoder-preferences.json` under `"smartRouting"`, ensuring choices survive terminal restarts.

### Notes for Reviewers

- **Fail-Safe Default**: Prompts with no explicit trivial keywords default to the primary strong model to prevent sending complex/ambiguous tasks to underpowered models.
- **Model Restoration Guarantee**: Model swapping is wrapped in a `try...finally` block inside `useChatHandler.tsx` so that aborts, errors, or cancellations always restore the user's primary model cleanly.

### Files

- `source/ai-sdk-client/smart-router.ts` — Core complexity classifier engine (`classifyTurnComplexity` & `autoSelectSimpleModel`)
- `source/ai-sdk-client/smart-router.spec.ts` — 16 unit tests for complexity classification rules, custom options, and model detection
- `source/commands/lazy-registry.ts` — Registered `/smartroute` in slash command lazy loader
- `source/app/utils/handlers/smartroute-handler.ts` — Slash command handler for `/smartroute` subcommands
- `source/app/utils/app-util.ts` — Special command dispatch integration
- `source/hooks/useAppState.tsx` — Managed `smartRouting` state with persistence handlers
- `source/hooks/chat-handler/useChatHandler.tsx` — Injected per-turn model swapping logic into the chat loop
- `source/hooks/useAppHandlers.tsx` & `source/types/app.ts` — Prop-drilled `smartRouting` state to message submission handlers
- `source/types/config.ts` & `source/config/preferences.ts` — Added `SmartRoutingState` interface & persistent storage methods for `nanocoder-preferences.json`
- `.changeset/smart-auto-routing.md` — Added changeset for release automation

---

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

## Changeset

### Automated Tests

- [x] All existing tests pass (`pnpm run test:types` and `pnpm run build` complete with 0 errors)

16 new tests in `source/ai-sdk-client/smart-router.spec.ts` covering trivial prompts, code block detection, complex keywords, sensitivity thresholds, word boundary matching, custom options, and lightweight model auto-selection.

`pnpm run test:types` passes with 0 errors.

### Manual Testing

- [ ] Tested with Ollama
- [ ] Tested with OpenRouter

Manually verified via CLI non-interactive and interactive TUI sessions:
- Verified `/smartroute on`, `/smartroute simple <model>`, and `/smartroute threshold <level>` subcommands.
- Verified persistence of `"smartRouting"` state across CLI process restarts in `nanocoder-preferences.json`.
- Verified model selection stream output during chat execution.

## Checklist

- [x] Self-review completed

### DIFF
--- a/.changeset/smart-auto-routing.md
+++ b/.changeset/smart-auto-routing.md
@@ -1,0 +1,5 @@
+---
+"@nanocollective/nanocoder": minor
+---
+
+Implement smart auto-routing cascading model strategy between lightweight and strong models (`/smartroute`).
--- a/source/ai-sdk-client/smart-router.spec.ts
+++ b/source/ai-sdk-client/smart-router.spec.ts
@@ -1,0 +1,130 @@
+import test from 'ava';
+import {
+	autoSelectSimpleModel,
+	classifyTurnComplexity,
+} from './smart-router.js';
+
+// ── classifyTurnComplexity ──────────────────────────────────────────
+
+test('returns simple for short, trivial prompts', t => {
+	t.is(classifyTurnComplexity('view file package.json'), 'simple');
+	t.is(classifyTurnComplexity('show status'), 'simple');
+	t.is(classifyTurnComplexity('what is line 5?'), 'simple');
+});
+
+test('returns simple for empty or whitespace-only input', t => {
+	t.is(classifyTurnComplexity(''), 'simple');
+	t.is(classifyTurnComplexity('   '), 'simple');
+});
+
+test('returns strong for a single code block', t => {
+	const prompt = 'Fix this:\n```js\nconsole.log("hello");\n```';
+	t.is(classifyTurnComplexity(prompt), 'strong');
+});
+
+test('returns strong for multiple code blocks', t => {
+	const prompt = [
+		'Compare these two:',
+		'```js',
+		'const a = 1;',
+		'```',
+		'and',
+		'```js',
+		'const b = 2;',
+		'```',
+	].join('\n');
+	t.is(classifyTurnComplexity(prompt), 'strong');
+});
+
+test('returns strong for complex keywords', t => {
+	t.is(
+		classifyTurnComplexity('Refactor authentication flow in login.tsx'),
+		'strong',
+	);
+	t.is(
+		classifyTurnComplexity('Architect a new state management system'),
+		'strong',
+	);
+});
+
+test('returns strong when no keyword matches (fail-safe default)', t => {
+	// "hello world" has no trivial or complex keywords — should default to strong
+	t.is(classifyTurnComplexity('hello world'), 'strong');
+});
+
+test('supports custom complex keywords', t => {
+	// "deploy" is NOT in the default complex list
+	t.is(classifyTurnComplexity('deploy to staging'), 'strong'); // no keyword → default strong
+	t.is(
+		classifyTurnComplexity('deploy to staging', {
+			customComplexKeywords: ['deploy'],
+		}),
+		'strong',
+	);
+});
+
+test('supports custom trivial keywords', t => {
+	// "inspect" is NOT in the default trivial list, so without custom it defaults to strong
+	t.is(classifyTurnComplexity('inspect log file'), 'strong');
+	t.is(
+		classifyTurnComplexity('inspect log file', {
+			customTrivialKeywords: ['inspect'],
+		}),
+		'simple',
+	);
+});
+
+test('respects word boundaries — does not match substrings', t => {
+	// "cat" is a trivial keyword, but "categories" should NOT trigger it.
+	// "categories" alone has no keyword match → defaults to strong.
+	t.is(classifyTurnComplexity('categories'), 'strong');
+	// But "cat the file" should match the trivial keyword "cat".
+	t.is(classifyTurnComplexity('cat the file'), 'simple');
+});
+
+test('respects threshold settings', t => {
+	const text = 'a'.repeat(150); // 150 chars, no keywords
+	t.is(classifyTurnComplexity(text, {threshold: 'high'}), 'strong'); // within 300 limit but no keyword → strong
+	t.is(classifyTurnComplexity(text, {threshold: 'low'}), 'strong'); // exceeds 100 limit → strong
+});
+
+test('long prompts are always strong regardless of keywords', t => {
+	const longTrivial = `view ${'a'.repeat(250)}`;
+	t.is(classifyTurnComplexity(longTrivial), 'strong');
+});
+
+// ── autoSelectSimpleModel ───────────────────────────────────────────
+
+test('autoSelectSimpleModel finds lightweight models', t => {
+	const models = [
+		'gpt-4o',
+		'gpt-4o-mini',
+		'claude-3-5-sonnet-20241022',
+		'claude-3-5-haiku-20241022',
+	];
+	t.is(autoSelectSimpleModel(models), 'gpt-4o-mini');
+});
+
+test('autoSelectSimpleModel finds local lightweight models', t => {
+	const ollamaModels = ['llama3.3:70b', 'qwen2.5-coder:7b'];
+	t.is(autoSelectSimpleModel(ollamaModels), 'qwen2.5-coder:7b');
+});
+
+test('autoSelectSimpleModel falls back to first model when no pattern matches', t => {
+	const models = ['custom-model-a', 'custom-model-b'];
+	t.is(autoSelectSimpleModel(models), 'custom-model-a');
+});
+
+test('autoSelectSimpleModel returns undefined for empty list', t => {
+	t.is(autoSelectSimpleModel([]), undefined);
+});
+
+test('autoSelectSimpleModel supports custom patterns', t => {
+	const models = ['model-v1-custom-fast', 'model-v1-standard'];
+	t.is(
+		autoSelectSimpleModel(models, {
+			customLightweightPatterns: [/fast/i],
+		}),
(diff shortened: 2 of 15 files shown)


akramcodez (commented):
Hey @atulya-srivastava, thanks for working on this! I went through the PR and the classifier itself looks really good - the fail-safe behavior, routing rules, and unit tests are all solid.

However, I don't think it's ready to merge yet. There are two important things I'd like you to address:

1. **Integration tests:** The current 16 tests only cover the pure classifier. Please add coverage for the actual `/smartroute` handler, preference persistence, and the per-turn model swap/restore behavior. We need to make sure the feature actually works end-to-end, not just that the classifier works.

2. **Model state mutation:** The current implementation temporarily mutates the shared client's model inside `useChatHandler`. I'd like us to discuss this approach before merging, particularly around concurrent submissions and ensuring the original model is always restored safely. If this approach is necessary, please document the invariants; otherwise, consider a safer way to pass the model to the LLM call.

Also, one small process note for future contributions: **please ask for assignment on the issue before starting/opening a PR**. It helps us coordinate contributors and avoid multiple people working on the same issue.

Once the integration coverage is added and we've discussed the model-state approach, we can take another look at this for review. The overall direction is good!

atulya-srivastava:
Hey @akramcodez ,

Thanks for the detailed feedback! glad to hear the classifier and routing logic look good.

I took another look at how `useChatHandler`, `AISDKClient`, and `useAppState` interact to better understand the concern about changing the model on the shared client.

Right now, turns from the interactive CLI are submitted one at a time while streaming. That said, I agree this is a UI level behavior rather than an architectural guarantee at the client layer.

Because of that, passing the model into each call is the safer approach. I propose adding an optional `model?: string` to `ModeOverrides` (passed into `LLMClient.chat()`) and having `AISDKClient.chat()` use it when provided, while keeping existing callers unchanged.

I will also add integration tests covering:

• `/smartroute` behavior and preference persistence
• Choosing a model per turn for simple and complex prompts
• Model restoration and error handling
• Routing being disabled

This avoids relying on shared mutable state and keeps the change fairly contained. I am happy to update the PR this way. What's you view regarding this approach?

Also noted on asking for assignment before starting an issue. I will make sure to do that for future contributions.

Thanks again for the review!

atulya-srivastava:
Hey @akramcodez,  just following up on this. Would you be okay with me proceeding with the per call `model` approach I proposed above? I am happy to make the changes once you confirm.

will-lamerton:
Nice work @atulya-srivastava, clean plumbing here - the stub command plus app-layer handler split, the preferences write-through, and the `try/finally` model restore all follow existing conventions well. My concerns are with the two pieces that decide whether the feature helps or hurts: the classifier's routing decisions, and the fact that the model swap doesn't propagate to the rest of the turn's state.

I checked out the branch and ran the classifier directly. All 20 assertions in the new spec pass, but on realistic prompts the routing goes the wrong way:

- `"ok now fix the auth race condition in the session store"` -> simple
- `"check why the login flow drops sessions and patch it"` -> simple
- `"find the bug that corrupts the checkpoint file and fix it"` -> simple
- `"no, that broke the build - please make it work again"` -> simple
- `"what is causing the deadlock in the daemon event router?"` -> simple
- `"list all the places we mutate global state and clean them up"` -> simple
- `"format the whole codebase and fix the lint errors"` -> simple
- `"yes"` -> simple

Ten of twelve realistic coding requests I tried routed to the weak model. Also `autoSelectSimpleModel(['text-embedding-3-large', 'gpt-4o'])` returns `text-embedding-3-large`, and `['whisper-1', 'gpt-4o']` returns `whisper-1`.

### Blocking

- **The fail-safe default is inverted in practice.** Rule 6 only falls back to strong when *no* keyword matches, but `DEFAULT_TRIVIAL_KEYWORDS` includes `ok`, `no`, `yes`, `check`, `find`, `where`, `what is`, `list`, `format`, `help`, and `read`. Those appear in nearly every real request, and a single trivial match anywhere in a sub-200-char prompt wins. The bias needs to be the other way: trivial only when the *whole* message is trivial (short, no imperative verb, no file path, no tool-triggering intent), not when it merely contains a trivial token.
- **Confirmations carry the most risk and get routed cheapest.** `"yes"` and `"ok, do it"` classify as simple, but that turn runs the full tool loop against the entire conversation history, including file edits. Classification looks only at the current message with no conversation context.
- **`autoSelectSimpleModel` falls back to `availableModels[0]`.** On an OpenAI-compatible endpoint that lists embedding or audio models, `/smartroute on` silently binds the simple model to something that cannot chat. The fallback should be `undefined` (routing stays off) rather than an arbitrary pick.
- **The swap bypasses all model-derived state.** `client.setModel()` mutates the client, but `currentModel` in React state is untouched, so for that turn:
  - `getBaseSystemPrompt(..., currentModel)` builds the prompt for the strong model
  - `toolManager.getAvailableToolNames(..., currentModel)` filters tools for the strong model
  - the `toolsDisabled` memo (`source/hooks/chat-handler/useChatHandler.tsx:112`) picks native vs XML tool calling from the strong model, so a simple model listed in the provider's `disableToolModels` still gets native tools
  - `buildResponseUsageBounded(result.usage, currentModel)` (`source/hooks/chat-handler/conversation/conversation-loop.tsx:539`) prices the response at the strong model's rate, so a cost-saving feature misreports cost
  - `setModel` fires `void this.updateContextSize()` unawaited, so the context window is momentarily the previous model's

  Routing through the existing model-switch path instead of mutating the client behind React's back would fix all of these at once.
- **No capability check on the simple model.** Nothing verifies it supports tool calling or has a usable context window before a tool-heavy turn is handed to it.

### Should fix

- Two of the sixteen tests assert nothing. `supports custom complex keywords` expects `'strong'` on both branches, which is also the no-match default, so the custom path is never exercised. `respects threshold settings` expects `'strong'` for both `high` and `low`, so the threshold is never exercised. Both need a case where the option flips the result.
- No tests for the two integration points that carry the risk: the swap and restore in `useChatHandler`, and the `/smartroute` subcommands.
- Docs are unchanged despite the checkbox. `/smartroute` should be added to the command list in `docs/features/commands.md`.
- The swap is invisible to the user. Nothing indicates which model answered, and the status line keeps showing the primary model. For a feature that silently downgrades the model, some per-turn indication seems important. The description mentions verifying "model selection stream output", but I don't see that output in the diff.
- Scope is TUI-only: `source/plain/`, ACP, subagents, and the daemon path are unaffected. That's a reasonable scope, but worth noting in the description since it lists non-interactive CLI testing.
- When auto-select picks a model equal to the current one, `simpleModel` stays `undefined` and the auto path re-runs every turn. Cheap in practice since `getAvailableModels()` resolves from a cached array, but it adds an `await` before the turn's `try` block on every message.

### Nits

- `useAppHandlers.tsx`: `tune: TuneConfig | undefined` was loosened to `tune?: TuneConfig`, an unrelated change that weakens a required prop.
- `app-util.ts`: the `options as Parameters<typeof handleSmartRouteCommand>[1]` cast is unnecessary now that `MessageSubmissionOptions` carries `smartRouting` and `setSmartRouting`.
- `smartroute-handler.ts`: the `SMART_ROUTING_DEFAULTS` import sits mid-file after the doc comment rather than with the other imports.
- Several inline `import('@/types/config').SmartRoutingState` references that could be a top-level `import type`.
- The restore path compares against the `currentModel` prop but captures and restores `client.getCurrentModel()`. Same value today, but two sources of truth for one decision.

Happy to look again once the classifier bias and the state propagation are addressed.

atulya-srivastava:
hi @will-lamerton, thanks for the detailed review. I am working through the feedback now, especially the classifier fail safe behavior, model state propagation, and capability checks. I will also add the missing integration tests and docs updates. I will push the fixes once these are addressed.

akramcodez:
Hey @atulya-srivastava, your PR checks are failing and there are several merge conflicts. Please fix the conflicts and get all checks passing, then tag me and I’ll review it.

akramcodez:
Hey @atulya-srivastava, your PR checks are failing and there are several merge conflicts. Please fix the conflicts and get all checks passing, then tag me and I’ll review it.

PR #917, opened 2026-08-21
feat: add expandable long user prompts

### LINKED ISSUE #865: [Feature] Collapse long user prompts in the conversation view
## Description

When a user sends a very long prompt, the entire prompt is currently displayed in the conversation, which can take up a large amount of vertical space and make the chat difficult to navigate.

For long prompts, we should automatically collapse the content after a reasonable limit, for example **40 words**, while keeping the full prompt accessible through an expand/collapse control.

## Expected Behavior

For prompts under the limit, display the full content normally.

For prompts exceeding the limit, show the first ~40 words followed by a **"Show more"** or expandable section.

For example:

> Add a `verify` mode that allows Nanocoder to act as an autonomous participant in the CI/CD lifecycle, automatically investigating CI failures, performing read-only PR reviews, and optionally proposing...

**Show more ↓**

Clicking the control should expand the complete original prompt, and clicking again should collapse it.

## Actual Behavior

The entire user prompt is displayed without any collapsing or truncation, even when it contains hundreds or thousands of words. This makes the conversation unnecessarily long and pushes the agent's response far down the page.

## Goal

Keep the conversation compact and readable while ensuring that users can still access the complete prompt whenever they need it.

The full prompt should remain available and should never be truncated or lost.

DeepamJha:
Closes #865

## Description

Adds automatic collapsing for long user prompts in the conversation view.

Prompts exceeding 40 words or 300 characters are collapsed for display while the original prompt remains intact.

The implementation:

- Collapses long prompts while keeping the full message available.
- Supports `Show more` / `Show less` interaction through keyboard focus.
- Keeps interactive user messages live in the inline transcript so the toggle can receive focus.
- Strips VS Code context only from the displayed content.
- Preserves the original message and token calculation.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

## Changeset

- [x] Added a changeset (`pnpm changeset`) describing this feature for the changelog.

### Automated Tests

- [x] Added and updated `user-message.spec.tsx`
- [x] Tested 40-word collapse boundary
- [x] Tested 300-character collapse boundary
- [x] Tested long unbroken strings
- [x] Tested VS Code context stripping
- [x] Tested interactive expand/collapse behavior
- [x] `pnpm test:format`
- [x] `pnpm test:lint`
- [x] `git diff --check`

### Manual Testing

Tested the feature in the real CLI:

- Long prompts collapse automatically.
- `Tab` focuses `Show more`.
- `Enter` expands the prompt.
- `Enter` again collapses it.
- Normal input behavior remains functional.

## Checklist

- [x] Self-review completed
- [x] No breaking changes

### DIFF
--- a/.changeset/quiet-cobras-itch.md
+++ b/.changeset/quiet-cobras-itch.md
@@ -1,0 +1,5 @@
+---
+"@nanocollective/nanocoder": minor
+---
+
+Add automatic collapsing for long user messages in the conversation view. Messages exceeding 40 words or 300 characters are collapsed for display, while the original prompt remains intact and can be recalled from input history.
--- a/source/components/user-message.spec.tsx
+++ b/source/components/user-message.spec.tsx
@@ -368,3 +368,111 @@ code2
 	t.false(output!.includes('code1'));
 	t.false(output!.includes('code2'));
 });
+// ============================================================================
+// Long message collapse tests
+// ============================================================================
+
+test('UserMessage does not collapse a message at 40 words', t => {
+	const message = Array.from({length: 40}, (_, index) => `word${index + 1}`).join(
+		' ',
+	);
+	const {lastFrame} = render(
+		<MockThemeProvider>
+			<UserMessage message={message} />
+		</MockThemeProvider>,
+	);
+	const output = lastFrame() ?? '';
+
+	t.true(output.includes('word40'));
+	t.false(output.includes('Full prompt: ↑ history'));
+	t.false(output.includes('...'));
+});
+
+test('UserMessage collapses a message over 40 words and points to history', t => {
+	const message = Array.from({length: 41}, (_, index) => `word${index + 1}`).join(
+		' ',
+	);
+	const {lastFrame} = render(
+		<MockThemeProvider>
+			<UserMessage message={message} />
+		</MockThemeProvider>,
+	);
+	const output = lastFrame() ?? '';
+
+	t.true(output.includes('word40'));
+	t.false(output.includes('word41'));
+	t.true(output.includes('Full prompt: ↑ history'));
+});
+
+test('UserMessage does not collapse a message at 300 characters', t => {
+	const message = `hello ${'a'.repeat(294)}`;
+	t.is(message.length, 300);
+	const {lastFrame} = render(
+		<MockThemeProvider>
+			<UserMessage message={message} />
+		</MockThemeProvider>,
+	);
+
+	t.false((lastFrame() ?? '').includes('Full prompt: ↑ history'));
+});
+
+test('UserMessage collapses a message over 300 characters', t => {
+	const message = `hello ${'a'.repeat(301)}`;
+	const {lastFrame} = render(
+		<MockThemeProvider>
+			<UserMessage message={message} />
+		</MockThemeProvider>,
+	);
+	const output = lastFrame() ?? '';
+
+	t.true(output.includes('...'));
+	t.true(output.includes('Full prompt: ↑ history'));
+	t.false(output.includes(message));
+});
+
+test('UserMessage enforces the character limit for long words', t => {
+	const message = Array.from(
+		{length: 41},
+		(_, index) => `word${String(index + 1).padStart(2, '0')}-${'x'.repeat(12)}`,
+	).join(' ');
+	const {lastFrame} = render(
+		<MockThemeProvider>
+			<UserMessage message={message} />
+		</MockThemeProvider>,
+	);
+	const output = lastFrame() ?? '';
+
+	t.true(output.includes('word01-'));
+	// The first 40 words exceed 300 characters, so the character cap must
+	// truncate before word20 rather than merely hiding word41.
+	t.false(output.includes('word20-'));
+	t.false(output.includes('word41-'));
+	t.true(output.includes('...'));
+});
+
+test('UserMessage collapses long unbroken strings', t => {
+	const message = `hello ${'s'.repeat(500)}`;
+	const {lastFrame} = render(
+		<MockThemeProvider>
+			<UserMessage message={message} />
+		</MockThemeProvider>,
+	);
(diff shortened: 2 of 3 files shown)


will-lamerton:
Thanks for taking this on @DeepamJha , and for the thorough manual testing notes. The feature itself matches what #865 asked for, and the new spec suite passes cleanly (31/31 locally). Two of the changes made to get the toggle working are regressions to the default TUI, though, and I think they point at a design tension worth resolving before this lands.

## Blocker 1: `interactive={true}` disables `<Static>` for the default inline TUI

`chat-history.tsx:142` becomes `disableStatic: fullscreen || interactive`, and `interactive-app.tsx:297` passes `interactive={true}` unconditionally. Alt-screen is opt-in (`cli.tsx:518-523` requires `--alt-screen` or the `alternateScreen` preference), so the default session is inline, and this takes the entire inline transcript off the `<Static>` path.

Three consequences:

- The transcript is silently capped at the last 60 entries. `ChatQueue`'s no-Static branch slices to `FULLSCREEN_TAIL_CAP` (`chat-queue.tsx:56-59`). I verified this: rendering 70 messages through `ChatHistory` with `fullscreen={false} interactive={true}`, messages 0-9 are absent from the frame and 10-69 remain; with `interactive={false}` all 70 render. In fullscreen that cap is harmless because the viewport clips anyway, but inline there is no viewport, so those messages are simply lost.
- Nothing is written to the terminal's native scrollback any more, and inline mode has no PageUp/PageDown fallback (`chat-history.tsx:107` gates scrolling on `fullscreen && scrollActive`). Older history becomes unreachable, not just off-screen.
- Every keystroke now redraws the whole transcript instead of a single input line. The comment at `chat-history.tsx:123-129` names this exact failure: it is the original "banner disappears" bug, because Ink erases and rewrites the entire non-Static region and clips its top once it outgrows the terminal.

To be fair, this isn't a missed detail so much as a real architectural conflict. `<Static>` writes each item once and never redraws it, so a `useState` toggle inside a transcript entry genuinely cannot work there. Turning Static off is just too expensive a way to buy it.

## Blocker 2: every long message joins the focus ring, so Tab and Shift+Tab steal the input

`useFocus` in `user-message.tsx` registers one focusable per long message. Ink handles Tab and Shift+Tab itself at the App level (`ink/build/components/App.js:189-206`), independently of what any userland `useInput` handler returns. I confirmed with a probe rendering a stub `useFocus({id: 'user-input'})` alongside one long `UserMessage`: after a single `\t` or `\x1B[Z`, the input reports `INPUT_FOCUSED=false`.

That breaks two live bindings:

- Shift+Tab is the documented mode toggle (`user-input.tsx:695`). It now toggles the mode and also moves focus to a transcript toggle. Combined with `focus={effectiveFocus}` on TextInput (`user-input.tsx:1004`) plus the new `if (!effectiveFocus) return;`, the user can no longer type.
- Tab drives command and file autocomplete (`user-input.tsx:748-794`). It still completes on that keypress, but focus then leaves the input.

Recovery requires Tab-cycling through every long message in the transcript, and there is no Escape handler in `UserMessage` to hand focus back. In a real session where most prompts clear 40 words, that ring gets long fast.

## Secondary notes

- The new early return at `user-input.tsx:724-728` sits above the Escape, Ctrl+V and Ctrl+X handlers, so those stop working whenever focus is elsewhere. It is largely redundant with the existing `focus={effectiveFocus}` on TextInput, and it makes the stuck-focus state above harder to escape from.
- Doc comments are now inaccurate: `chat-history.tsx:36-51` still describes inline as "the classic Ink Static append-only flow", `chat-queue.tsx:54-55` and `61-66` still describe the no-Static branch as the fullscreen path, and `FULLSCREEN_TAIL_CAP` no longer applies only to fullscreen.
- The spec's VS Code fixtures were rewritten to put `<!--/vscode-context-->` on its own line, but the app only ever emits it inline after the closing fence (`useVSCodePromptHandling.tsx:42-73`). The tests now exercise a shape that never occurs in practice. The regex handles both forms, so this is coverage lost for no gain.
- Word-limit truncation takes priority over the char limit, and it rejoins via `words.slice(0, 40).join(' ')`. So 40 very long words still render unbounded, and newlines plus indentation are flattened in the preview. The char cap probably wants to apply on both branches.
- The interactive test relies on three `setTimeout(25)` sleeps, which is likely to be the first thing to flake in CI.
- Nothing covers the two changes outside `UserMessage`: the new `interactive` prop on `ChatHistory` and the focus gate in `user-input`.
- There is a fair amount of unrelated churn in the spec (blank lines after every `lastFrame()`, deleted explanatory comments, the `image{s}` JSX reflow) which makes the diff harder to read. The `console.log(React.version)` removal and the `"Line 1\nLine 2"` to `{'Line 1\nLine 2'}` fix are genuine improvements worth keeping.

## Suggested direction

Keep `<Static>` and drop the per-message focusable. Two options that satisfy #865 without touching the transcript architecture:

- Render the collapsed preview statically with a non-interactive hint such as `... +180 words`, and handle expansion through a slash command or a key owned by the input, re-emitting the full prompt as a new transcript entry. No focus ring, no Static change.
- If in-place toggling is a hard requirement, scope it to fullscreen only (at which point `interactive` is redundant with `fullscreen`) and give it a dedicated keybinding rather than the Tab ring, so the mode toggle and autocomplete stay intact.

Happy to talk through either approach if it would help.

DeepamJha:
@will-lamerton Thanks for the detailed review. I see the issues with both the Static behavior and the focus ring interfering with the input. I agree that changing the transcript away from Static isn't worth the tradeoff here, so I'll rework it to keep the existing Static flow and avoid making each long message focusable.

I'll also clean up the related test/fixture issues you pointed out and keep the changes scoped to #865.

akramcodez (approved):
LGTM

Issue #890, opened 2026-08-18
[Feature] Codebase Intelligence & Repo Map (`/repomap`)


akramcodez:
### Description
Introduce a local tree-sitter or `@swc/core` based symbol index and directed edge graph (file -> referenced file) to generate a Repo Map. This map can be injected into the system prompt to give the agent codebase-wide context.

### Use Case
Currently, every "find X" requires either a grep or a full LLM tool round-trip. A lightweight symbol index would allow the agent to understand codebase structure without LLM invocations, saving time and tokens.

### Proposed Solution
Create a new `source/repo-map/` module with parsers, graph generation, and in-house PageRank. Add a `/repomap` slash command. Auto-inject a 1024-token preview into the system prompt.

### Alternatives Considered
Relying entirely on LSP (Nanocoder has an LSP client but doesn't proactively use it for background context-gathering).

### Additional Context
- [x] I have searched existing issues to ensure this is not a duplicate
- [x] This feature aligns with the project's goals (local-first AI assistance)

addyCooks:
Hey @akramcodez!
 
Happy to take this on. 
A few implementation notes:

**1. Skip tree-sitter for v1.**
**2. No import resolution.**
**3. Injection can't happen inside `buildSystemPrompt`.**
**4. Startup latency is the real risk,**

Suggest splitting into 3 PRs: 
(1) `source/repo-map/` + `/repomap` command, no injection; 
(2) prompt injection + `repo_map` tool; 
(3) optional tree-sitter extractor.

akramcodez:
sure @addyCooks go for it :)

PR #919, opened 2026-08-22
feat(repo-map): local symbol index and /repomap command

addyCooks:
## Description

Adds `source/repo-map/`, a dependency-free codebase index, and the `/repomap` slash command that renders it.

The scanner walks the project through the existing `walkProjectEntries` helper, so `.gitignore` and the default ignore directories apply for free. Per file it extracts the symbols that file defines and the identifiers it references, links a referencing file to every file defining a referenced name, and ranks the resulting directed graph with an in-house PageRank.

Three rules keep the graph honest without a real parser: only exported symbols count as definitions (a non-exported symbol cannot be referenced across files anyway), property accesses are not references, and for languages with named imports a reference only creates an edge when the file actually imports that name. Comments and string literals are stripped before either pass.

Covers TypeScript/JavaScript, Python, Go, Rust, Java/Kotlin/C#/Swift, Ruby, PHP and C/C++. Output is budgeted to 1024 tokens by default; `/repomap --tokens <n>` widens it. Scanning nanocoder itself takes ~850ms for 842 files.

This is the first of the three steps agreed on the issue: module plus command only. Prompt injection, the `repo_map` tool and an optional tree-sitter extractor stay out of scope here.

Refs #890

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Documentation update one row added to `docs/features/commands.md`

## Changeset

- [x] Added a changeset (`.changeset/repo-map-command.md`, `minor`)

### Automated Tests

- [x] New features include passing tests in `.spec.ts/tsx` files **35 tests**, all passing (23 in `source/repo-map/index.spec.ts`, 12 in `source/commands/repomap.spec.tsx`)
- [ ] All existing tests pass (`pnpm test:all` completes successfully) **cannot be checked; see note below**

<details>
<summary>What the 35 tests cover</summary>

Ranking order, symbol ordering by cross-file weight, even weight split across multiple definers, self-reference exclusion, empty repos, symbol-less files, `.gitignore` + default-ignore dirs, per-file byte cap, file-count cap, token-budget truncation, per-file symbol cap, comment/string stripping, export-only indexing, import-scoped references, property-access exclusion, CRLF sources, per-language extraction across all nine language groups, cycle convergence with normalised ranks, deterministic ordering across runs, missing directory, the full `--tokens` argument surface (inline, separated, fractional, clamped, below-minimum, missing, non-numeric, unknown flag), and real Ink renders of the populated / truncated / empty views plus the error path.

</details>

**On `pnpm test:all`:** it does not complete on `main` today, and this PR neither causes nor fixes that. It stops at `test:types`, because the pinned TypeScript 7.0.2 rejects the repo's own `tsconfig.json` (`TS5102: Option 'baseUrl' has been removed`). `test:ava` separately refuses to start under the pinned AVA 8 (`The extensions option must be an array`  `package.json` still has the object form). Both reproduce on a clean `upstream/main` tree with none of my files present, and `git diff upstream/main` confirms this branch touches neither config.

Stage-by-stage, on this branch:

| Stage | Result |
|---|---|
| `test:format` | ✅ pass |
| `test:types` | ❌ pre-existing `tsconfig.json` `baseUrl` vs TypeScript 7 |
| `test:types:vscode` | ❌ pre-existing `TS6059 rootDir` in `plugins/vscode/` under TS 7 |
| `test:lint` | ✅ pass (2 pre-existing warnings in `chat-panel-harness.ts`) |
| `test:ava` | ❌ pre-existing config rejection; with only `ava.extensions` shimmed to an array, the suite runs my 35 pass, and the 19 failing files are **byte-identical** to a clean-`main` baseline (Windows path/shell specs: `getTasksPath` expecting `.nanocoder/tasks.json`, bash-executor, clipboard, CLI discovery) |
| `test:knip` | ✅ pass |
| `test:audit` | ✅ pass |
| `test:security` | ⏭️ skipped semgrep not installed locally |

I ran the failing-file comparison twice (AVA 7 and AVA 8) against a stashed clean tree; the set was identical every time.

### Manual Testing

- [ ] Tested with Ollama
- [ ] Tested with OpenRouter

## Checklist

- [x] If this was for an open issue, I was assigned to it assigned on #890
- [x] Code follows project style guidelines biome clean, tabs/single quotes, `@/*` aliases
- [x] Self-review completed
- [x] Documentation updated (if needed) `docs/features/commands.md`
- [x] No breaking changes (or clearly documented) purely additive: one new module, one new command file, one registry entry
- [ ] Appropriate logging added using structured logging **none added, deliberately.** The command is synchronous request/response with no background work or failure modes worth tracing; errors surface directly in the returned `errorMsg` element. Happy to add `logger` calls to the scan if you'd rather have visibility into per-run file counts and timings.

### DIFF
--- a/.changeset/repo-map-command.md
+++ b/.changeset/repo-map-command.md
@@ -1,0 +1,5 @@
+---
+"@nanocollective/nanocoder": minor
+---
+
+Added `/repomap`, a local codebase map. Nanocoder indexes the symbols defined in each source file, links files that reference each other's symbols into a directed graph, and ranks that graph with PageRank - so the map leads with the files the rest of the codebase leans on most. Everything runs on your machine with no LLM round-trip, covering TypeScript/JavaScript, Python, Go, Rust, Java/Kotlin/C#/Swift, Ruby, PHP, and C/C++. The map is budgeted to 1024 tokens by default; `/repomap --tokens <n>` widens it. Thanks to @akramcodez. Refs #890.
--- a/docs/features/commands.md
+++ b/docs/features/commands.md
@@ -34,6 +34,7 @@ Type `/` in the chat input to see available commands. All commands start with `/
 | `/update` | Update Nanocoder to the latest version |
 | `/usage` | Get current model context usage visually |
 | `/lsp` | List connected LSP servers |
+| `/repomap` | Show a PageRank-ordered map of the codebase - the most-referenced files and the symbols they define. Use `/repomap --tokens <n>` to widen the map beyond its default 1024-token budget |
 | `/schedule` | Read-only view of cron subscriptions declared by skills (see [Skills → Event subscriptions](skills.md#event-subscriptions)) |
 | `/skills` | List and inspect loaded skills; scaffold new bundle skills with AI assistance (see [Skills](skills.md)) |
 | `/resume` | Resume a previous chat session (aliases: `/sessions`, `/history`). Also available at launch via the `--resume`/`--continue` CLI flags. See [Session Management](session-management.md) |
--- a/source/commands/lazy-registry.ts
+++ b/source/commands/lazy-registry.ts
@@ -217,6 +217,12 @@ export const lazyCommands: LazyCommand[] = [
 		load: () => import('@/commands/skills').then(m => m.skillsCommand),
 	},
 	{
+		name: 'repomap',
+		description:
+			'Show a ranked map of the codebase (files and their key symbols). Use --tokens <n> to widen it.',
+		load: () => import('@/commands/repomap').then(m => m.repomapCommand),
+	},
+	{
 		name: 'privacy',
 		description:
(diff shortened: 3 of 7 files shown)


addyCooks:
Hey @will-lamerton @akramcodez 

heads up, the red CI here isn't this PR : )

`main` is broken for everyone. repro on a clean checkout, zero changes:
$ pnpm test:types
tsconfig.json(19,3): error TS5102: Option 'baseUrl' has been removed.

fix is up separately in #921 since it unblocks the whole repo, not just me. this branch is feature-only 7 files, 
all repo-map.
goes green once that lands and I rebase

will-lamerton:
Great work @addyCooks :)

addyCooks:
Thanks @will-lamerton !!

PR #920, opened 2026-08-22
Toggle todo

bigattichouse:
## Description

Closes #918.

The todo list eats a lot of vertical space in the transcript, which is wasteful once you already know what the agent is working on. **Ctrl-T** collapses it into a single segment on the status bar and brings it back on demand.

While collapsed the segment keeps the progress that was on screen, so hiding the list doesn't mean losing track of it:

| Segment | Meaning |
|---|---|
| `Todo (2/5 Ctrl-t)` | two of five done |
| `Todo (~2/5 Ctrl-t)` | ...and one currently in progress |
| `Todo (~2/5* Ctrl-t)` | ...and the list changed since you hid it |

The unread marker renders in the warning colour, so a change while the list is collapsed is visible without expanding it. On narrow terminals the key hint is dropped, and the segment is measured into the existing width budget alongside the tune and context segments so it collapses under pressure like everything else on that bar.

Visibility lives in `useAppState` next to the other session-scoped UI state; the key is handled in `user-input` alongside the existing control shortcuts.

## Type of Change

- [ ] Bug fix
- [X] New feature
- [ ] Breaking change
- [ ] Documentation update

## Changeset

- [X] Added a changeset (`minor`)

### Automated Tests

- [X] New features include passing tests in `.spec.ts/tsx` files — 134 tests across `chat-input`, `development-mode-indicator`, `user-input`, `useAppState` and `interactive-app`
- [X] All existing tests pass
- [X] Tests cover both success and error scenarios

> [!NOTE]
> Rebased onto `main` now that #923 has landed. One remaining check depends on #925, which fixes a `FORCE_COLOR` sensitive assertion in `location-step.spec.tsx` that is unrelated to this branch. Everything else passes: `format`, `lint`, `types`, `types:vscode`, `knip`, `audit` and `build`.

### Manual Testing

- [X] Tested with Ollama

## Checklist

- [X] Code follows project style guidelines
- [X] Self-review completed
- [X] No breaking changes

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01J7jWv6TLZ5UCr5BVfE2FP7

PR #925, opened 2026-08-23
fix: two environment-sensitive test failures blocking CI

bigattichouse:
## Description

Follow-up to #923. With CI now getting as far as the test step, two long-hidden test failures are the last red checks. Both are test-only; no source changes.

### 1. `location-step` compares a styled line against a bare string

`LocationStep shows the resolved project path next to the option` compares a rendered Ink line against a bare path with only `.trim()` applied. The styled path carries SGR escapes that `trim()` doesn't touch, so the test passes whenever colour is off and fails whenever it's on — with characteristically baffling output, two visually identical strings:

```

- '/test/project'
+ '/test/project'
```

CI sets `FORCE_COLOR=1`, so this has been broken there all along. `strip-ansi` is already a devDependency and is the convention the other render-comparing specs use (see `user-input.spec.tsx`).

### 2. `prompt-history` teardown races its own debounced write

The `after.always` hook already passed `recursive`/`force`, but `addPrompt` persists on a debounced timer, so a write can land while `rmSync` is walking the tree:

```
ENOTEMPTY: directory not empty, rmdir '/tmp/nanocoder-test-...'
1 hook failed
11 tests skipped
```

Node exposes `maxRetries`/`retryDelay` on `fs.rm` precisely for this class of error, so the teardown now backs off instead of failing the run. This one is intermittent — it passed on two other runs of the same suite, which is what makes it worth pinning down rather than just re-running.

## Type of Change

- [X] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Changeset

- [X] Added a changeset (empty — test-only, nothing user-facing to release)

## Testing

Full suite under CI conditions (`FORCE_COLOR=1 CI=true NODE_ENV=test`): **6565 passed, 0 failed**.

`location-step.spec.tsx` now passes with `FORCE_COLOR=1` and `FORCE_COLOR=0`, where previously only the latter passed.

## Checklist

- [X] Code follows project style guidelines
- [X] Self-review completed
- [X] No breaking changes

https://claude.ai/code/session_01J7jWv6TLZ5UCr5BVfE2FP7


bigattichouse:
Green across all 11 checks — this is the first fully green run the repo has had since the TypeScript 7 and AVA 8 bumps, now that #923 has landed and CI actually reaches the test step.

Flagging that this unblocks #920: that branch is rebased onto current `main` and passes 10 of 11 checks, with the only remaining failure being the `location-step` assertion fixed here. Nothing else stands between it and green.

Both fixes are test-only, no source changes:

- **`location-step`** compared a styled Ink line against a bare string with only `.trim()`, so it passed without colour and failed with it. CI sets `FORCE_COLOR=1`, so it has been broken there the whole time — it was just never reached, because the coverage job used to die at `pnpm build` or at AVA startup first.
- **`prompt-history`** raced its own debounced write during teardown. That one is intermittent; it passed on two other runs of the same suite, which is why it seemed worth pinning down rather than re-running.

Happy to split these into two PRs if you'd prefer them reviewed separately.

will-lamerton:
Thanks for this @bigattichouse :D

PR #928, opened 2026-08-24
feat: migrate file search to a ripgrep-backed implementation

### LINKED ISSUE #889: [Feature] Replace Hand-rolled JS File Search with Ripgrep
### Description
Replace the manual filesystem walk and regex search in `source/utils/file-search.ts` with a bundled `rg` (ripgrep) binary to drastically improve file search speeds.

### Use Case
On large repositories, the current JS-based file search (`lstat`, `readdir`, per-line regex) is 10–50x slower than ripgrep. `search_file_contents` is one of the most frequently called tools, so this causes noticeable lag.

### Proposed Solution
Vendor a ripgrep binary (similar to how VS Code does it) and shell out to it for `search_file_contents`.

### Alternatives Considered
Offloading the JS walk to a worker thread (Tier 2 improvement, but still much slower than ripgrep).

### Additional Context

### Implementation Notes
This is the biggest raw speed win for large codebases.

kishore280:
- Replaces the hand-rolled JS file walker/glob matcher with `ripgrep`.
- Glob matching is now a DP-table matcher instead of regex, so no ReDoS.
- `.gitignore`-aware traversal and consistent behavior across platforms, Windows included.
- Hit and fixed a real Node `spawn({timeout})` bug along the way - upstream at `nodejs/node#65506`.
- Closes #889.


will-lamerton:
Nice work @kishore280 - the DP glob matcher is a real improvement and the edge-case tests around it read like they came from actual failures. A few things need addressing before this lands.

**Blocking**

1. **Old system `rg` fails silently.** `resolveRipgrepPath` prefers whatever `rg` is on PATH and only checks that it prints `ripgrep `. The args use `--no-require-git`, which landed in ripgrep 13.0.0 (Ubuntu 20.04 ships 11.0.2, bullseye 12.1.1). Those exit 2 with `error: Found argument '--no-require-git'...`, which matches none of `FATAL_RIPGREP_ERROR_PATTERNS`, so the close handler falls through to `resolve(stdout)` with empty stdout. `find_files`, `search_file_contents`, `@`-autocomplete and `/repomap` all return zero results forever with no error. Suggest dropping the system probe entirely (the bundled `@vscode/ripgrep` is already a hard dep with all platform optionals pinned), and inverting the error check so `code > 1 && stdout.length === 0` rejects regardless of whether stderr is recognized. Note CI never exercises the bundled path - coverage shows `ripgrep-path.ts:27-29` uncovered.

2. **Add `--no-config`.** rg reads `RIPGREP_CONFIG_PATH` unless told not to. A user with `--follow` in their rc file defeats the symlink containment this PR's own tests assert; `-uu` defeats gitignore filtering.

3. **Resolve `rg` to an absolute path.** `systemRipgrepPath` returns the bare string `'rg.exe'`, and libuv's Windows path search checks the current directory before PATH. Opening nanocoder in a repo containing `rg.exe` executes it during the probe, and it then gets cached and trusted. POSIX is fine, but this PR is specifically targeting Windows.

**Regressions needing tests**

4. **`.nanocoderignore` is dropped.** The old walker went through `loadGitignore(cwd)`, which layers it on. The rg path only reproduces `DEFAULT_IGNORE_DIRS`. `list-directory` and `file-tree` still honor it, so the file tree now hides files that search returns - and this contradicts the documented contract in `gitignore-loader.ts:52`. Fix is `--ignore-file <cwd>/.nanocoderignore` when present, plus the same in `walkEmptyDirectories`.

5. **`find_files` can't find binary-extension files.** `binaryExcludeGlobs()` is now applied to `walkProjectEntries`, but binary filtering previously only existed in content search. `find_files "*.svg"` / `"*.png"` / `"*.wasm"` return nothing, and `@`-autocomplete stops suggesting them. `.svg` especially. Suggest keeping those globs only in `searchProjectContents`.

**Worth a look**

- Regex dialect went JS to Rust, so no lookahead/lookbehind/backrefs. The tool description at `search-file-contents.tsx:116` still promises "extended regex" and models emit `(?=...)` regularly. Either add `--pcre2` or update the description.
- A nonexistent `searchPath` now returns empty instead of throwing (old code `lstat`'d it). Same root cause as item 1.
- `walkEmptyDirectories` readdirs the whole tree, not just empty dirs, so every call pays an rg pass plus a full JS walk - and two of three consumers discard directory entries anyway. An `includeDirectories` option would let only `findMatchingPaths` pay for it.
- With `contextLines`, the kill fires on the Nth match line before its trailing context flushes, so the last block is silently short. `maxResults + 1` as `maxMatches` would fix it; the overlap test doesn't reach this.
- `expandBraces` is untouched, so `MAX_GLOB_PATTERN_LENGTH` still admits ~200 `{a,b}` groups. The ReDoS is gone but this DoS path isn't, and the pattern is LLM-supplied.
- `stdout += chunk` is unbounded in `--files` mode, and `maxResults` no longer bounds rg's work the way the old early-stopping walker did.
- `resolveRipgrepPath caches the result across calls` passes trivially because the prior test warmed the module cache.

CI red here is the known pre-existing `location-step` / `vscode-server` failures on main, not this PR.

will-lamerton:
Great work @kishore280 :)

Issue #861, opened 2026-08-12
[Feature] Agentic CI/CD Gate Phase 2: Read-Only PR Review (`verify --pr`)


akramcodez:
## Description
Code reviews take significant developer time, and the agent currently requires manual prompting inside an interactive session to review code. A headless PR review mode is needed to allow Nanocoder to automatically read a PR diff, run static analysis, and post a structured code review comment.

## Expected Behavior
* Implement the `verify` CLI subcommand parser in `source/cli.tsx` to handle flags like `--pr <n>` and `--post-review`.
* Create a built-in subagent (`verify-pr-review.md`) optimized for code review using the tools built in Phase 1.
* Run a static analysis pass using `semgrep` (`source/verify/security-scan.ts`) and inject the findings into the LLM context.
* Validate and format the subagent's raw output into a structured review comment before posting.
* If `--post-review` is absent, print the structured review to stdout for local dry-runs.

## Actual Behavior
Users must manually check out a PR, open Nanocoder, and prompt it to read diffs and find issues. The agent cannot act as an automated, headless PR reviewer.

## Goal
Enable a command (`nanocoder verify --pr <n>`) that performs a comprehensive, read-only analysis of a PR and posts its findings, acting as a tireless automated code reviewer. *(Note: All PRs for this issue should be opened against the `feature/agentic-cicd-gate` branch.)*

rohanshrma222:
@akramcodez  Working on the 2nd phase

akramcodez:
cool @rohanshrma222 go for it 👍

PR #940, opened 2026-08-24
feat(verify): add headless PR review command

rohanshrma222:
## Description

Implements issue #861 (Phase 2 of the Agentic CI/CD Gate roadmap): adds `nanocoder verify --pr <n>`, a headless CLI command that reviews a pull request and produces a structured Markdown review. It runs a semgrep static-analysis pass and a new read-only `verify-pr-review` subagent (using Phase 1's `git_pr` diff/checks/logs actions), then prints the combined review to stdout by default, or posts it as a single PR comment with `--post-review`. The reviewing subagent never posts to GitHub itself — the CLI harness formats and posts the review after the subagent returns, since a headless run has no UI to approve a mutating tool call.

Also includes a small correctness fix carried over from Phase 1: reverts `git_pr`'s `readOnly: true` (added in the prior PR to enable parallel-batching of read-only calls) after finding that `resolveToolApproval` bypasses per-tool approval checks entirely in yolo mode — which could have let a mutating `git_pr` call (create/comment/review) get parallel-batched with other calls instead of running alone.

## Type of Change

- [x] New feature
- [ ] Bug fix
- [ ] Breaking change
- [ ] Documentation update

## Changeset

### Automated Tests

- [x] All existing tests pass (ran the full relevant suite — `verify/*`, `git-pr`, `tool-manager`, `subagent-loader`, `trust` — plus `test:types`/`test:lint`/`test:format`/`test:knip`, all clean)

### Manual Testing

- [ ] Tested with Ollama
- [ ] Tested with OpenRouter

> Manual testing note: I verified the CLI wiring end-to-end against the compiled binary (`--help` text, flag validation, error paths) and confirmed the `verify-pr-review` subagent loads correctly via the real subagent loader — but haven't run a full live `verify --pr <n>` against a real LLM provider/PR, since that needs your `gh`/provider credentials. Worth doing before merge if you can.

## Checklist

- [x] Self-review completed (ran `/code-review`, fixed the findings worth fixing)


will-lamerton (commented):
Nice work on this, the CLI layer is clean and the deps seam in `cli.spec.ts` is the right pattern. I checked your two safety claims against the code and both hold up: `approval-policy.ts` does short-circuit yolo before the per-tool `approval` fn, so the `readOnly` revert is correct, and `tool-approval-queue`'s slot defaults to `() => false`, so harness-posting is the right call.

A few things I'd like addressed before merge:

**1. Nothing establishes the working tree is the PR's head.** `runSecurityScan({cwd: projectRoot})` scans whatever is checked out, and the subagent prompt tells the model to read the local tree "around changed lines". Reviewing someone else's PR from your own branch means both are reading unrelated code while `git_pr diff` returns the real diff. Either `gh pr checkout` first, or fail fast when the current branch does not match the PR's `headRefName`. Related: the value passed as `targetBranch` is the current local branch, not the PR's base, which will mislead the model.

**2. semgrep scans the whole repo rather than the diff.** With `MAX_FINDINGS = 50` sorted by severity, 50 pre-existing repo-wide ERRORs can crowd out everything the PR actually introduced, and `formatFindingsSection` never discloses the truncation. Please pass the changed paths explicitly, and say "50 of N" when it caps. This also mostly resolves the 120s timeout, which I suspect is too short today for `--config auto` plus a full-tree scan, and whose failure mode is a quiet "scan failed to run" line in a posted review.

**3. Drop `fetch_url` and `web_search` from `verify-pr-review`.** A PR diff is attacker-controlled text and headless mode auto-executes both without approval, so an injected instruction can exfiltrate local file contents via a fetch URL, and the output then gets posted publicly. A reviewer that reads the diff and the local tree does not need to browse.

**4. Please consume `verify/trust.ts`.** `tool-manager.ts:57` says verbatim that trust.ts is meant to replace the blunt `git_pr` exclusion "once the verify subcommand (issue #861) consumes it", and this is #861. Right now the tool list is hardcoded in the subagent frontmatter instead, so there are two allowlists that will drift. This also fixes a real gap: `MODE_EXCLUDED_TOOLS.headless` is only `['ask_user', 'agent']`, so `git_pr` is available with all actions and only the approval queue's default-deny stops a mutating call. A `verify` run is currently less restricted than plan mode.

**5. Test coverage for `security-scan.ts`.** All the parsing logic is untested: `parseFindings` (severity normalization, sort, the cap) and `runSemgrepProcess` (timeout, empty stdout, spawn error). The one `runSecurityScan` test `t.pass()`es whenever semgrep is installed, so it asserts nothing on most machines. Export `parseFindings`, or add a process-runner dep seam like `cli.ts` already has.

Minor, take or leave:

- `parseArgs` silently ignores unknown flags, so `--post-revieww` quietly does a dry run
- `--provider`/`--model` are in `USAGE` but missing from `cli.tsx`'s `--help`
- `init.client as LLMClient` is redundant, `PlainInitResult.client` is already that type
- The `Promise.all` comment claims overlap, but `isSemgrepAvailable()`'s `execSync` blocks the loop through semgrep's startup first
- The SIGINT handler aborts without exiting and suppresses the default handler, so Ctrl-C looks dead if the executor is slow to honor it
- `docs/features/` has a page per user-facing feature, `verify` adds none

Your caveat about no live run stands, and items 1 to 3 are exactly what a live run would have surfaced.

rohanshrma222:
> Nice work on this, the CLI layer is clean and the deps seam in `cli.spec.ts` is the right pattern. I checked your two safety claims against the code and both hold up: `approval-policy.ts` does short-circuit yolo before the per-tool `approval` fn, so the `readOnly` revert is correct, and `tool-approval-queue`'s slot defaults to `() => false`, so harness-posting is the right call.
> 
> A few things I'd like addressed before merge:
> 
> **1. Nothing establishes the working tree is the PR's head.** `runSecurityScan({cwd: projectRoot})` scans whatever is checked out, and the subagent prompt tells the model to read the local tree "around changed lines". Reviewing someone else's PR from your own branch means both are reading unrelated code while `git_pr diff` returns the real diff. Either `gh pr checkout` first, or fail fast when the current branch does not match the PR's `headRefName`. Related: the value passed as `targetBranch` is the current local branch, not the PR's base, which will mislead the model.
> 
> **2. semgrep scans the whole repo rather than the diff.** With `MAX_FINDINGS = 50` sorted by severity, 50 pre-existing repo-wide ERRORs can crowd out everything the PR actually introduced, and `formatFindingsSection` never discloses the truncation. Please pass the changed paths explicitly, and say "50 of N" when it caps. This also mostly resolves the 120s timeout, which I suspect is too short today for `--config auto` plus a full-tree scan, and whose failure mode is a quiet "scan failed to run" line in a posted review.
> 
> **3. Drop `fetch_url` and `web_search` from `verify-pr-review`.** A PR diff is attacker-controlled text and headless mode auto-executes both without approval, so an injected instruction can exfiltrate local file contents via a fetch URL, and the output then gets posted publicly. A reviewer that reads the diff and the local tree does not need to browse.
> 
> **4. Please consume `verify/trust.ts`.** `tool-manager.ts:57` says verbatim that trust.ts is meant to replace the blunt `git_pr` exclusion "once the verify subcommand (issue #861) consumes it", and this is #861. Right now the tool list is hardcoded in the subagent frontmatter instead, so there are two allowlists that will drift. This also fixes a real gap: `MODE_EXCLUDED_TOOLS.headless` is only `['ask_user', 'agent']`, so `git_pr` is available with all actions and only the approval queue's default-deny stops a mutating call. A `verify` run is currently less restricted than plan mode.
> 
> **5. Test coverage for `security-scan.ts`.** All the parsing logic is untested: `parseFindings` (severity normalization, sort, the cap) and `runSemgrepProcess` (timeout, empty stdout, spawn error). The one `runSecurityScan` test `t.pass()`es whenever semgrep is installed, so it asserts nothing on most machines. Export `parseFindings`, or add a process-runner dep seam like `cli.ts` already has.
> 
> Minor, take or leave:
> 
> * `parseArgs` silently ignores unknown flags, so `--post-revieww` quietly does a dry run
> * `--provider`/`--model` are in `USAGE` but missing from `cli.tsx`'s `--help`
> * `init.client as LLMClient` is redundant, `PlainInitResult.client` is already that type
> * The `Promise.all` comment claims overlap, but `isSemgrepAvailable()`'s `execSync` blocks the loop through semgrep's startup first
> * The SIGINT handler aborts without exiting and suppresses the default handler, so Ctrl-C looks dead if the executor is slow to honor it
> * `docs/features/` has a page per user-facing feature, `verify` adds none
> 
> Your caveat about no live run stands, and items 1 to 3 are exactly what a live run would have surfaced.

Thx for the review, i will try to address all of these and will self review it.

will-lamerton:
Thanks @rohanshrma222 - once done I'd be happy to merge into the feature branch :)

Let me know @akramcodez the plan for the feature branch or whether this can go straight to main!

rohanshrma222:
While in there, I ran another pass myself and found a few more, not in the above list:
- git_pr's error banner was a substring match on "Error:" -  false-positived on any CI log/diff containing that text. Fixed to startsWith.
- The two new gh metadata calls had no timeout (unlike the log fetch next to them).
- My branch check compared names, which breaks under detached HEAD (normal for CI checkouts) - switched to SHA comparison.
- gh pr diff --name-only includes deleted/renamed files, which broke semgrep if passed straight through - now filtered against the checkout.
- Very large PRs could hit an argv length limit passing changed-file paths to semgrep - added a cap with whole-tree fallback.

Left two things alone: the duplicated timeout/kill logic between execProcess and runSemgrepProcess (different resolve semantics, didn't want to risk unifying them here), and git_pr's actions being tracked across four separate allowlists (already flagged in the code as a known stopgap for a later phase). Can split either into a follow-up if you'd rather.

PR #943, opened 2026-08-24
feat(vscode): show created and modified files in the context panel

### LINKED ISSUE #857: [Feature] Show created and modified files in the context panel
## Description

When the agent creates, edits, or updates a file during a task, the affected file should automatically appear in the context/editor panel.

Currently, file operations can complete successfully, but the created or modified files are not surfaced in the context panel. This makes it harder to immediately inspect what the agent changed.

The experience should be similar to VS Code: when a particular file is created or modified and the user selects it, the file should open in the editor/context panel so the changes can be reviewed immediately.

## Expected Behavior

When the agent creates or modifies a file:

1. Detect the affected file.
2. Add it to the context panel.
3. Allow the user to select the file and view its current contents in the editor.
4. For modified files, show the latest version after the agent's changes.

For example:

**Agent:** Creates `src/components/Button.tsx`

The file should automatically become available in the context/editor panel, allowing the user to open and review `Button.tsx` without manually searching for it.

## Actual Behavior

The agent creates or modifies the file successfully, but the file does not appear in the context/editor panel.

The user has to manually locate the file instead of being able to immediately inspect the changes from the UI.

## Why This Matters

The context panel should provide a clear connection between the agent's actions and the files affected by those actions. Automatically surfacing created and modified files would make it much easier to review agent changes and understand what happened during a task.

## Additional Context

This should apply to file creation, modification, and potentially deletion events where appropriate. The goal is to provide a VS Code-like workflow where files affected by the agent are immediately accessible for review.

addyCooks:
## Description

A file the agent wrote was only reachable through its edit card's diff - there was no list of what a turn had touched, and nothing that opened the file itself. Every edit that completes now adds the file to the context row above the composer, where clicking it opens the current version in the editor and the x dismisses it.

The path is taken from the tool call's ACP `locations` (falling back to the diff block) and remembered per tool call id until the call lands: the completion update carries a status and nothing else, so by then the path is off the wire. An edit that was denied, cancelled or failed changes nothing, so it contributes nothing.

Agent entries share the row with the files the user attached but are not the same thing, so they are dashed, are left out of the next prompt - the agent just wrote them, and re-inlining a file up to 100KB would spend a local model's whole context window to say nothing new - and survive a send instead of being consumed by it. Attaching one deliberately promotes it back to an ordinary attachment. Starting or resuming a conversation clears them.

The row follows the rest of the file lifecycle too: a delete takes its chip away - the user's own attachment included, since it would otherwise ride along on the next prompt and expand to nothing - and a rename moves the chip to the new path instead of leaving one pointed at a file that has moved.

Two tools had to start describing themselves accurately over ACP for that to work. `diff_edit` reported kind `other`, so it had neither a file card nor a chip - and it is the `nano` profile's editing tool, so on a small local model it was the only writer a client ever saw. `file_op` fell through `buildToolCallMeta`'s generic branch, so a delete, a rename and a copy all arrived as an untyped call naming their source path; it now reports the kind its operation actually performs (`delete`, `move`, `edit` for a copy, `other` for `mkdir`), and `locations` keeps one rule - the file the call leaves behind comes last - so a move reports its source then its destination.

Closes #857.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

## Changeset

### Automated Tests

### Manual Testing

- [ ] Tested with Ollama
- [ ] Tested with OpenRouter

## Checklist

- [x] Self-review completed


will-lamerton (changes requested):
Nice work - the core mechanism is right. Caching `locations` per `toolCallId` is the correct call given `emitToolCallUpdate` only sends status/rawOutput/title, gating on `completed` so denied and cancelled calls leave the row alone is the right posture, and the `diff_edit` / `file_op` ACP mappings are worth having on their own merit. Test coverage is thorough. I ran `test:types`, `biome check`, `knip` and both spec files locally: all clean.

A few things before this lands.

**1. The chip row is unbounded, with no bulk dismiss.** `#context-chips` is `flex flex-wrap` with no `max-height` and no overflow, inside a `shrink-0` composer (`chat-panel.html:234`). That was fine when the row only held what a user hand-attached. Now a refactor turn touching 30 files produces 30 chips at up to 200px each in a ~300px sidebar, growing the composer and squeezing the message list, and the only way out is clicking the x thirty times or starting a new conversation. Please add `max-h-24 overflow-y-auto` to the container, plus either a "clear changed files" control or a cap with a "+N more" chip.

**2. `updateLocations` does not check the path is absolute, but a test says it does.** The test at `chat-panel-tool-cards.spec.ts:306` is named "an edit that names no absolute path is skipped" and its comment says the host can only open an absolute one, but the filter is just `typeof path === 'string' && !!path`. It passes because that case has no path at all. A relative location from an MCP or custom tool would produce a chip that `vscode.Uri.file()` resolves against the filesystem root, and `showTextDocument` on the missing path rejects unhandled at `chat-webview-provider.ts:377`. A one-line guard would make the code match the test's claim.

**3. Path matching is exact string equality across two different producers.** Agent chips come from `resolve()` in `acp-tool-call.ts`; user chips come from `uri.fsPath` in the extension host. `addChangedFileChip`, the promotion branch in `attachPath` and `dropChip` all compare with `===`. On Windows those two sources can disagree on drive-letter case, which shows the same file twice (one solid, one dashed) and breaks promotion. Normalizing before comparison closes it.

**4. `buildFileOpMeta` reports a location for `mkdir` while its comment says there is nothing to open.** `if (path && operation !== 'copy')` pushes the directory path for `mkdir` too. Harmless in this panel, but ACP follow-along clients treat `locations` as things to open, so excluding `mkdir` would match the stated intent. The mkdir test only asserts `kind`, not `locations`.

Smaller, take or leave:

- `resolveEditCardState` is now a general status normalizer for `delete` and `move` as well, and its doc comment still says "an edit card's label and icon". A note at the `trackFileChanges` call site would save a future reader the double take.
- `file_op` delete and move land in the aggregated list with a raw `file_op: move /a → /b` title, since `TOOL_VERBS` has no entry. Same as before this PR, but the comment above `TOOL_VERBS` was updated for the edit family and does not mention it.
- `announceFileOp` in the spec builds titles with `->` while `buildFileOpMeta` emits `→`. Nothing asserts on it, but the fixture drifts from the real thing.

One pre-existing thing worth flagging rather than fixing here: `acp-tool-call.ts` resolves with a bare `resolve(path)` while `diff_edit` and `file_op` resolve against `getSafeSessionCwd()` (`diff-edit.tsx:252`). Not a regression, but this PR promotes `locations` from decoration into a clickable affordance, so a divergent session cwd now produces a visibly broken chip. Fine as a follow-up.

Finally, could you confirm a real VS Code session with a multi-file turn? The harness coverage is good, but issue 1 lives exactly where the harness cannot see.

addyCooks:
Thanks @will-lamerton,
all four addressed, plus the smaller ones. Pushed in 73c12044, main merged in.

**1. Unbounded row.** `max-h-24 overflow-y-auto` on `#context-chips`, and a
**Clear N changed files** control *outside* the scroll box so it stays reachable. Clears
only the agent's chips. Checked the utility actually compiles, not just lands in markup:
`.max-h-24{max-height:calc(var(--spacing) * 24)}`.

**2. Absolute-path guard.** `isAbsolutePath` (POSIX, UNC, drive-qualified), applied to
`locations` and the diff fallback. Renamed the test to what it covered and added real
relative cases. The host-side `showTextDocument` at chat-webview-provider.ts:450 is still
a floating promise, so this stops the bad chip existing rather than handling the rejection.

**3. Normalisation.** `pathKey` folds separators and the drive letter only the rest is
untouched so it stays correct on case-sensitive filesystems. Routed through
`addChangedFileChip`, `attachPath`, `dropChip`, and the x-remove handler, which compared
raw too.

**4. mkdir** reports no location; its test now asserts `locations`, not just `kind`.

Smaller: notes at the `trackFileChanges` call site and on `resolveEditCardState`;
`TOOL_VERBS` comment covers `file_op`; fixture uses `→`. Left the `resolve()` vs
`getSafeSessionCwd()` divergence as a follow-up.

**Conflicts.** Both from the Tailwind v4 migration. `chat-panel.css` was modify/delete 
took the deletion, the chip style already lives in `src/styles.css`. In `chat-panel.js`
both sides improved the same two lines (`kind` hoisted here, `existingCard` on main) 
kept both.

**On the live session, I haven't done one.** There's a harness test driving 30 edits and
asserting the control appears and empties the row, but how 96px of chips sits in a ~300px
sidebar is exactly what the harness can't see. Please treat item 1 as visually unconfirmed.

**Verification.** Mutation pass over the fixes caught 7/9 first time the two escapes were
genuinely untested fixes of mine (diff-fallback guard, `addChangedFileChip` dedupe). Added
tests; 9/9 now. Plus 130/130 tests, tsc, biome, knip, changesets clean; CI green, coverage
92.21% vs the 92.17% baseline.

Also fixed a harness gap: it built `#context-chips` without the `hidden` class the markup
carries, so an empty panel looked different in tests than on load.

akramcodez (approved):
LGTM

will-lamerton (approved):

PR #944, opened 2026-08-24
Display compound bash commands on separate lines in execute_bash output

### LINKED ISSUE #1262: [Feature] Split compound bash commands onto separate lines in the tool preview
## Description

Display compound bash commands on separate lines in execute_bash output
* Part1 `;`
  * https://github.com/Nano-Collective/nanocoder/pull/944
* Part2 `&&` `||`
  * Pending implementation

Before:
```
Command:
 cd /home/<>/nanocoder && ls -la dist/utils/shell-command-display.js 2>&1; echo "---global---"; NPM_BIN=$(dirname "$(readlink -f "$(which nanocoder)")"); echo "bin dir: $NPM_BIN"; find
"$NPM_BIN/.." -maxdepth 3 -name "shell-command-display.js" 2>/dev/null
```

After:
```
Command:
cd /home/<>/nanocoder && ls -la dist/utils/shell-command-display.js 2>&1;
echo "---global---";
NPM_BIN=$(dirname "$(readlink -f "$(which nanocoder)")");
echo "bin dir: $NPM_BIN";
find "$NPM_BIN/.." -maxdepth 3 -name "shell-command-display.js" 2>/dev/null

```

## Use Case

See description

## Proposed Solution

Update display only

## Alternatives Considered

## Additional Context

* https://github.com/Nano-Collective/nanocoder/pull/944

coralsundy:
## Description
Display compound bash commands on separate lines in execute_bash output

Resolves https://github.com/Nano-Collective/nanocoder/issues/1262

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

## Changeset

### Automated Tests

### Manual Testing

- [ ] Tested with Ollama
- [ ] Tested with OpenRouter

## Checklist

- [ ] Self-review completed

### DIFF
--- a/.changeset/shell-command-display.md
+++ b/.changeset/shell-command-display.md
@@ -1,0 +1,5 @@
+---
+"@nanocollective/nanocoder": patch
+---
+
+Display compound bash commands on separate lines in execute_bash output
--- a/source/components/bash-progress.spec.tsx
+++ b/source/components/bash-progress.spec.tsx
@@ -53,6 +53,40 @@ test('BashProgress displays the command', t => {
 	t.regex(output!, /npm run build/);
 });
 
+test('BashProgress splits compound commands onto separate lines', t => {
+	const command = 'dolt version; echo "==="; ls -la /usr/local/bin/dolt';
+	const {lastFrame} = renderWithTheme(
+		<BashProgress
+			executionId="test-id"
+			command={command}
+			completedState={createCompletedState({command})}
+		/>,
+	);
+
+	const output = lastFrame();
+	t.truthy(output);
+	const lines = output!.split('\n').map(line => line.trim());
+	t.true(lines.includes('dolt version;'));
+	t.true(lines.includes('echo "===";'));
+	t.true(lines.includes('ls -la /usr/local/bin/dolt'));
+});
+
+test('BashProgress keeps a quoted semicolon on one line', t => {
+	const command = 'echo "hello; world"';
+	const {lastFrame} = renderWithTheme(
+		<BashProgress
+			executionId="test-id"
+			command={command}
+			completedState={createCompletedState({command})}
+		/>,
+	);
+
+	const output = lastFrame();
+	t.truthy(output);
+	const lines = output!.split('\n').map(line => line.trim());
+	t.true(lines.includes('echo "hello; world"'));
+});
+
 test('BashProgress displays execute_bash tool name', t => {
 	const {lastFrame} = renderWithTheme(
 		<BashProgress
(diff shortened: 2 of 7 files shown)


will-lamerton (changes requested):
Thanks for this, the readability goal is a good one. Two things need addressing before it can land.

**1. The preview no longer matches the command that runs.**

These two components are the tool-approval surface, so what they show is what the user consents to execute. Splitting on `/[;\n]+/` drops the separator and breaks shell syntax. Rendered output on this branch:

- `echo "hello; world"` shows as `echo "hello` / `world"` (the `;` is gone, neither line is valid)
- `find . -name "*.tmp" -exec rm {} \;` loses the `\;` terminator entirely
- `for f in *.ts; do echo $f; done` shows as three fragments that each look like a failing standalone command
- `awk '{print $1; print $2}' file` splits mid-script
- a heredoc loses blank lines in its body (`[;\n]+` is greedy), so the preview misstates the file being written

`git commit -m "fix; refactor"` displays a commit message that is not the one being used. Also worth noting `&&` and `||` aren't handled at all, despite "compound commands" in the title.

Options, roughly in order of preference:

- split only on top-level `;` outside quotes, and skip splitting entirely when the command contains a heredoc, an unbalanced quote, or a `for`/`while`/`if`/`case` keyword. A small quote-aware scanner is ~20 lines and covers every case above, and it extends naturally to `&&`/`||`.
- or at minimum keep the separator on the line it terminates (capturing-group split) so the display stays lossless.

**2. Both new tests pass without the change.**

I reverted the two `.tsx` source edits on this branch and re-ran the specs: all 57 pass, including the two new ones. The `t.regex` calls are substring matches against the whole frame, and ink-testing-library renders at a fixed 100 columns so the unsplit 52-character command never wrapped, meaning every regex matched before the change too. `t.is(matches.length, 1)` counts occurrences rather than lines, so it also passes either way.

Assert on lines instead, e.g. `t.true(output!.split('\n').includes('echo "==="'))`, and add cases for the quoted-semicolon, heredoc, and `for` loop inputs above.

**Minor:**

- the six-line render block is duplicated verbatim across both files. A shared `splitCommandForDisplay(command: string): string[]` helper with its own spec would give the tricky inputs one place to live.
- trailing or repeated separators yield empty segments that render as blank lines (`ls;` produces `ls` plus an empty `<Text>`). Filter empties after trimming.
- `.changeset/display-multiline-bash.md` has trailing whitespace and no newline at EOF. Biome doesn't cover `.changeset/*.md` so CI won't flag it.

Format and lint are clean, and execution is genuinely untouched, so the blast radius is limited to these two components. Happy to look again once the split is quote-aware and the tests actually fail without it.

coralsundy:
Take a look sometime this week.

coralsundy:
Hi @will-lamerton ,

I have re-implement the feature per your suggestions.
And the logic now sits in a small helper file.

It is currently only focus on `;` and ignored `&&` `||` as that would be in a second PR if needed.
It is already looking much better.

Thanks.

will-lamerton:
Great work @coralsundy :)

PR #945, opened 2026-08-24
 add showAgentBashOutput preference to keep agent bash output visible

### LINKED ISSUE #1293: [Feature] Keep agent `execute_bash` output visible on the completed card
## Description

PR #714 (commit `59dc92f5`) made user-typed `!commands` render their captured output on the completed `execute_bash` card. Agent-invoked `execute_bash` calls were left out: they always collapse to a compact card (command + status + token count), discarding the output that had streamed live while the command ran. Once the command finishes that output is gone from the transcript, even though the model still receives it.

One idea is to add a `showAgentBashOutput` preference to `nanocoder-preferences.json` that keeps the output on agent bash cards:

- `false` (default) — agent `execute_bash` cards stay compact, exactly as today
- `true` — agent `execute_bash` cards keep their output, even when compact tool display is on

Similarly to success run, a failed command keeps its error output rather than collapsing. Commands you type yourself (`!command`) already always show their output and are unaffected by this setting.

Users can set `"showAgentBashOutput": true`, or toggle it at `/settings` → **Behavior** → **Tool Results and Thinking** → **Agent Bash Output**.

Before / showAgentBashOutput=false
================================================================
⚒ execute_bash
Command:
printf "line one\nline two\nline three\n"
Status: ●
Tokens: ~7

After / showAgentBashOutput=true
================================================================
⚒ execute_bash
Command:
printf "line one\nline two\nline three\n"
Status: ●
line one
line two
line three
Tokens: ~7

## Implementation
* https://github.com/Nano-Collective/nanocoder/pull/945

## Use Case

Anyone who wants to see the bash execute output

## Proposed Solution

See description 

## Alternatives Considered

## Additional Context

This idea was initially implemented in a sloppy `user/tool/both/none` setup which is harder to maintain, and confusing to users, thus this new issue to re-design the approach, see:
* https://github.com/Nano-Collective/nanocoder/pull/945#pullrequestreview-5061116466

coralsundy:
### New design 20260913

add showAgentBashOutput preference to keep agent bash output visible

Resolves https://github.com/Nano-Collective/nanocoder/issues/1293

* https://github.com/Nano-Collective/nanocoder/issues/1293

Before / showAgentBashOutput=false
================================================================
⚒ execute_bash
Command:
printf "line one\nline two\nline three\n"
Status: ●
Tokens: ~7

After / showAgentBashOutput=true
================================================================
⚒ execute_bash
Command:
printf "line one\nline two\nline three\n"
Status: ●
line one
line two
line three
Tokens: ~7

<details>

<summary>Old design 2026/08/24 </summary>

add bashOutputMode preference to control completed bash output

This extends the showOutput mechanism introduced in [#714](https://github.com/Nano-Collective/nanocoder/pull/714) (commit 59dc92f5), which made user-typed !commands render their captured output on the completed execute_bash card while agent-invoked execute_bash calls always collapsed to a compact card (command + status dot + token count) — discarding the output that had streamed live while the command ran.

This adds a bashOutputMode preference to nanocoder-preferences.json that controls which execution sources retain their output on the completed card:

user — only user-typed !commands show output (default, preserves original behavior)
tool — only agent execute_bash calls show output
both — every bash execution shows output
none — all completed cards stay compact

The two call sites resolve the mode against their known execution source, the BashProgress component is unchanged, and the setting is documented in the preferences reference with tests covering the accessors. Users opt into seeing agent bash output with "bashOutputMode": "both".

This is to introduce similar behavior compares to other coding agent, plus, me personally really likes to see output trail as it is very useful for debug instead of just let AI with all the information.

</details>

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

## Changeset

### Automated Tests

### Manual Testing

- [ ] Tested with Ollama
- [ ] Tested with OpenRouter

## Checklist

- [ ] Self-review completed


will-lamerton (changes requested):
Thanks for this. The concept is right and generalising the `showOutput` flag from #714 rather than adding a second mechanism is the correct shape. A few things need fixing before it can land.

**1. The feature is a no-op under default settings.**

`compactToolDisplay` defaults to `true` (`source/config/preferences.ts:199`, `source/hooks/useAppState.tsx:180`), and in `displayExecutedTool` the compact branch is evaluated before the `execute_bash` branch (`source/hooks/chat-handler/conversation/tool-executor.tsx:137`). `execute_bash` isn't in `ALWAYS_EXPANDED_TOOLS` (`source/utils/tool-result-display.tsx:15`), so on a default install agent bash calls fold into the compact tally and the `BashProgress` card this PR modifies is never rendered. Setting `"bashOutputMode": "both"` does nothing until the user also sets `"compactToolDisplay": false`.

Best fix is to treat an explicit `tool`/`both` as intent to expand (exempt `execute_bash` from the compact branch in that case). At minimum the dependency needs documenting, otherwise users will be debugging a silent setting.

**2. Unvalidated value silently degrades to `none`.**

`getBashOutputMode` returns `preferences.bashOutputMode ?? 'user'` with no runtime check, and `loadPreferences` casts parsed JSON straight to `UserPreferences`. A hand-edited typo like `"all"` fails every comparison at both call sites and silently disables output for user `!commands` too, which is a regression from current behaviour. Since editing the file by hand is the only documented way to set this, please validate against the allowed values and fall back to `'user'`.

**3. The changeset describes the wrong change.**

`.changeset/bash-output-mode.md` describes the pre-existing behaviour from #714 and never mentions `bashOutputMode`. As written the changelog entry will describe a change that already shipped. It also has a trailing-whitespace line and no newline at EOF.

**4. `loadPreferences()` on every bash completion.**

It re-reads and re-parses the file on each call, and on a missing/unreadable file it calls `logError`, which pushes a rendered error message into the chat queue (`source/utils/message-queue.tsx:245`). A user without a preferences file would get a "Failed to load preferences" message after every bash command. Read the mode once per session or thread it through app state like `compactToolDisplayRef`.

**Smaller points:**

- `updateBashOutputMode` has no production caller, only the spec. Either wire a "Bash Output" entry into `/settings` (`source/app/components/settings-selector.tsx`), consistent with paste threshold and notifications, or drop the setter.
- The union is spelled out three times. Export a `BashOutputMode` type from `source/types/config.ts` and reuse it, as with `TitleShape` / `NanocoderShape`.
- The resolution expression is duplicated at both call sites. A single `shouldShowBashOutput(source: 'user' | 'tool')` helper next to the accessor keeps the two from drifting and is easier to test.
- Stale comments in `source/components/bash-progress.tsx:18` and `:96` still say output is only for user-typed `!commands`.

**Tests:** the four new tests only cover the accessors. Nothing verifies the actual behaviour change at either call site, which is where issue 1 would have surfaced. Please add coverage for `displayExecutedTool` and `handleBashCommand` across the four modes, plus the invalid-value fallback.

coralsundy:
Take a look sometime this week.

coralsundy:
Hi @will-lamerton this is ready to review now.

I re-design the approach based on your feedback, and record in this issue here (please assign me): https://github.com/Nano-Collective/nanocoder/issues/1293

* `showAgentBashOutput` now becomes a boolean parameter, can be controlled by the /settings, saved in cache, with proper invalidation
* Also add `cachedPreference` to be re-use by other settings if needed
* The changes are fairly straight forward and small compares to my initial attempt

Thanks!

akramcodez:
Thanks for the PR @coralsundy

PR #956, opened 2026-08-25
feat(export): auto-generate descriptive filenames from first user message

### LINKED ISSUE #934: [Feature] Auto-Title for /export Filenames
## Description

Automatically generate a descriptive filename when exporting a session via `/export`, rather than defaulting to a generic timestamp-based filename.

## Use Case

When users export multiple sessions, filenames like `session-1623491.md` make it impossible to tell what the file contains without opening it.

## Proposed Solution

- When `/export` is run without a specified filename, take the first 3-4 words of the user's initial prompt in that session.
- Sanitize the string (replace spaces with hyphens, remove special characters) and append the timestamp or `.md` extension (e.g., `setup-react-router.md`).

## Alternatives Considered

- Asking an LLM to generate the title. This is too slow and wastes tokens for a simple export command.

## Additional Context

- Update `source/commands/export.tsx`. 
- Use a simple regex to grab the first few alphanumeric words of `messages[0].content`.

tusharui:
## Description

Auto-generate descriptive filenames for `/export` instead of generic timestamps. Exports now use the first 4 words of the user's first message as a slug, making files instantly recognizable in the filesystem.

Closes #934

## Before

/export
Chat exported to nanocoder-chat-2026-08-25T09-54-14.353Z.md

## After

/export
Chat exported to fix-my-button-2026-08-25.md

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

## Changeset

## Changes

- Add `generateExportFilename()` utility — extracts first 4 words from user's first message, sanitizes to slug, appends date
- Add `uniqueFilename()` — handles collisions by appending `-2`, `-3` etc. (never overwrites)
- Add `isUnsafeFilename()` — blocks path traversal attacks (`../`, `/`, null bytes)
- Add `truncateAtWordBoundary()` — long slugs break at word boundaries, never mid-word
- Update `source/commands/export.tsx` to use shared utilities
- Add 32 tests across 2 spec files covering success, edge cases, and security

## Examples

| User's first message | Generated filename |
|---|---|
| `fix the login bug` | `fix-the-login-bug-2026-08-25.md` |
| `add dark mode toggle to the navbar` | `add-dark-mode-toggle-2026-08-25.md` |
| `hello` | `hello-2026-08-25.md` |
| (no messages) | `nanocoder-chat-2026-08-25.md` |

### Automated Tests

pnpm run test:types          — passes
pnpm run test:lint           — passes
pnpm run test:ava source/utils/generate-export-filename.spec.ts  — 16/16 pass
pnpm run test:ava source/commands/export.spec.tsx                — 16/16 pass

### Manual Testing

- [x] Tested with Ollama
- [ ] Tested with OpenRouter

## Checklist

- [x] Self-review completed

### DIFF
--- a/.changeset/fiery-hats-pick.md
+++ b/.changeset/fiery-hats-pick.md
@@ -1,0 +1,5 @@
+---
+"@nanocollective/nanocoder": patch
+---
+
+Auto-generate descriptive filenames for /export instead of generic timestamps. Closes #934
--- a/source/commands/export.spec.tsx
+++ b/source/commands/export.spec.tsx
@@ -203,12 +203,17 @@ test('exportCommand renders Export component with correct filename', async t => 
 });
 
 test('exportCommand rejects path traversal in filename', async t => {
-	const result = await exportCommand.handler(
+	const result = (await exportCommand.handler(
 		['../../../etc/passwd'],
 		testMessages,
 		testMetadata,
-	);
+	)) as React.ReactElement;
 
 	t.is(mockWriteFileCalls.length, 0);
-	t.truthy(React.isValidElement(result));
+
+	const {lastFrame} = render(<MockThemeProvider>{result}</MockThemeProvider>);
+	const output = lastFrame();
+	t.truthy(output);
+	t.regex(output!, /Invalid filename/);
+	t.false(output!.includes('Chat exported'));
 });
--- a/source/commands/export.spec.tsx
+++ b/source/commands/export.spec.tsx
@@ -245,3 +245,52 @@ test('exportCommand rejects path traversal in filename', async t => {
 	t.regex(output!, /Invalid filename/);
 	t.false(output!.includes('Chat exported'));
 });
+
+	test('exportCommand allows exporting into a subdirectory', async t => {
+	// isValidFilePath is segment-aware, so a subdirectory export must work while
+	// traversal is still blocked.
+	await exportCommand.handler(
+		['reports/chat.md'],
+		testMessages,
+		testMetadata,
+	);
+
+	t.is(mockWriteFileCalls.length, 1);
+	t.true(mockWriteFileCalls[0].path.endsWith('chat.md'));
+	t.true(
+		mockWriteFileCalls[0].path
+			.split(/[\\/]/)
+			.slice(-2)
+			.join('/') === 'reports/chat.md',
+	);
+});
+
+test('exportCommand rejects a filename with a null byte', async t => {
+	const result = (await exportCommand.handler(
+		['evil\u0000.md'],
+		testMessages,
+		testMetadata,
+	)) as React.ReactElement;
+
+	t.is(mockWriteFileCalls.length, 0);
+
+	const {lastFrame} = render(<MockThemeProvider>{result}</MockThemeProvider>);
+	const output = lastFrame();
+	t.truthy(output);
+	t.regex(output!, /Invalid filename/);
+});
+
+test('exportCommand rejects a home-directory shorthand path', async t => {
+	const result = (await exportCommand.handler(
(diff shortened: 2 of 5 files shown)


tusharui:
Hey @akramcodez can you review this and approve this pr and how many days it takes to do merge pr

will-lamerton (changes requested):
Nice feature, and the test coverage is a good start. A few things to address before this can go in.

**1. `isUnsafeFilename` duplicates (and weakens) an existing validator**

We already have `isValidFilePath(filePath, containmentRoot)` in `source/utils/path-validation.ts`, used by `read_file` / `write_file` / `string_replace`. It handles segment-wise `..`, null bytes, `~`, POSIX-masquerading Windows drive paths, and real containment via `realpathSync`. Please reuse it with `process.cwd()` as the root rather than adding a second implementation.

This also matters behaviourally: `basename(f) === f` rejects any path with a separator, so `/export reports/chat.md` and `/export ~/chat.md` used to work and now hard-fail. `isValidFilePath` keeps subdirectory exports working while still blocking traversal.

**2. The rejection path renders as a success message**

`export.tsx` returns `Export` with `filename: 'Error: invalid filename'`, which goes through `SuccessMessage` and renders as a green `Chat exported to Error: invalid filename` even though nothing was written. Use `ErrorMessage` from `source/components/message-box.tsx`. The new spec only asserts `React.isValidElement(result)`, so it passes either way; please assert on the rendered text instead.

**3. `uniqueFilename` can still overwrite**

After 998 attempts the loop returns the original `filepath`, which `fs.writeFile` then clobbers. That contradicts "never overwrites" in the description. Throw (or surface an error) instead of falling through. A random/timestamp suffix after a few tries is also cheaper than up to ~999 sequential `access` calls.

**4. Explicit filenames now get auto-renamed**

`/export notes.md` twice previously produced one file, now it produces `notes.md` and `notes-2.md`. Auto-suffixing makes sense for generated names, but for a name the user typed, overwrite is the less surprising behaviour. Happy to be argued out of this one, but it should at least be intentional and documented.

**5. Non-ASCII first messages always fall back**

`replace(/[^\w\s-]/g, '')` strips CJK, Cyrillic, accented Latin and emoji, so a first message of `修复登录问题` yields an empty slug and falls back to `nanocoder-chat-<date>.md`. Since the time component is gone, those users now collide on every export and get `-2`, `-3`, `-4` where they previously had unique timestamps. `\p{L}\p{N}` with the `u` flag would preserve most of this.

**Smaller points**

- No test covers the >998-collision fallback (the branch that can lose data) or `isUnsafeFilename` with a null byte, despite the description calling the latter out.
- `filename.split('-20')[0]!` in the truncation test is brittle: a message containing `-20` (e.g. `fix 2024-2025 rollover`) splits in the wrong place. Prefer a regex capture against the trailing `-\d{4}-\d{2}-\d{2}\.md`.
- `export {isUnsafeFilename, uniqueFilename};` sits mid-file between declarations while `generateExportFilename` uses an inline `export`. Please pick one, inline matches the rest of `source/utils/`.
- `uniqueFilename` is a generic path helper with nothing export-specific about it, so `generate-export-filename.ts` is a hard place to find it. Consider `path-validation.ts` or its own module.
- The date comes from `toISOString()`, so it is UTC rather than local. Not a regression, just worth being deliberate about.

One thing to flag rather than change: export filenames now carry the first four words of the prompt, so `/export` leaks prompt content into directory listings, shell history and anything that indexes filenames. Almost certainly fine for this feature, but it is a real change in what the filesystem reveals.

will-lamerton:
Nice work @tusharui :)

PR #958, opened 2026-08-25
feat(ui): add session auto-save indicator (fixes #932)

### LINKED ISSUE #932: [Feature] Session Auto-Save Indicator
## Description

Add a subtle visual indicator in the corner of the CLI that briefly flashes when the session state is autosaved to disk.

## Use Case

Since Nanocoder is local-first and maintains continuous semantic memory, users want peace of mind knowing that their complex agent interactions and context are safely persisted to disk. A small visual confirmation builds trust in the auto-save mechanism without interrupting the chat flow.

## Proposed Solution

- Add a small `isSaving` boolean state to the global application context.
- When the internal `saveSession` or `updateHistory` function triggers, set `isSaving` to true.
- In the main layout component (e.g., the bottom status bar or next to the token counter), conditionally render a subtle or `[Saving...]` text.
- Revert the state back to false after the write operation completes.

## Alternatives Considered

- Printing a "Session Saved" log message in the chat feed (too noisy and clutters the UI).

## Additional Context

- Ink.js renders very fast, so if the disk write is nearly instantaneous, the indicator might flicker too fast to see. Consider wrapping the state revert in a `setTimeout(..., 500)` to ensure it's visible for at least half a second.

rishu685:
## Summary

Adds a subtle visual `[Saving...]` indicator in the CLI status bar that briefly flashes whenever session state is autosaved to disk (fixing #932).

## Description & Motivation

Since Nanocoder is local-first and maintains continuous session history, users need peace of mind knowing their conversation state has been safely persisted to disk. This PR adds visual feedback in the CLI footer line alongside mode, tune, and context window metrics.

To avoid visual flickering on fast SSD disk writes (<5ms), a minimum 500ms display threshold is enforced before hiding the indicator.

## Key Changes

- **Hook State (`source/hooks/useSessionAutosave.ts`)**: Exposed `isSaving` state lifecycle with a 500ms minimum display duration threshold.
- **Component Plumbing (`App.tsx`, `interactive-app.tsx`, `chat-input.tsx`, `user-input.tsx`)**: Threaded `isSaving` prop down to the status bar component.
- **UI Indicator (`source/components/development-mode-indicator.tsx`)**: Rendered `[Saving...]` in green italic text and updated the responsive width budget calculation.
- **Backward-Compatible Git Utils Tests**: Made `git init` test setups in spec files compatible with older Git CLI versions.

## Verification & Testing

-  **Development Mode Indicator Tests**: `pnpm run test:ava source/components/development-mode-indicator.spec.tsx` (31/31 passed)
-  **Session Autosave Tests**: `pnpm run test:ava source/hooks/useSessionAutosave.spec.ts` (7/7 passed)
-  **TypeScript Build**: `pnpm run build` (0 compilation errors)
- **Manual Testing**: Launched interactive CLI and verified `[Saving...]` indicator flashes cleanly in the status bar on autosave.

Fixes #932

### DIFF
--- a/.changeset/session-autosave-indicator.md
+++ b/.changeset/session-autosave-indicator.md
@@ -1,0 +1,5 @@
+---
+"@nanocollective/nanocoder": patch
+---
+
+Added a visual `saving` indicator in the CLI status line that briefly displays whenever session state is autosaved to disk. Thanks to @rishu685. Closes #932.
--- a/source/app/App.tsx
+++ b/source/app/App.tsx
@@ -592,7 +592,7 @@ export default function App({
 	});
 
 	// Setup session autosave
-	useSessionAutosave({
+	const {isSaving} = useSessionAutosave({
 		messages: appState.messages,
 		currentProvider: appState.currentProvider,
 		currentModel: appState.currentModel,
@@ -784,6 +784,7 @@ export default function App({
 							handleUserSubmit={handleUserSubmit}
 							userMessageQueue={userMessageQueue}
 							handleIdeSelect={handleIdeSelect}
+							isSaving={isSaving}
 						/>
 					)}
 				</PrivacyContext.Provider>
--- a/source/app/components/chat-input.tsx
+++ b/source/app/components/chat-input.tsx
@@ -101,6 +101,7 @@ export interface ChatInputProps {
 	 * transcript and isn't clipped by the scroll viewport's overflow="hidden".
 	 */
 	fullscreen?: boolean;
+	isSaving?: boolean;
 }
 
 /**
@@ -155,6 +156,7 @@ export function ChatInput({
 	activeEditor,
 	onDismissActiveEditor,
 	fullscreen = false,
+	isSaving,
 }: ChatInputProps): React.ReactElement {
 	const {colors} = useTheme();
 	const activeToolCall = pendingToolCalls[currentToolIndex];
@@ -253,6 +255,7 @@ export function ChatInput({
 					currentModel={currentModel}
 					activeEditor={activeEditor}
 					onDismissActiveEditor={onDismissActiveEditor}
+					isSaving={isSaving}
 				/>
 			) : /* Client Missing */
 			mcpInitialized && !client ? (
--- a/source/app/sections/interactive-app.tsx
+++ b/source/app/sections/interactive-app.tsx
@@ -50,6 +50,7 @@ interface InteractiveAppProps {
 	 * the inline Static-based flow with native scrollback.
 	 */
 	altScreenActive?: boolean;
+	isSaving?: boolean;
 }
 
 /**
@@ -76,6 +77,7 @@ export function InteractiveApp({
 	handleIdeSelect,
 	clearKey,
 	altScreenActive = false,
+	isSaving,
 }: InteractiveAppProps): React.ReactElement {
 	const nextRestoredDraftIdRef = React.useRef(1);
 	// Tune / IDE are launched by closing settings first, so their exit has no way
@@ -445,6 +447,7 @@ export function InteractiveApp({
 								tune={appState.tune}
 								currentModel={appState.currentModel}
 								fullscreen={fullscreen}
+								isSaving={isSaving}
 							/>
 						</UIStateProvider>
 					)}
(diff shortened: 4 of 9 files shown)


will-lamerton (changes requested):
Thanks for this, the feature is worth having and the plumbing follows the existing pattern for `tune` / `activeEditor`. A few things need addressing before it can go in.

**Blocking**

1. **Needs a rebase.** The branch is 53 commits behind `main` and conflicts in every file it touches. This isn't mechanical: `development-mode-indicator.tsx` gained the `taskInfo` badge with its own drop-order logic, and `useSessionAutosave.ts` gained the internal-walkthrough message filter. The width budget and the try/finally placement both need re-deriving against current `main`.

2. **The 500ms floor is awaited inside `runSave`, so it delays process exit.** `runSave` is what `flush()` chains onto `saveChainRef` and awaits, and `flush` is registered as a shutdown handler at priority -10. Every quit now blocks for up to an extra 500ms on a purely cosmetic delay, at a point where the UI is already torn down and nobody can see the badge. It also holds the serialised save chain open, so a queued save waits behind the animation rather than behind the write. Make the hide presentational instead: record the start time, and on completion `setTimeout(() => setIsSaving(false), remaining)` without awaiting it.

3. **Nothing cancels the pending timer.** The inline `await new Promise(setTimeout)` isn't cancellable, so at exit it leaves a live handle on the event loop and calls `setIsSaving(false)` on an unmounted tree. The existing cleanup effect only clears `timeoutRef`. Moving to a `hideTimerRef` cleared in that same cleanup fixes this and (2) together.

4. **The saving segment steals width from the session and editor labels.** `savingSegment.length` is added to both `requiredWidth` and `fixedWidth`, and `remaining` (which feeds `sessionMax` / `filenameMax`) is derived from `fixedWidth`. So on a near-full terminal the session name and filename lose 14 columns and re-truncate for the duration of the flash, then restore, every 30 seconds. That's a bigger visual jump than the sub-frame flicker the 500ms floor exists to prevent. It should be an *optional* segment that drops when there's no room, joining `taskHintExtra` / `editorSuffix` / `shiftHint` in the existing drop order, rather than a required one that displaces user content.

**Should fix**

- `await initPromiseRef.current` moved out of the `try`. Harmless today since `initPromiseRef` is built with a `.catch` returning `false`, but it removes the safety net: `flush()` awaits the chain directly, so a future rejection would surface inside a shutdown handler. Keeping the await inside the `try` and calling `setIsSaving(true)` after the early returns costs nothing.
- Post-rebase, the `persistedMessages.length === 0` early return sits inside the `try`, so the badge will flash for 500ms on saves that write nothing.
- **Please split out the four `git init` rewrites.** They're unrelated to this PR, and `git init -q -b main` is on `main` with CI green, so there's no failure being fixed. The replacement is also weaker: `catch {}` swallows genuine failures, and when it does the test silently continues on `master` and fails later with a confusing message. If the compatibility concern is real, `git init -q` followed by `git symbolic-ref HEAD refs/heads/main` is deterministic on every git version and needs no try/catch.

**Tests**

The two added assertions (badge present / absent) don't cover the behaviour the PR is actually about. Worth adding: a hook-level test for the 500ms minimum-display floor, a test that the badge doesn't extend shutdown flush time, and a width-budget test alongside the existing drop-order tests. Note `ink-testing-library` renders at a fixed 100 columns, so a width test has to drive `useResponsiveTerminal` rather than set `process.stdout.columns`.

**Nits**

- `[Saving...]` doesn't match the surrounding style, which is bare and lowercase-keyed (`tune: full`, `ctx: 42%`). Something like `saving` is more consistent and shorter, which helps the width issue too.
- `colors.success` (green) reads as a completed state. The other transient segments use `colors.info` / `colors.secondary`.

**Verified as fine**

With `sessions.autoSave: false` the guard returns before the badge ever shows; saves are serialised through `saveChainRef` so there's no overlapping-flag desync; and the prop chain covers every render path (`--plain` and the VS Code / ACP surfaces have their own status rendering and are correctly untouched).

will-lamerton:
Nice job @rishu685 :)

will-lamerton:
Thanks for this PR @rishu685 - feel free to add yourself as a contributor to our website via a PR which I will approve :)

PR #959, opened 2026-08-25
Feature: theme aware syntax highlighting

### LINKED ISSUE #935: [Feature] Syntax Highlighting Theme Customization
## Description

Allow users to define a custom syntax highlighting theme for the CLI markdown renderer via the configuration file.

## Use Case

Developers are highly opinionated about their colors. If their terminal uses a specific color palette (like Dracula or Nord), the default syntax highlighting in Nanocoder might clash or be hard to read.

## Proposed Solution

- Add a `syntaxTheme` key to the config file (e.g., `"syntaxTheme": "dracula"`).
- Pass this theme string to the underlying markdown/syntax-highlighter package used by Ink.js when rendering code blocks.

## Alternatives Considered

- Auto-detecting the terminal theme. This is notoriously difficult and unreliable across different OS terminals.

## Additional Context

- Check which highlighter the project uses (e.g., `cli-highlight`, `shiki`, or chalk-based) and map the config string to the corresponding supported themes.

addyCooks:
Closes #935.

## Description

Syntax highlighting now follows your theme, and takes a `syntaxTheme` preference when you want code to keep a palette of its own.

Every `cli-highlight` call site passed `theme: 'default'`, a string where the library expects a token-to-formatter map. The option was silently discarded, so code always rendered in cli-highlight's own palette no matter which of the 50 themes was selected half feature request, half dead code.

`getSyntaxTheme()` now builds that map from a theme's palette: keywords take `primary`, built-ins and declarations `tool`, strings `success`, numbers `warning`, comments `secondary`, attributes and variables `info`, and everything else the theme's body `text`. It covers every token cli-highlight styles itself, so nothing falls back to the library default, and it's memoised per palette because the diff and file previews highlight line by line.

Applied at all five sites: markdown code blocks, both `string_replace` diff-context branches, the `write_file` preview, and the file explorer preview.

Code follows `selectedTheme` by default. Setting `syntaxTheme` in `nanocoder-preferences.json` points code at any other theme's palette while the rest of the UI stays put:

```json
{"selectedTheme": "tokyo-night", "syntaxTheme": "dracula"}
```

It names one of the 50 existing themes rather than introducing a second registry, so the two cannot drift apart. An unknown or misspelt name falls back to `selectedTheme` rather than dropping the styling.

`@/config/preferences` imports `@/config/index`, which imports `@/config/themes`. Neither touches the other at module scope, so the cycle resolves verified against the built output for the three entry orders that reach it.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Documentation update

## Changeset

### Automated Tests

### Manual Testing

- [ ] Tested with Ollama
- [ ] Tested with OpenRouter

## Checklist

- [x] Self-review completed

## Note on the proposed solution

The issue suggests mapping the config string to "the corresponding supported themes" of the highlighter. cli-highlight has no named themes to map onto its `theme` option takes a token-to-formatter map and nothing else so the string resolves against nanocoder's own 50 themes instead. That keeps Dracula and Nord (the palettes named in the issue) available without a parallel registry to maintain.


will-lamerton (changes requested):
Nice catch on the dead `theme: 'default'` option, and reusing the existing 50 presets instead of a parallel highlighter registry is the right call. I verified the cycle resolves for all three entry orders, all five call sites are converted, and the palette objects are identity-stable so the `WeakMap` actually hits. `tsc`, `biome`, and the touched specs (plus `write-file` / `string-replace` / `file-explorer` with a `syntaxTheme` set) are all clean.

Two things I'd like changed before merge, both in `resolveSyntaxPalette`:

**1. The cache only invalidates on something that never changes in production.**

`overrideCache` re-resolves when `NANOCODER_CONFIG_DIR` changes, which is a test-only signal. In a real session that env var is constant, so `syntaxTheme` is read once per process and any later preferences write is ignored. `getPreferencesVersion()` in `source/config/preferences.ts` already exists for exactly this: a monotonic counter bumped on every `savePreferences`, documented as free to read with no disk I/O. Keying on both makes the cache correct at runtime and keeps the test path working:

```ts
const dir = process.env.NANOCODER_CONFIG_DIR;
const version = getPreferencesVersion();
if (!overrideCache || overrideCache.dir !== dir || overrideCache.version !== version) {
```

**2. `preset in themes` should be `Object.hasOwn(themes, preset)`.**

`themes` comes from `JSON.parse`, so it inherits `Object.prototype`. `"syntaxTheme": "constructor"` (or `toString`, `valueOf`) passes the `in` check, `themes[preset].colors` is then `undefined`, and the `?? colors` fallback happens to rescue it. It works by accident, and the `Record<ThemePreset, Theme>` type hides the hole from tsc.

Smaller points, happy for any of these to be follow-ups:

- The unknown-theme fallback is completely silent, and there's no `/settings` surface for `syntaxTheme`, so someone who types `draclua` gets the pre-PR behaviour with no way to tell why. A one-time warn naming the unknown value would help.
- The file explorer highlights once and stores the ANSI in state, so changing theme with it open leaves the preview in the old palette until reselect.
- In `markdown-parser/index.spec.ts`, the comment about needing "a distinct palette object, so the memoised syntax theme for it is built while colour is forced on" isn't right: chalk builders resolve `chalk.level` at call time, not creation time. The `chalk.level = 3` save/restore is still needed, the distinct-object trick isn't. Worth fixing so the next reader doesn't preserve a constraint that doesn't exist.
- `Colors` from `@/types/markdown-parser` is the right type to reuse, but the name now under-describes it. Not blocking.

addyCooks:
Thanks @will-lamerton, 
both blocking items are done.

**1.** Cache now keys on `getPreferencesVersion()` as well as the dir, so a mid-session write lands on the next highlight. This needed a `main` merge first (branch was 101 behind, which is where `getPreferencesVersion` lives). Added a test that fails without the version key.

**2.** `Object.hasOwn` in place, with tests for `constructor` / `toString` / `valueOf` / `__proto__`.

Also did the three follow-ups: one-time warn naming the unknown value, file explorer keeps the plain source and derives the highlight in a memo keyed on the palette, and `RenderPalette` names the shared subset at its declaration (`Colors` kept as an alias so the call sites are untouched).

Two notes:

**The chalk one you're half right, and it changed the fix.** Level *is* re-checked per call for whether to emit, but the colour *model* is baked when the builder is created. On 4.1.2:

built at level=0, called at level=3 : "\u001b[91m…"          ← ansi16
built at level=3, called at level=3 : "\u001b[38;2;255;0;0m…" ← truecolor


`getSyntaxTheme` memoises per palette identity, so reusing `mockColors` reuses builders frozen at the runner's default level and the assertion compares ansi16 against truecolor. I removed the distinct object and the test failed. Kept it, rewrote the comment to say what's actually going on  same goal, opposite direction.

**Used the structured logger, not `logWarning`.** `@/utils/message-queue` reaches `message-box` → `useTheme` → back to this module, and `getSyntaxTheme` runs from render, so queueing there is a state update during another component's render. Reasoning is in a comment. Downside is it lands in the log rather than the TUI the `/settings` surface you mentioned is the real fix for visibility, happy to take that as a follow-up.

will-lamerton:
Nice work again @addyCooks :)

addyCooks:
Thankyou so much @will-lamerton !! : )

PR #961, opened 2026-08-26
feature(voice): Native Realtime cloud audio (PR 6/6)

RONAK-AI647:
## What this does

This is the sixth and final PR for Realtime Voice Mode (see #622). It adds two 
optional extras on top of the now-complete local-first voice pipeline from 
PR1-5. Neither changes anything for someone who doesn't touch the new settings -local voice keeps working exactly as before.

1. **Optional cloud speech, if you want it.** `/voice stt cloud` and 
   `/voice tts cloud` let you use a cloud provider (if your current AI provider 
   supports it) instead of the local tools. Falls back to local automatically 
   if cloud fails for any reason. Still local by default - nothing changes 
   unless you explicitly ask for cloud.
2. **Groundwork for future real-time providers.** Some AI providers may 
   eventually offer true instant two-way voice (no separate record → transcribe → 
   reply → speak steps). This PR adds the plumbing to detect and use that when 
   it exists, without building against any provider that doesn't actually 
   support it yet.

## Also added
- `/voice status` - see your current voice settings at a glance.

### DIFF
--- a/docs/features/commands.md
+++ b/docs/features/commands.md
@@ -41,10 +41,22 @@ Type `/` in the chat input to see available commands. All commands start with `/
 | `/retry` | Re-run the last user turn. Use `/retry --model <id>` or `/retry --provider <name> --model <id>` to switch models first |
 | `/rename` | Rename the current session. Name must be non-empty and 100 characters or less. See [Session Management](session-management.md) |
 | `/explorer` | Interactive file browser to navigate, preview, and select files for context |
-| `/tune` | Configure runtime model behaviour — tool profiles, compaction, native tools, model parameters (see [Tune](tune.md)) |
 | `/ide` | Connect to an IDE for live integration (e.g., VS Code diff previews) |
+| `/voice` | Configure Voice Mode (`/voice [hands-free|ptt|stt <local|cloud>|tts <local|cloud>|status|mode <push-to-talk|hands-free>]`) |
 
-## Special Input Syntax
+## Voice Mode
+
+Nanocoder includes a local-first Realtime Voice Mode supporting push-to-talk, hands-free voice activity detection (VAD), barge-in interruption, and optional cloud STT/TTS.
+
+| Command | Description |
+|---------|-------------|
+| `/voice` | Toggle voice mode on/off |
+| `/voice ptt` or `/voice push-to-talk` | Switch to push-to-talk mode (use `Ctrl+T` to record/submit) |
+| `/voice hands-free` | Switch to hands-free mode (automatic VAD speech detection) |
+| `/voice stt [local\|cloud]` | Configure speech-to-text backend (local Whisper or opt-in cloud) |
+| `/voice tts [local\|cloud]` | Configure text-to-speech backend (local Piper or opt-in cloud) |
+| `/voice status` | Display current voice configuration and backend status |
+| `/voice mode <push-to-talk\|hands-free>` | Set specific activation mode |
 
 These shortcuts work directly in the chat input — no `/` prefix needed.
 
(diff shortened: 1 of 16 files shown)


RONAK-AI647:
I tested all features , working well ! I will paste the demo recording as well . @will-lamerton 
Do take a look !
And a guide for using for using the voice mode for MAC and windows user in the next PR!

akramcodez (commented):
Hey @RONAK-AI647, thanks for the work on this. I went through the implementation carefully, including the realtime plumbing, cloud STT/TTS flow, voice commands, cancellation changes, and the tests.

Overall, the implementation is in a good shape and the cloud fallback path looks solid. I especially like that cloud STT/TTS remains opt-in and falls back to local execution, and the tool cancellation changes are nicely isolated. 

I did find a few things that I'd like addressed before merging:

### 1. Realtime capability is currently unused

The `RealtimeCapability` types are added, but there is currently no production consumer for `openRealtimeSession`, `sendAudioChunk`, or `sendTextMessage`. The only references outside the type definition are in tests. `isRealtimeCapable` and `activeRealtimeSession` therefore don't currently provide any real functionality, and the provider-switch cleanup can never actually close a session opened by this implementation.  

I think we should either:

* actually wire the capability into `useVoice`, or
* keep this as future groundwork but remove the unused public return values until there is a real consumer.

I would prefer the second option if native realtime isn't being implemented yet. That keeps the current implementation honest and avoids exposing an API that isn't actually functional.

### 2. The SIGKILL test doesn't actually verify the SIGKILL fallback

The new test starts a process that ignores `SIGTERM`, which is good, but it awaits `cancel()` immediately and checks `isComplete`. `cancel()` can resolve through its existing cancellation bookkeeping before the 2-second SIGKILL fallback has even executed.

So the test can pass without proving that the SIGKILL fallback actually killed the process. 

I'd change the test so that it verifies the actual invariant:

* wait until the SIGKILL timeout has elapsed and
* verify the process is actually dead,

or alternatively capture/verify the signal received by the child process.

### 3. `/voice mode <unknown>` silently falls back

An invalid command such as:

```text
/voice mode foobar
```

currently falls back to push-to-talk instead of reporting that the mode is invalid. 

I'd recommend returning an error/help message listing the valid modes rather than silently changing the user's configuration.

### 4. `/voice ptt` doesn't enable voice mode

The `push-to-talk` branch changes `activationMode`, but unlike the hands-free branch, it doesn't set `updatedConfig.enabled = true`.

That means `/voice ptt` can report that voice mode has been configured while voice itself remains disabled. 

This should be made consistent with the hands-free path.

### 5. The no-loop test isn't actually testing the loop condition

The test that checks that declining the voice dependency installation doesn't repeatedly trigger the prompt rerenders the component several times, but none of the effect dependencies change between those renders.

As a result, the effect doesn't actually execute again, so the test can pass without proving that the loop is fixed. 

I'd make the test actually retrigger the effect, for example:

```text
hands-free
    ↓
push-to-talk
    ↓
hands-free
```

and then assert that the installation prompt is still only triggered once.

### 6. User-facing voice documentation is still missing

The feature now exposes `/voice`, cloud STT/TTS configuration, and the Ctrl+T push-to-talk interaction, but the existing command and keyboard-shortcut documentation doesn't contain these. 

Since this is now user-facing functionality, I'd like the relevant docs updated before we consider the feature complete.

### Smaller cleanup

A few non-blocking things I noticed:

* `format-for-speech.spec.ts` appears to have a formatting issue that should be caught by the formatter. 
* `images?: ImageAttachment[]` was added to `handleUserSubmit`, but the voice call sites don't currently use it. 
* `transcribeCloudAudio` could use explicit AbortSignal tests, especially since interruption is an important part of voice mode. 
* The `RealtimeCapability` guard has a redundant `'supportsRealtimeAudio' in client` check. 
* There is a small potential race around the voice installation prompt if the preference is toggled while the prompt is being declined. It's unlikely, but worth keeping in mind. 

**Once these are addressed, I'm happy to take another look.**

RONAK-AI647:
@akramcodez , PTAL !

akramcodez:
Over time, such great work @RONAK-AI647, one more request could you please post a demo video showing how it works?

RONAK-AI647:
> Over time, such great work @RONAK-AI647, one more request could you please post a demo video showing how it works?

Sure Sure !! I will post it soon !

PR #963, opened 2026-08-26
feat(stats): add /stats lifetime usage command (#933)

### LINKED ISSUE #933: [Feature] `/stats` Command (Lifetime Usage)
## Description

Add a `/stats` command that displays lifetime usage metrics, such as total prompts sent, total tokens processed, and total sessions created.

## Use Case

Gamification and curiosity. Users enjoy seeing how much they use a tool. It also gives them a sense of how many tokens they've burned through over the lifetime of their installation.

## Proposed Solution

- Create a hidden `.nanocoder-stats.json` file in the user's config directory.
- Increment counters for `totalPrompts`, `totalTokens`, and `totalSessions` as the application runs.
- Add a `/stats` command to read and pretty-print this JSON file in the CLI.

## Alternatives Considered

- Displaying stats on every exit (too spammy). A dedicated slash command is unobtrusive.

## Additional Context

- Ensure the stats file is updated asynchronously so it doesn't block the main rendering thread or slow down chat responses.

puri-adityakumar:
## Description

Closes #933.

Adds a `/stats` slash command that shows **lifetime** usage metrics (sessions, prompts, tokens, estimated cost), separate from `/usage` which remains the current-context view.

Persists a local ledger at app-data `.nanocoder-stats.json`, increments as the app runs (session create, user prompts, API usage — including subagents and plain/headless), and renders an Ink card with:

- Ranges **7d / 3m / all-time** (`←`/`→` to switch, matching Settings tabs; Esc/Enter to close)
- Peak day, streak, since (all-time)
- Vertical cumulative bar chart
- Top 5 `(provider, model)` pairs by tokens

### update

- Added provider-reported token accounting for subagent calls.
- Added cache-aware pricing and partial-usage handling.
- Fixed all-time totals, safe concurrent writes, reset handling, and legacy filename migration.
- Resolved the latest upstream merge conflicts and verified the feature checks.

### `/stats` — 7d
image

### `/stats` — all-time
image

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

## Changeset

### Automated Tests

  - Note: format/types/lint green; stats specs 36/36 passing. Full `test:all` also reported 9 pre-existing failures in `source/tools/git/utils.spec.ts` (husky rejects fixture `-m init`), unrelated to this change.

### Manual Testing

- [x] Exercised `/stats` UI locally (ranges, ←/→, chart, top providers)
- [ ] Tested with Ollama — not available on my machine
- [x] Tested with OpenRouter
- [ ] Tested MCP integration (if applicable) — N/A for this change

## Checklist

- [x] Self-review completed
- [ ] Appropriate logging added using structured logging (see [CONTRIBUTING.md](../CONTRIBUTING.md#logging)) — storage failures use existing `logWarning`; no new Pino logger module


will-lamerton (changes requested):
Thanks for this, the aggregation and chart code is genuinely nice work and well covered by tests. A few things need fixing before I can merge.

## Blocking

**1. The composer-hiding guard also hides the input during `!bash` commands** (`source/app/sections/interactive-app.tsx`)

`setIsToolExecuting(true)` has only three call sites today, all in `app-util.ts`: `handleBashCommand`, `handleCopilotLogin`, `handleCodexLogin`. Model-driven tool execution runs with `isGenerating`, not `isToolExecuting` (see the comment at `App.tsx:329`). So `liveComponent && isToolExecuting && pendingToolCalls.length === 0` never fires for the tool case the comment says it protects, and it does fire for a user-typed `!npm run build`: live `BashProgress`, `isToolExecuting` true, no pending tool calls. The input box and message queuing vanish for the whole command.

Please drive this off an explicit flag instead of inferring it, e.g. `handleStatsCommand` sets something like `setLiveComponentCapturesInput(true)` in `useAppState` and the composer gates on that.

**2. Multi-process writes clobber each other** (`source/stats/record.ts`)

`getLedger()` reads the file once per process and caches it for the process lifetime, and every flush writes the whole cached object back. Two nanocoder sessions in two terminals means whichever flushes last overwrites everything the other recorded, not just a narrow race. `source/usage/storage.ts` avoids the worst of this by doing read-modify-write per call. Simplest fix is to re-read and apply deltas inside `flushStatsLedgerSync`.

## Should fix

- **Subagent tokens are not comparable to the rest of the ledger.** `finalTokenCount` comes from `subagentProgress.tokenCount`, which increments once per streamed chunk in `onToken` / `onReasoningToken`. That is output-only stream chunks, not provider-reported usage, so it lands in the same `totalTokens` bucket as real usage and skews both the headline and the top-pair shares. Either record from a real usage report or leave subagents out for now.
- **Subagent runs increment `Prompts`** via `recordUserPrompt`. One user turn fanning out five agents reads as six prompts.
- **Esc double-dispatch.** `isToolExecuting` makes `cancellable` true in `interactive-app.tsx`, so Esc to close the card also runs `handleCancel()`, which cancels active bash executions and aborts the abort controller if one exists. The explicit-flag fix above can cover this too.
- **Dead code / knip.** `createEmptyDailyStats` is never called anywhere, `invalidateStatsCache` is never called, `MAX_STATS_DAILY_DAYS` is exported but only used in-file, and most of the `source/stats/index.ts` barrel is unreferenced (the specs import the modules directly). Please confirm `pnpm run test:knip` is green on the branch.
- **Two tests assert on source text** rather than behavior: `stats.spec.tsx` regexes `app-util.ts`, and `storage.spec.ts` regexes `subagent-executor.ts`. Those pass even if the wiring is dead and break on harmless refactors. Worth swapping for a real call through the handler with stub options, plus some coverage of the interactive key handling (`ink-testing-library` handles `stdin.write('[C')` for arrows).

## Nits

- All-time chart vs all-time total diverge once `pruneDaily` starts dropping days past 400, since the headline uses uncapped `totalTokens` and the chart sums `daily`.
- Est. cost drops `cacheReadTokens` / `cacheWriteTokens`, and the subagent path passes only `totalTokens`, which cannot be priced, so those land as 0. Fine given the "Est." label, just noting it will not match `/usage`.
- The static transcript snapshot still shows the "←/→ switch range" hint, which does nothing there.
- Top-pair rows overflow in narrow terminals (no `wrap="truncate-end"` unlike the chart rows).
- `STATS_LEDGER_VERSION = 2` for a format with no shipped v1 is confusing, and `normalizeLedger` ignores the stored version entirely, so there is no migration hook later.
- `.nanocoder-stats.json` is inconsistent with its neighbours in app-data (`usage.json`, `sessions/`).
- Worth a line in `docs/configuration/preferences.md` about the new file, and a follow-up for resetting the ledger (`clearStatsLedger` exists but is unreachable).
- ACP has its own loop and is not instrumented, so ACP sessions record nothing. Happy to defer that.

puri-adityakumar:
@will-lamerton  I’ve addressed your requested fixes, merged the latest upstream changes, and resolved the conflicts, could you please review and let me know if anything else needs fixing?

will-lamerton:
Thanks for this PR @puri-adityakumar - feel free to add yourself as a contributor to our website via a PR which I will approve :)

PR #965, opened 2026-08-26
fix(checkpoints): snapshot file bytes, and surface an incomplete restore

### LINKED ISSUE #962: [Bug] Checkpoint snapshots are UTF-8-only at every layer, irrecoverably corrupting binary files on save and restore
## Description

Every layer of the checkpoint pipeline reads and writes file contents as UTF-8 text. Binary files, images, icons, `.vsix` bundles, sqlite databases, fonts, any compiled artifact are not valid UTF-8, so each invalid byte is replaced with U+FFFD (`EF BF BD`). The file is destroyed, no error is raised, and nothing in the UI indicates anything went wrong.

There are four chokepoints, and the encoding is hardcoded at all of them:

| # | Location | Operation |
|---|---|---|
| 1 | `source/services/file-snapshot.ts:29` | `captureFiles` reads the working file as `'utf-8'` |
| 2 | `source/services/checkpoint-manager.ts:174` | writes the snapshot into `files/` as `'utf-8'` |
| 3 | `source/services/checkpoint-manager.ts:228` | `loadCheckpoint` reads it back as `'utf-8'` |
| 4 | `source/services/file-snapshot.ts:58` | `restoreFiles` writes over the working file as `'utf-8'` |

**The most serious consequence is at steps 1–2, not step 4.** The bytes are already lost when the checkpoint is *created*. The copy persisted under `.nanocoder/checkpoints/<name>/files/` is corrupt at rest, so the original content cannot be recovered from the checkpoint either. A repair tool written after the fact would have nothing to recover this is unrecoverable data loss at save time, and the overwrite at step 4 is merely where the user finds out.

Nothing filters binary files out of the path. `getModifiedFiles` (`source/services/file-snapshot.ts:73`) returns everything from `git diff --name-only HEAD` plus `git ls-files --others --exclude-standard`, filtered only by `.gitignore` and capped at `MAX_CHECKPOINT_FILES`. There is no text/binary check anywhere between `git` and `fs.writeFile`.

**Affected trigger paths** all three call the same service:

- `/checkpoint create` (`source/commands/checkpoint.tsx:82`) manual
- Every daemon-triggered skill run (`source/skills/dispatcher.ts:121`) automatic, unless in plan mode
- The safety backup taken immediately *before* every restore (`source/hooks/useAppHandlers.tsx:446`) automatic

That last one compounds the damage: the safety net is itself built from lossy snapshots, so restoring twice overwrites the last clean copy on disk with a doubly-corrupted one.

## Environment

- **OS**: Windows 11 platform-independent; the defect is in the encoding, not the filesystem
- **Node version**: v22.22.3
- **Provider**: N/A
- **Model**: N/A reproducible with no LLM involved; the checkpoint service is provider-agnostic

## Steps to Reproduce

1. In a git repo, commit a binary file (e.g. a PNG at `assets/logo.png`)
2. Modify it so `git diff --name-only HEAD` reports it
3. Run `/checkpoint create test`
4. Inspect `.nanocoder/checkpoints/test/files/assets/logo.png` it is **already corrupt**, before any restore has happened
5. Modify the working file again
6. Run `/checkpoint restore test`
7. The working file is now corrupt too, and the original is unrecoverable

## Expected Behavior

A checkpoint round-trip is byte-preserving. Any file that can be captured can be restored byte-identically, regardless of its contents. If a file genuinely cannot be snapshotted, that is surfaced to the user rather than silently degraded.

## Actual Behavior

Bytes are destroyed at capture, and again at restore. Output on 1.29.0:

```
original bytes : 67  89504e470d0a1a0a
restored bytes : 77  efbfbd504e470d0a
identical      : false
```

The PNG magic byte `0x89` becomes `EF BF BD` (U+FFFD), and the file grows 67 → 77 bytes as every non-UTF-8 byte is replaced. The image no longer opens in any viewer, and neither the working file nor the stored snapshot contains the original data.

## Logs

Nanocoder emits no logs for this no error occurs at any layer, so there is nothing to capture. The evidence is a byte comparison instead.

Driving the real `CheckpointManager` end to end (save → inspect stored copy → restore):

```
  file                          bytes   first 12 bytes
  ------------------------------------------------------------------
  original on disk              67      89 50 4e 47 0d 0a 1a 0a 00 00 00 0d
  stored inside the checkpoint  77      ef bf bd 50 4e 47 0d 0a 1a 0a 00 00
  restored to the workspace     77      ef bf bd 50 4e 47 0d 0a 1a 0a 00 00

  checkpoint copy matches original : false
  restored file matches original   : false
  valid PNG signature preserved    : false
```

The PNG signature byte `0x89` is replaced by `EF BF BD` (U+FFFD) and the file grows 67 → 77 bytes. Note the second row: the copy **stored inside the checkpoint** is already corrupt, before any restore has been attempted. The original bytes are gone at save time, so no recovery from the checkpoint is possible.

### Proposed spec

Per CONTRIBUTING ("No spec? Write one"):

**Fix:** stop specifying an encoding. Change `fileSnapshots` from `Map<string, string>` to `Map<string, Buffer>` and drop the `'utf-8'` argument at all four sites. Node writes a Buffer verbatim, so this needs no binary detection, no encoding marker, and no metadata changeand text files still round-trip byte-identically.

**Blast radius:** contained. Nothing reads snapshot *content* as a string `types/checkpoint.ts:29` (the type), `file-snapshot.ts`, `checkpoint-manager.ts`, and two call sites that only touch `.size`.

**Compatibility:** old checkpoints still load. Their binary entries stay corrupt, that data is already gone and text ones are unaffected.

**Alternative rejected:** base64-encoding only detected binaries needs a metadata schema change, a heuristic that can be wrong at the margins, and a migration path, to land where the Buffer approach lands with none of them.

**Tests:** the repro above, plus a `CheckpointManager` round-trip asserting the stored copy under `files/` is byte-identical that second one pins chokepoints 2 and 3, which the service-level test doesn't reach.

### Two adjacent gaps in the same path

Both are smaller, and I would be glad to split them into their own issue if you would prefer this one stay single-purpose:

- `captureFiles` skips unreadable files with only a warning (`file-snapshot.ts:32`), so a restore can silently be partial likely on Windows, where editors and antivirus hold file locks
- `getModifiedFiles` truncates at `MAX_CHECKPOINT_FILES = 50` and warns at capture time, but nothing surfaces at restore time, so the user is never told their restore was incomplete

### Note

This repository ships `assets/nanocoder-vscode.vsix`, regenerated by `pnpm run build:vscode` and modified in open PRs #887 and #826 a tracked binary changed in exactly the workflow where checkpoints run.

Happy to open a PR for this if the approach looks right.

addyCooks:
## Description

Closes #962.

Every layer of the checkpoint pipeline read and wrote file contents as UTF-8, and
nothing between git and `fs.writeFile` checked whether a file was text.
`getModifiedFiles` returns whatever `git diff --name-only HEAD` reports, so images,
`.vsix` bundles, sqlite databases and every other tracked binary went through the
same path, and each invalid byte came back as U+FFFD. No error was raised and
nothing surfaced in the UI.

The loss happened at save, not at restore. The copy written under
`.nanocoder/checkpoints/<name>/files/` was already corrupt at rest, so the original
bytes could not be recovered from the checkpoint either the overwrite on restore
is only where the user finds out. All three trigger paths were affected, the safety
backup taken before every restore included, so restoring twice replaced the last
clean copy on disk with a doubly-corrupted one.

**Encoding.** Carry snapshots as `Buffer` and drop the encoding argument at all four
chokepoints. Node writes a `Buffer` verbatim, so this needs no binary detection, no
encoding marker and no metadata change, and text still round-trips byte-identically.
Nothing downstream read a snapshot as a string: only `Map.size` is used outside
these two services.

**Incomplete restores.** A restore could also put back less than the whole workspace
and still report plain success. Three causes, none of which reached the user:

| Cause | Was | Now |
|---|---|---|
| File unreadable at capture (a lock, permissions) | `logWarning`, dropped | `skippedFiles` in metadata, reported at restore with the reason |
| `MAX_CHECKPOINT_FILES` truncation | warned at capture, invisible after | `truncatedFileCount` in metadata, reported at restore |
| Stored snapshot unreadable at load | `logWarning`, dropped | derived at restore from `filesChanged` vs what loaded |

The third is not listed in the issue `loadCheckpoint` skips an unreadable stored
snapshot the same way `captureFiles` skipped an unreadable source file, and a
missing `files/` directory drops every file at once. Same symptom, so it uses the
same mechanism. All three restore paths share one message so they cannot drift.

Both metadata fields are optional and written only when non-zero, so a clean
capture's metadata is unchanged and older checkpoints load as before.

Capture is deliberately not made fatal: a locked file is ordinary on Windows, and
hard-failing would also block the automatic pre-restore backup a worse failure
than the one being fixed.

Old checkpoints still load. Binary entries captured before this stay corrupt that
data is already gone and text ones are unaffected.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Changeset

### Automated Tests

### Manual Testing

- [ ] Tested with Ollama
- [ ] Tested with OpenRouter

## Checklist

- [x] Self-review completed


will-lamerton (changes requested):
Thanks for this - the diagnosis is right, and the point that the corruption happened at save rather than restore is the important one. Carrying snapshots as `Buffer` and dropping the encoding argument at all four chokepoints is the correct minimal fix, and the tests assert at the byte level where a regression would actually show.

Two things to sort before this can go in.

**1. It does not compile against current `main`.** GitHub reports the branch as conflicting, and the conflict is semantic rather than textual:

- `source/services/file-snapshot.ts` on main already splits this into `getModifiedFiles(): string[]` delegating to `getModifiedFilesResult(): {files, truncated, available}`, added so the action timeline can refuse to trust a truncated scan. This PR changes `getModifiedFiles`'s own signature instead. On rebase, please drop `ModifiedFiles` and read from `getModifiedFilesResult()`; the only gap there is that it reports a boolean rather than a count, which is one field to add.
- `source/services/timeline-manager.ts` is a new consumer this breaks in three places: it iterates `captureFiles(...)` directly as a Map (a `CaptureResult` is not iterable), passes the value to `isProbablyBinary(content: string)`, and passes a `Map<string, string>` to `restoreFiles`. The `isProbablyBinary` one is worth a second look rather than a mechanical retype, since timeline-manager deliberately excludes binaries.

**2. Deleted files become permanent false gap warnings.** `git diff --name-only HEAD` reports files deleted in the working tree, so they reach `captureFiles`, fail with `ENOENT`, and are now persisted to `metadata.skippedFiles` and shown at every restore as "could not be read when this checkpoint was taken ... have NOT been restored". Deleting a file then checkpointing is routine, especially with the daemon's automatic pre-trigger checkpoint, so this turns a complete restore into a scary warning - the failure mode this feature exists to prevent, inverted. Suggest excluding `ENOENT` from `skipped` and keeping the `logWarning`, so `skippedFiles` means "existed but unreadable". The `CheckpointManager records files it could not capture` test pins the current behaviour with a nonexistent `locked.txt`, so it would need a real read failure instead (`chmod 000`, as the `validateRestorePath` tests already do).

Smaller things, none blocking:

- The gaps block is copy-pasted three times (`checkpoint.tsx` twice, `useAppHandlers.tsx` once). `CheckpointManager.restoreFiles` already receives the whole `checkpointData`, so having it return the gaps would stop a fourth restore path from forgetting.
- `if (metadata.truncatedFileCount)` relies on truthiness next to an explicit `skipped.length > 0`. Prefer `> 0`.
- The six-space indent in `describeCheckpointGaps` is coupled to the bullet prefix added by `describeGapsMessage`. Two functions co-owning one layout will drift.
- Both detail lists are unbounded; with the 50-file cap plus full error strings a bad checkpoint can dump 50 lines into the chat. Worth capping with "and N more".
- `skipped.push({path: filePath})` stores the raw input path while `filesChanged` stores the normalized form. Benign today, but the two lists carry different guarantees.
- The truncation test handles gpg signing but not global hooks; `--no-verify` and `-c core.hooksPath=` would make it robust for contributors with a global hook path.
- The changeset is a single ~250 word paragraph, much longer than the rest of the changelog. Two or three sentences plus the issue reference would fit better.

addyCooks:
Thanks @will-lamerton,  
great review.... both blocking items are done.

**1.** Merged `main` in. Dropped `ModifiedFiles`, added `truncatedCount` to `getModifiedFilesResult()`. All three `timeline-manager` breakages fixed and you were right that `isProbablyBinary` needed more than a retype: it now takes `string | Buffer` and checks byte `0` directly, so the NUL check runs on raw bytes before any decode.

**2.** ENOENT excluded from `skipped`, `logWarning` kept. Also covered on the other path `filesChanged` is built from captured snapshots, so a deleted file can't resurface as an unreadable-at-load gap either.

One deviation: I used a directory where a file is expected (EISDIR) instead of `chmod 000`. Windows can't drop read permission that way, so that test would fail for Windows contributors awkward given locked files on Windows are the case this was written for.

All seven smaller things are in too.

Verified against the issue's original repro table every row that read `false` now reads `true`.

One left for you: `snapshotPaths` still decodes UTF-8 after the binary check. Not a regression (it matches main today), but it is a decode inside the PR that removes decodes. Carrying `Buffer` end-to-end is ~6 edits plus spec churn here or a follow-up, your call.

will-lamerton:
Great work :)

addyCooks:
Thanks @will-lamerton : )

PR #966, opened 2026-08-26
fix(events): normalize path separators so file.changed globs match on Windows

### LINKED ISSUE #964: [Bug] Windows: event-router glob matching assumes "/" separators, so file.changed subscriptions silently miss files or fire out of scope
## Description

The daemon's file-watch feature does not work correctly on Windows. Path-scoped `file.changed` subscriptions either never fire, or fire on files outside their scope silently, with nothing logged either way.

**What's happening:** chokidar reports changed files using the platform separator, so on Windows it emits `docs\guide.md`. The router's glob matcher (`globToRegex`, `source/events/event-router.ts:111`) is written for POSIX only it treats `/` as a literal and compiles `*` to `[^/]`. So it tries to match the pattern `docs/**` against the path `docs\guide.md`, and the two can never meet. `FileWatcherSource.emit` (`source/events/sources/file-watcher.ts:68`) forwards the platform path unchanged, so nothing normalizes it in between.

This produces two distinct failures:

- **Silent under-match** any pattern containing a path segment (`docs/**`, `src/**/*.ts`) never matches. The daemon starts cleanly, the watcher attaches, events are emitted and evaluated normally. The subscription just never runs, and nothing reports a problem.
- **Silent over-match** `[^/]` does not exclude `\`, so `*` crosses directory boundaries on Windows. A root-scoped pattern like `*.md` matches the nested `sub\a.md`, dispatching an unattended agent run against a file the user deliberately scoped out.

**Every path example in the documentation is affected.** `docs/features/skills.md` publishes `paths: ["docs/**"]` (lines 48 and 136) and `paths: ["k8s/**/*.yaml"]` (line 82):

```
  SILENT  docs/**        vs "docs\\guide.md"
  SILENT  docs/**        vs "docs\\api\\ref.md"
  SILENT  k8s/**/*.yaml  vs "k8s\\deploy.yaml"
  SILENT  k8s/**/*.yaml  vs "k8s\\prod\\deploy.yaml"
  MATCH   *.md           vs "sub\\a.md"     <- should NOT match, file is nested
```

A Windows user who copies the documented example gets a daemon that runs, reports healthy, and does nothing.

Not every pattern fails, which is worth stating plainly: patterns *starting* with `**` match by accident, because `.*` spans a backslash happily. So `**/*.md` works while `docs/**` does not. That inconsistency makes the failure look intermittent rather than systematic, and is likely why it has gone unnoticed.

There is a smaller downstream consequence: `SkillDispatcher` JSON-stringifies the raw payload into the model's prompt (`source/skills/dispatcher.ts:179-180`), so a triggered agent receives `"docs\\guide.md"` with escaped separators rather than the path shape the rest of the tooling uses.

## Environment

- **OS**: Windows 11 (any Windows; macOS and Linux are unaffected)
- **Node version**: v22.22.3
- **Provider**: N/A
- **Model**: N/A reproducible with no LLM involved; the event router is provider-agnostic

## Steps to Reproduce

1. On Windows, create a skill with `subscribe: [{kind: file.changed, paths: ["docs/**"]}]` the example straight from `docs/features/skills.md`
2. Run `nanocoder daemon start`
3. Edit any file under `docs/`
4. The subscription never fires. `nanocoder daemon status` shows the daemon healthy and `nanocoder daemon logs` shows nothing, because no error occurred

## Expected Behavior

A `paths: ["docs/**"]` subscription fires when a file under `docs/` changes, on every platform. A root-scoped pattern such as `*.md` does not match nested files.

## Actual Behavior

On Windows the subscription never fires, and root-scoped patterns match out of scope. Output from the test above:

```
  platform      : win32
  emitted paths : ["docs\\guide.md"]
  contains "\\" : true
  contains "/"  : false

  matchGlob("docs/**", emitted)    = false
  matchGlob("docs/*.md", emitted)  = false
  matchGlob("**/*.md", emitted)    = true
  matchGlob("*.md", "sub\\a.md")   = true   <- should be false, file is nested
```

The first line confirms the emitted path really does contain a backslash and no forward slash this is the actual output of the shipped `FileWatcherSource`, not a constructed string.

## Logs/Screenshots

Nothing is logged. The daemon starts, the watcher attaches, events are emitted and evaluated normally the pattern simply never matches, so no code path treats it as an error. The absence of any log line is part of the bug: there is no signal that a subscription is dead. The test output above is the evidence.

### Proposed spec

**Fix:** normalize separators to `/` in `FileWatcherSource.emit`, and defensively in `matchGlob` for both pattern and path. Once the path uses `/`, `[^/]` stops at a real separator so the over-match disappears along with the under-match.

**Why the watcher and not just the matcher:** normalizing at the boundary means the router, activity reports, and the payload handed to the model all speak one path shape.

**On picomatch:** the comment at `event-router.ts:100` proposes deleting this helper in favour of chokidar's picomatch. Not free under pnpm's strict linking it isn't resolvable as a transitive dependency, so it would need adding as a direct one. Fine as a follow-up; shouldn't gate this fix.

**Tests:** table-driven cases in both directions `docs/**` matches `docs\guide.md`, `*.md` does *not* match `sub\a.md`. The suite covers POSIX paths only today, which is why CI is green.

addyCooks:
## Description

chokidar reports changed files using the platform separator, so on Windows the router was asked to match a `/`-authored pattern like `docs/**` against `docs\guide.md` a comparison that can never succeed. Path-scoped `file.changed` subscriptions therefore never fired, and nothing reported a problem: the daemon started, reported healthy and logged nothing, because no code path treats a non-match as an error.

The same mismatch ran the other way too. `globToRegex` compiles `*` to `[^/]*`, which does not exclude a backslash, so a root-scoped pattern such as `*.md` matched the nested `sub\a.md` and dispatched an unattended agent run against a file that was deliberately scoped out.

Normalize in `FileWatcherSource.emit` rather than only in the matcher, so the router, activity reports and the payload handed to a triggered agent all read one path shape; `matchGlob` normalizes both pattern and path defensively. Once the path uses `/`, `[^/]` stops at a real separator and the over-match disappears along with the under-match.

The picomatch swap noted at `event-router.ts:100` is left as a follow-up it is not resolvable as a transitive dependency under pnpm's strict linking and would need adding as a direct one.

Closes #964.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Changeset

### Automated Tests

### Manual Testing

- [ ] Tested with Ollama
- [ ] Tested with OpenRouter

## Checklist

- [x] Self-review completed

### DIFF
--- a/.changeset/fix-windows-event-glob-separators.md
+++ b/.changeset/fix-windows-event-glob-separators.md
@@ -1,0 +1,5 @@
+---
+"@nanocollective/nanocoder": patch
+---
+
+Fixed `file.changed` skill subscriptions silently never firing on Windows. Chokidar reports changed files using the platform separator, so a documented pattern like `docs/**` was asked to match `docs\guide.md` and never could — the daemon started, reported healthy, logged nothing, and did nothing. The same mismatch let a root-scoped pattern such as `*.md` match the nested `sub\a.md`, dispatching an unattended agent run against a file that was deliberately scoped out. Paths are now normalized to `/` at the watcher boundary, so the router, activity reports, and the payload handed to a triggered agent all read one path shape. Closes #964.
--- a/source/events/event-router.spec.ts
+++ b/source/events/event-router.spec.ts
@@ -165,3 +165,33 @@ test('matchGlob: basic patterns', t => {
 	t.true(matchGlob('file.txt', 'file.txt'));
 	t.false(matchGlob('file.txt', 'other.txt'));
 });
+
+// Chokidar reports the platform separator, so on Windows a `/`-authored
+// pattern is asked to match `docs\guide.md`. Both directions matter: a
+// segment pattern must still match, and `*` must not cross a backslash any
+// more than it crosses a slash.
+test('matchGlob: paths using Windows separators', t => {
+	const cases: Array<[pattern: string, path: string, expected: boolean]> = [
+		['docs/**', 'docs\\guide.md', true],
+		['docs/**', 'docs\\api\\ref.md', true],
+		['docs/*.md', 'docs\\guide.md', true],
+		['k8s/**/*.yaml', 'k8s\\deploy.yaml', true],
+		['k8s/**/*.yaml', 'k8s\\prod\\deploy.yaml', true],
+		['src/**/*.ts', 'src\\deep\\file.ts', true],
+		['**/*.md', 'docs\\intro.md', true],
+		['docs/**', 'src\\index.ts', false],
+		// `*` stops at a directory boundary, whichever separator draws it
+		['*.md', 'sub\\a.md', false],
+		['src/*.ts', 'src\\deep\\file.ts', false],
+		['?.md', 'sub\\a.md', false],
+	];
+	for (const [pattern, path, expected] of cases) {
+		t.is(matchGlob(pattern, path), expected, `${pattern} vs ${path}`);
+	}
+});
+
+test('matchGlob: patterns authored with Windows separators', t => {
+	t.true(matchGlob('docs\\**', 'docs/guide.md'));
+	t.true(matchGlob('docs\\**', 'docs\\guide.md'));
+	t.false(matchGlob('docs\\*.md', 'docs/deep/guide.md'));
+});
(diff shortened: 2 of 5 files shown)


will-lamerton:
Looking great @addyCooks :)

addyCooks:
Thankyou @will-lamerton !

PR #967, opened 2026-08-27
feat: add JSON schema for agents.config.json (closes #960)

### LINKED ISSUE #960: [Feature] JSON Schema for agents.config.json
## Description

There are typescript definitions here: https://github.com/Nano-Collective/nanocoder/blob/main/source/types/config.ts#L140. Can nanocoder generate and publish a JSON schema from these?

## Use Case

Enable IDEs to show the schema when editing agents.config.json.

## Any alternative solutions or features you've considered

I generated my own using [typescript-json-schema](https://www.npmjs.com/package/typescript-json-schema) and serve it to myself as a [gist](https://gist.github.com/garyg1/ca9d004ba9514d565158e80f8ee57286).

i.e.

```jsonc
{
    "$schema": "https://gist.githubusercontent.com/garyg1/ca9d004ba9514d565158e80f8ee57286/raw/d6b743c0c1ae959b2396c9e172cb57f3a2f05cb2/agents.config.schema.json",
    "nanocoder": { /* ... */ }
}
```

## Additional Context

Link to the existing documentation: [configuration/index.md](https://github.com/Nano-Collective/nanocoder/blob/main/docs/configuration/index.md)

tusharui:
Closes #960

## Description
Auto-generate a JSON schema from `AppConfig` types so editors (VS Code, JetBrains, Neovim) get autocompletion and validation for `agents.config.json`.

## Changes
- Generate schema via `typescript-json-schema` with `--required` flag
- Wrap under `nanocoder` key with `additionalProperties: false`
- Rename ugly TypeScript-generic definition names (e.g. `Record<string,string>` → `RecordStringString`)
- Extract descriptions from JSDoc comments
- Add `jsonValidation` to VS Code extension for auto-linking
- Add drift detection tests (9 tests)
- Add schema validation tests (8 tests)
- Update docs with usage example
- Add `generate:schema` script, runs on `prepublishOnly`

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

## Changeset

### Automated Tests

### Manual Testing
- [x] Schema validates against valid configs
- [x] Schema rejects unknown keys, missing required fields, extra properties

## Checklist
- [x] I was assigned to issue #960
- [x] Self-review completed
- [x] Documentation updated
- [x] No breaking changes
- [x] Appropriate logging added


will-lamerton (changes requested):
Thanks for this - the approach is right (types as the source of truth, schema committed and shipped in `files`), and the definition renaming is clean. A few things need fixing before it can go in.

## Blocking

1. **`additionalProperties: false` on `AppConfig` rejects `retries` and `defaultMode`.** Both are read by the loader (`source/config/index.ts:364` and `:560`) and documented (`docs/configuration/index.md:183`, `:235`), but neither is a member of `AppConfig`, so both are missing from the schema. Anyone who followed those docs now gets a false error. Best fix is adding them to `AppConfig`, since the type is genuinely incomplete.

2. **Root `additionalProperties: false` rejects two more valid things.** The top-level `providers` form supported at `source/config/mcp-config-loader.ts:274` and used in `docs/configuration/providers/openrouter.md:40` and `docs/features/tune.md:121`, and the `$schema` key the new docs section tells users to add. For the latter, declare `"$schema": {"type": "string"}` in root properties.

3. **`--required` is too strict for a file schema.** It marks all five `AutoCompactConfig` fields required, but the loader defaults each one independently (`source/config/index.ts:179-197`), so `{"autoCompact": {"threshold": 70}}` is valid at runtime and invalid per the schema. Same for `paste` and `notifications`. Suggest dropping `--required` and hand-listing the few genuinely required fields.

4. **The changeset names a package that does not exist.** It uses `nanocoder`; the package is `@nanocollective/nanocoder`, so `changeset version` will error. Should also be `minor` rather than `patch` for a feature.

5. **`ajv` is added to devDependencies but never imported**, so `pnpm run test:knip` will fail on an unused dependency. The description mentions 8 schema validation tests but they are not in the diff, and those are exactly the tests that would have caught items 1 to 3.

## Worth addressing

- No `--tsconfig`, so the `@/` path aliases fail to resolve and the script filters those failures out as expected. The visible cost is `ModeProvidersConfig` collapsing to `{"type": "object"}`, so `modeProviders` gets no completion at all. Passing `tsconfig.json` as the first argument should fix it and let you delete the whole error-swallowing block.
- `extractDescriptions` plus the two hardcoded description maps duplicate JSDoc support `typescript-json-schema` already has. Converting the comments in `source/types/config.ts` to `/** */` would remove all of it, and avoid the current parser keying descriptions by bare property name across the whole file (`enabled`, `timeout`, `name` collide across interfaces).
- The drift tests use a hardcoded property list, so adding a field to `AppConfig` without regenerating passes green. A CI step running `pnpm run generate:schema && git diff --exit-code schemas/` catches every drift with nothing to maintain by hand.
- `notifications` goes the other way from item 1: it is read from preferences (`source/config/preferences.ts:134`), not `agents.config.json`, so the schema advertises a key that file ignores. The underlying point is that `AppConfig` is the resolved runtime shape rather than the on-disk shape, and the two have drifted in both directions.

Happy to take another look once these are sorted.

will-lamerton (changes requested):
Much better - the separate `DiskConfig` type is a cleaner answer than what I asked for, and the CI drift job is exactly right. I checked out the head and ran it: `test:types`, `biome check`, `test:knip` and the 29 schema tests all pass, and `pnpm generate:schema && git diff --exit-code schemas/` is clean, so the generator really is reproducible.

Items 3, 4 and 5 are resolved, as are all four of the "worth addressing" points. Two of the blockers each had two parts, and one part of each is still open.

## Blocking

1. **`nanocoder.retries` is still missing.** `defaultMode` was added, `retries` was not. It is on `AppConfig` already (`source/types/config.ts:318`), read by the loader (`source/config/index.ts:363`) and documented (`docs/configuration/index.md:196`), but absent from `DiskNanocoderConfig`, which is `additionalProperties: false`. Ajv against the committed schema:

```
REJECT  {"nanocoder": {"retries": {"maxRepeatedToolCalls": 5}}}
        → /nanocoder must NOT have additional properties
```

2. **Top-level `providers` is still rejected.** `$schema` was added, but root properties are only `$schema` and `nanocoder` under `additionalProperties: false`, while `loadProjectProviderConfigs` still falls back to `config.providers` (`source/config/mcp-config-loader.ts:274`), a form used in `docs/configuration/providers/openrouter.md:40` and `docs/features/tune.md:121`.

```
REJECT  {"providers": [{"name": "OpenRouter", "models": ["a"]}]}
        →  must NOT have additional properties
```

## Worth addressing

- The `Sessions` docs section is now a false-error trap. `docs/configuration/index.md:116` states application settings go in `agents.config.json` under `nanocoder`, and `### Sessions` at `:146` sits under that heading with a matching example, but the loader reads sessions from `nanocoder-preferences.json` (`source/config/index.ts:260`). Excluding it from the schema is correct, so the fix is in the docs - that section should name the preferences file, otherwise the first thing this schema does for anyone who followed those docs is flag a working config.
- The `ModeProviders` value schema is hand-written inside `expandRecords` rather than `$ref`-ing `ModeProviderConfig`. Add a field to that interface and regeneration is byte-identical, so the drift job passes green while the schema is wrong - the same hand-maintained-list failure mode the old drift tests had.
- `modeProviders` keys are unconstrained: `DiskNanocoderConfig` uses `Record<string, ModeProviderConfig>` where the loader accepts only `VALID_MODES` and logs an error otherwise (`source/config/index.ts:497`). Keying on `DevelopmentMode` would catch typos and give completion for the mode names.

Close on these and it should be good to go.

will-lamerton:
Nice work @tusharui

Issue #808, opened 2026-08-06
[Feature] AI-Generated Smart Session Titles & Manual Renaming


akramcodez:
## Description

Replace the current raw-text session naming logic—which simply captures the user's first prompt (e.g., "hi", "fix this", or "can you help")—with a lightweight background LLM step that generates a concise, descriptive title for the session based on the actual work being done.

Additionally, add a manual "Edit" button in the session history sidebar (to the left of the delete button) so users can manually rename sessions if they want to override the AI's title.

## Use Case

When users check their chat history panel, seeing a dozen sessions named "hi", "fix", or "hello nanocoder" makes it impossible to navigate and find past work. 

By having the agent automatically summarize the actual goal of the session (e.g., "Refactoring API Auth" or "Debugging CSS Grid") after the conversation gets going, the session history becomes vastly more useful, searchable, and professional—similar to how ChatGPT, Claude, and Antigravity IDE handle session names.

## Proposed Solution

- In `source/session/session-manager.ts` (or the chat handler loop), introduce a trigger that fires after the first agent tool execution or after the second conversational turn.
- This trigger sends a low-token, background LLM request: `"Summarize the goal of this conversation into a 3-5 word title."`
- Overwrite the initial raw prompt session name in the state with this newly generated smart title.
- Emit a UI event so the Webview/CLI updates the sidebar to reflect the new name dynamically without requiring a page refresh.
- **Manual Edit Button**: Add an "Edit" (pencil) icon to the left of the existing "Delete" button in the session history panel. Clicking it should turn the session title into a text input, allowing the user to manually rename the session.

## Alternatives Considered

- **First N Characters of Prompt**: (The current behavior). Fast, requires no API calls, but often useless for finding past context.
- **Only AI Auto-Naming (No manual edit)**: This is too restrictive; users should always have the ability to correct or refine the session name if the AI gets it wrong.

## Additional Context

- [x] I have searched existing issues to ensure this is not a duplicate
- [x] This feature aligns with the project's goals (local-first AI assistance)

arpan7sarkar:
Hey @akramcodez Just want to tell you that i have alredy implemneted the manual name changing in #814 , and for reference https://github.com/Nano-Collective/nanocoder/issues/749#issuecomment-5199817998 , Now only the ai based Title generation is left so assign me it so i  can add that .

akramcodez:
[...]

PR #978, opened 2026-08-27
feat:Auto AI generated titles in extension

arpan7sarkar:
## Description

Implements #808

Sessions in the History list are named after the opening prompt verbatim, so they end up called "hi" or "fix this". The agent now generates a descriptive title once per session and pushes it to the extension over ACP as a `_nanocoder/sessionTitleChanged` notification, so the History list refreshes in place. It only fires when the opening prompt is too thin to be useful and there's real context to name, uses the session's own client by default (`sessions.titleModel` / `titleProvider` to override, `sessions.smartTitles: false` to disable), and never overwrites a manual rename.

While wiring this up, found the CLI's autosave derived its title from the *latest* user message and rewrote it on every save, into the same `sessions.json` the extension reads, so it would have stomped any generated title within 30 seconds. Fixed that, plus `loadSessionConfig` silently dropping the new config keys.

https://github.com/user-attachments/assets/4c4df7b2-868a-4e52-8abe-fa45ca85179c

## Notes for reviewers

Existing sessions are retitled from their first message on their next autosave.
That follows from the autosave fix above, but it is a visible one-time reshuffle of everyone's History list.

Titling runs in the ACP agent, so it applies to the VS Code extension and other ACP clients.
The CLI keeps its heuristic title.

`source/hooks/chat-handler/utils/message-helpers.spec.ts` is unrelated to titling.
It fixes a stale fourth argument left over from 6c1e5f39 and swaps a `t.pass()` for real assertions on the rendered output.
Happy to pull it into a separate PR if you would rather it did not ride along here.

The three new `sessions.*` keys are documented in the config type and the changeset, but not yet in `docs/`.
Note they are read from `nanocoder-preferences.json`, not `agents.config.json`.

## Type of Change

- [x] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

## Changeset

### Automated Tests

- [x] All existing tests pass (pnpm test:all completes successfully)

### Manual Testing

- [ ] Tested with Ollama
- [ ] Tested with OpenRouter

## Checklist

- [x] Self-review completed


arpan7sarkar:
Hey @will-lamerton , @akramcodez , @Avtrkrb  have a look here  :-)

akramcodez:
@arpan7sarkar could you please resolve the conflicts so we can do the final review? Btw, great work! 🔥

will-lamerton (changes requested):
Nice work - the layering is right (one impure function, everything else pure and directly tested), the re-read-before-write guard against a mid-flight rename is a real race most implementations miss, and `acp-ext-notification.spec.ts` verifying `notify()` -> `extNotification` over a real ndJSON stream is the one thing a stub could never catch.

I verified the branch locally: `tsc --noEmit`, `knip`, and `biome check` are all clean, and the touched specs pass.

A few things I'd like addressed before merge.

### Blocking

**1. A model that never returns a usable title re-calls on every turn, forever.** `titleGenerated` is only set on success, so if `generateSessionTitle` returns `null` (paragraph response, connection refused, timeout) nothing is recorded and the next successful turn runs the whole thing again, indefinitely. Small local models are both the target audience and the population most likely to ignore "reply with only the title", so this is the expected path rather than the edge case. The result is a silent extra LLM call every turn for the life of the session. A per-process failure counter (give up after 2-3) or a persisted attempt count would cap it.

**2. The `inFlight` guard has a gap it claims to close.** `inFlight.has()` is checked at `maybe-generate-title.ts:53` but `add()` is at `:68`, with `await manager.readSession()` in between, so two callers can both pass the check before either adds. Small window in practice, but the fix is free: move `inFlight.add(sessionId)` up to right after the synchronous guards (line 62) and extend the existing `try/finally` to cover it.

**3. A `chat()` that ignores the abort signal wedges the session permanently.** `TITLE_TIMEOUT_MS` relies entirely on the provider honouring `signal`. If one doesn't, the promise never settles, `inFlight.delete` never runs, and that session can never be titled again for the process lifetime. `Promise.race([generateSessionTitle(...), timeout])` would make the 20s bound actually hold.

### Worth answering

**4. Titling tokens are invisible in usage tracking.** `buildResponseUsage` is called from `conversation-loop.tsx` and `acp-conversation.ts`, not from inside `client.chat`, so these tokens are billed by the provider but never show up in `/usage`. The feature is on by default, so this deserves at least a line in the changeset.

**5. `smartTitles` is a global config key but only the ACP path generates.** `maybeGenerateTitle` has exactly one call site. CLI users get the documented config keys but no generated titles. Either wire it into the CLI turn completion or scope the naming and docs to the extension. (Minor: the comment at `maybe-generate-title.spec.ts:902` says "Both call sites" - there is one.)

**6. Test-coverage gap on the headline autosave fix.** `useSessionAutosave.spec.ts` wasn't updated, and every existing `deriveSessionTitle` case has at most one real user message, so they pass identically under both the old backward scan and the new forward scan. Nothing pins the behaviour the PR fixed. A case with two substantive user turns asserting the first wins would close it.

Also worth calling out in the description: on next autosave, existing sessions get retitled from their first message. Expected, but it's a visible one-time churn in everyone's history list.

### Smaller notes

- `PATH_ARG_KEYS` includes `command`, so bash command strings reach the title model. Fine by default (same model as the session), but with `titleProvider` set, paths, commands and up to 3 user turns go to a third party. One sentence in the config doc comment would cover it.
- `ACTIVE_FILE_PREFIX` doesn't match `\r\n`, and the same regex is duplicated inline at `acp-agent.ts:906`. Worth having that call site use the exported constant so the two can't drift.
- `WEAK_TITLE_THRESHOLD = 40` characters is very high for CJK, where 40 chars is a long prompt. The comment acknowledges the tradeoff, but in practice nearly every CJK session will trigger a call.
- `acp-ext-notification.spec.ts` never closes its streams or connections. It passes, but a `t.teardown` closing the writables would stop a future change leaving the AVA worker hanging.
- The new `acp-agent.spec.ts` negative assertions use fixed 200ms sleeps against fire-and-forget work. Deterministic today, but the kind of thing that goes flaky on a loaded CI box.
- `cachedClient` in `title-client.ts` is never invalidated on config change. `resetTitleClientCache` already exists, so calling it from `clearAppConfig` would be cheap.
- The `message-helpers.spec.ts` rewrite is unrelated to this feature (it fixes a stale 4th argument left over from 6c1e5f39 and swaps `t.pass()` for real prop assertions). Strict improvement, but it belongs in its own commit or a line in the description.

arpan7sarkar:
Thanks @will-lamerton  for the review 🫡 , I will look into them all and finalize the pr let's see what new things I learn from here

will-lamerton:
All looking great @arpan7sarkar - just a conflict to resolve and we're good to merge :)

will-lamerton:
Great work @arpan7sarkar :)

arpan7sarkar:
Thank you @will-lamerton  :-)

PR #123, opened 2025-12-03
Move MCP and LSP connection status to background initialization with summarized status display

Avtrkrb:
## Description

This pull request implements the proposal to move MCP and LSP connection status from verbose startup chat messages to background initialization with a persistent, summarized status display in the UI status component.

**Key Changes:**
- **Background Initialization**: MCP and LSP servers now connect silently in the background without cluttering the chat area
- **Dynamic Status Display**: Real-time connection progress shown as "MCP: X connected, Y errors" and "LSP: X ready, Y errors" in the status component
- **Clean User Experience**: Users can immediately start typing after the welcome message
- **Preserved Functionality**: `/mcp` and `/lsp` commands continue to work for detailed connection information

**Technical Implementation:**
- Enhanced `useAppInitialization.tsx` with Map-based server state tracking for real-time progress updates
- Modified `status.tsx` to display MCP/LSP connection status with proper formatting for auto-discovery phases
- Updated `app.tsx` to pass connection status props and enable dynamic re-rendering
- Implemented proper React state management with immutable updates

## Type of Change

- [x] Bug fix (LSP status display during auto-discovery)
- [x] New feature (Dynamic MCP/LSP status display)
- [ ] Breaking change
- [ ] Documentation update

### Automated Tests

- [x] New features include passing tests in `.spec.ts/tsx` files
- [x] All existing tests pass (`pnpm test:all` completes successfully - 1693/1694 tests pass, with 1 timeout unrelated to changes)
- [x] Tests cover both success and error scenarios

**New Test Coverage (15 tests total):**
- `source/components/status.spec.tsx` - 6 tests for status component MCP/LSP display logic
- `source/hooks/useAppInitialization.spec.ts` - 4 tests for app initialization status management  
- `source/hooks/__tests__/connection-status.spec.ts` - 5 tests for React state update patterns and React.memo behavior

**Test Scenarios Covered:**
- MCP connection progress tracking with dynamic server counting
- LSP auto-discovery phase handling (critical fix for `totalCount: 0` during discovery)
- State object immutability patterns for React re-renders
- Status formatting logic edge cases and error conditions
- React.memo comparison behavior for optimal rendering

### Manual Testing

- [x] Tested with Ollama (MCP servers show dynamic connection progress)
- [x] Tested with OpenRouter (LSP servers display real-time status during discovery)
- [x] Tested with OpenAI-compatible API (Background initialization works correctly)
- [x] Tested MCP integration (Multiple servers connect in parallel with accurate status updates)

[...]

PR #990, opened 2026-08-27
fix: show working directory and git branch in boot summary

saksham-rathore:
## Description

Fixes the boot summary display so NanoCoder shows the current working directory and active Git branch.

Previously, the boot summary could display the `agents.config.json` path, which does not represent the project the user is currently working in.

The boot summary now displays the working directory together with the current Git branch in a compact format, for example:

`~/nanocoder · main`

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Changeset

### Automated Tests

- [x] Added/updated tests for the boot summary behavior

### Manual Testing

- [x] Verified the boot summary displays the current working directory
- [x] Verified the active Git branch is displayed
- [x] Verified the display updates correctly when using a different Git branch
- [x] Verified the display remains compact
- [x] Verified behavior outside a Git repository, if applicable

## Checklist

- [x] Self-review completed
- [x] No breaking changes
- [ ] Appropriate logging added using structured logging (if applicable)


saksham-rathore:
image
hey, @akramcodez can you help me with this thing I don't know what this means this is my first PR.

will-lamerton (changes requested):
Thanks for the PR, and welcome! The core idea is right: the boot summary showing the resolved `agents.config.json` path rather than the directory you are working in was genuinely confusing. A few things need addressing before this can go in, plus a walkthrough of the changeset question you asked about at the bottom.

## Blocking

### 1. `source/tools/git/utils.ts` changes a shared formatter, not just the boot summary

`formatGitStatusSummary` is also consumed by the `/status` panel (`source/components/status.tsx:24`). Its JSDoc says as much: it exists so "the boot summary and the /status panel can't drift on wording." Deleting the `isDefault` branch silently changes `/status` too, from `Git: main (default)` to `Git: main`, which is outside the scope of this fix.

If you want the compact label in the boot summary, strip the marker at the call site instead: `formatBootSummaryGitLabel` in `source/app/components/app-container.tsx:19`. Leave the shared helper alone.

This also breaks a test. Applying the diff on a clean checkout and running the affected specs:

```
components › status › Status marks default branch in normal layout
source/components/status.spec.tsx:801
Value must match expression: /main \(default\)/
│  Git: main
1 test failed
```

So the "All existing tests pass" checkbox in the description isn't accurate right now. Worth re-running `pnpm run test:ava source/components/status.spec.tsx` locally to confirm once you've made the change.

One knock-on effect if the removal does stay: nothing in production would read `GitStatusSummary.isDefault` any more, yet `getGitStatusSummarySync` still runs the full `getDefaultBranchSync` filesystem walk (packed-refs parsing, `refs/heads` probing) on every boot to compute it. knip won't catch that because it's an object property rather than an export.

### 2. `.github/dependabot.yml` is unrelated to this change

The only diff is the deletion of the trailing newline (`\ No newline at end of file`). Looks like an editor artifact. Please revert that file.

### 3. Missing changeset

The bot is right, there's no `.changeset/*.md` in the diff. This is user-facing so it needs one. Full walkthrough below.

## Non-blocking, but worth fixing while you're here

**Stale naming and comments.** `app-container.tsx:57` still names the variable `shortConfig` when it now holds the cwd. Two doc comments also go stale: the `BootSummary` JSDoc ("Minimal one-liner showing provider/model (+ optional mode) + config path") and the `nonInteractiveMode` prop doc ("provider · model · mode · config").

**`process.cwd()` now runs on every render.** `status.tsx:29` captures it once at module load (`const cwd = process.cwd()`). Worth hoisting out of the component for consistency.

**No test covers the new behaviour.** The only spec change flips an existing assertion. `app-container.spec.tsx` already has a `createStaticComponents` harness, so a test asserting the home-relative cwd actually appears in the boot summary would cover the fix directly.

**Narrow terminals.** The `isNarrow` branch never renders `shortConfig`, so the working directory isn't shown there at all. That's pre-existing behaviour rather than something you introduced, but flagging it since the description mentions verifying the display stays compact.

## Two questions

1. The config path is removed entirely rather than moved. Given nanocoder's project / platform / legacy config resolution order, knowing which `agents.config.json` is actually in effect was a useful signal. Is dropping it intended, or should the line carry both?
2. Is there an issue this closes? The description doesn't link one, so it'd be good to reference it.

## How to add a changeset

This is the bit you asked about, and it's a reasonable thing to trip over on a first PR. From the repo root, on your branch:

```bash
pnpm changeset
```

It'll ask you for a bump type, then open an editor for the entry:

- **patch** for bug fixes with no behaviour change. That's this PR.
- **minor** for new backwards-compatible features.
- **major** for breaking changes.

The important part: the markdown body you write becomes the CHANGELOG entry verbatim. So write it as a finished release note for users, not as a commit message. Then commit the generated `.changeset/*.md` file along with the rest of your changes and push.

If you'd rather skip the interactive prompts, hand-writing the file is equally valid. Create `.changeset/boot-summary-cwd.md`:

```markdown
---
"@nanocollective/nanocoder": patch
---

Fixed the boot summary showing the resolved `agents.config.json` path instead of the directory you are actually working in. It now shows the current working directory alongside the active Git branch (`~/nanocoder · main`). Thanks to @saksham-rathore.
```

House voice, going by the existing entries in `.changeset/`: past tense, lead with the user-visible symptom, then what changed, then attribution (`Thanks to @username. Closes #123.`). `CONTRIBUTING.md` lines 378 to 402 is the canonical write-up.

Two things to avoid: don't bump the version in `package.json`, and don't edit `CHANGELOG.md` directly. Those are maintainer-only. The `release-prepare.yml` workflow rolls pending changesets into the changelog automatically.

## One process tip for next time

This PR is opened from your fork's `main` branch. It works fine, but a dedicated branch per PR (`git checkout -b fix/boot-summary-cwd`) makes iterating easier and keeps your fork's `main` clean for syncing with upstream. No need to redo it here.

Happy to re-review once the above is addressed. Thanks again for jumping in.

will-lamerton:
Thanks for this PR @saksham-rathore - feel free to add yourself as a contributor to our website via a PR which I will approve :)

PR #992, opened 2026-08-27
fix: compact provider, model, and mode on the chat input

### LINKED ISSUE #859: [Feature] Explore a cleaner layout for chat input controls
## Description

The chat input currently has three separate controls at the bottom:

* AI Provider: `claude`
* Model: `claude-sonnet-5`
* Approval Mode: `auto-accept`

How can we improve the layout and UX of these controls?

The current design takes up a significant amount of horizontal space and makes the input area feel slightly crowded.

## Question

Could we explore a cleaner way to organize these three options while keeping them easily accessible?

For example, could some of the settings be grouped into a single **Settings** or **Configuration** menu, while keeping the most frequently changed option directly visible?

The goal would be to make the chat input feel cleaner and more focused without making provider, model, or approval settings difficult to access.

## Additional Context

Open to different UI/UX approaches here. This is mainly a design exploration issue rather than a specific implementation proposal.

Dhirenderchoudhary:
## Description

Compacts the chat input footer so provider, model, and approval mode no longer sit as three labeled controls. Mode stays first with the short labels (`⏵⏵ auto` instead of `⏵⏵ auto-accept mode on`). Provider and model share one unlabeled chip (`claude/claude-sonnet-5`). `/model` and Shift+Tab are unchanged.

Closes #859.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

## Changeset

### Automated Tests

### Manual Testing

- [ ] Tested with Ollama
- [ ] Tested with OpenRouter

## Checklist

- [x] Self-review completed


will-lamerton (changes requested):
Thanks for picking this up! A few things need addressing before this can land.

**1. This doesn't close #859.** The "three separate controls at the bottom" in that issue are the VS Code webview composer dropdowns (`provider-trigger`, `model-trigger`, `mode-trigger` in `plugins/vscode/media/chat-panel.html`), not the terminal footer. The TUI footer never showed provider or model at all before this PR, so it can't be what the issue is describing. This change actually adds two more segments to that line rather than compacting anything, and the VS Code triggers are untouched. Relatedly, the new test's `notRegex(/AI Provider/)` and `notRegex(/Approval Mode/)` assertions are vacuous, those strings have never existed in this component.

**2. The branch is stale and conflicts with main.** `development-mode-indicator.tsx` has since gained the task badge (`taskLabelBase` / `taskHintExtra`) inside the exact `requiredWidth` / `fixedWidth` blocks you edit, and `test:types:vscode` is already in `package.json` on main. Please rebase and drop the `package.json` commit, it's out of scope here anyway.

**3. The chip is unbounded width inside the never-truncate budget.** Mode, tune and ctx are documented as never truncating, with session name and the editor filename sharing what's left. `modelSegment` is counted in `requiredWidth`/`fixedWidth` but never truncated itself. Something like `openrouter/qwen3-coder-480b-a35b-instruct` is around 45 columns, so on an 80-column terminal with a session name and an editor pill, `remaining` clamps to 0, session and filename fall back to their 6-char minimum, and the row wraps regardless. Either truncate the chip, drop it (or the provider prefix) when `isNarrow`, or move it into the truncatable pool.

**4. The spec updates don't test the change.** Every rewritten assertion is a prefix of the old label, `/▶ normal/` matches `▶ normal mode on` and `/⏵⏵ auto/` matches `⏵⏵ auto-accept mode on`. Since ink-testing-library renders at a fixed 100 columns these are all on the wide path, so they pass against both old and new code. Add `t.notRegex(output!, /mode on/)` somewhere, or assert the exact label. The `modelChip` fallbacks (provider-only, model-only, neither) are also uncovered, as is a long provider/model on a narrow terminal, which is where the width problem is.

**5. Minor:** wide terminals now lose mode discoverability. The `(Shift+Tab to cycle)` hint is gated on `isNarrow`, so wide terminals relied on the long label to explain the mode, and now get `⏵⏵ auto` with no hint at all. If the short label stays, consider showing the hint on wide too. Also worth noting `BootSummary` already prints `provider · model` at startup, so the footer chip duplicates it, defensible since the boot summary is frozen inside `<Static>`, but it should be a deliberate call. And the changeset describes a removal of "three titled controls" that doesn't happen in the diff.

The plumbing itself is clean and consistent with how `currentModel` and `tune` are already threaded, so none of this is wasted. My suggestion: split it into the VS Code composer work that #859 is actually asking for, and, if the shorter mode labels and a provider/model chip are wanted on their own merits, a separate PR rebased on main that bounds the chip width and keeps the specs discriminating.

will-lamerton (changes requested):
Thanks for taking the rewrite on board, this is a much better PR than the one I reviewed. The diff is now five files, none in `source/components/`, and it does what #859 actually asked for: group some settings into a Configuration menu, keep the most frequently changed one directly visible. Points 1 through 4 from my last review are addressed.

Specifically: the TUI footer changes are gone along with the vacuous `notRegex(/AI Provider/)` assertions; the branch merges cleanly and `package.json` is out of the diff; and `chat-panel-composer.spec.ts` is a real test file. The markup test slices `add-menu-btn`..`send-stop-btn` and asserts `provider-trigger` / `mode-trigger` are absent from the row, which fails against `main`, so it discriminates properly. `test:types` is clean and all 182 `source/vscode/` tests pass on the branch. The changeset is accurate now too.

Two things still need resolving, one of them a rendering bug.

**1. The nested-dropdown behaviour is occluded, so it buys nothing.**

`#composer-settings` (line 286), `#provider-dropdown` (303) and `#mode-dropdown` (305) are all siblings inside `#composer-box`, and all three carry identical `absolute bottom-[calc(100%+8px)] left-0 w-full ... z-50`. Equal z-index means DOM order decides paint order, and the dropdowns come later, so the provider list renders full-width directly on top of the settings popover and hides it completely.

The `closeAllDropdowns(keepId)` plumbing exists specifically to keep the popover open in that case, but the user never sees it stay open. Either offset or nest the sub-dropdown so the popover remains visible behind it, or drop the `keepId` branch entirely and let settings close like the model list does. As written it is complexity with no visible effect.

Worth noting the two tests asserting this (`opening a nested provider list keeps composer settings open` and the mode equivalent) pass because `createPanel()` builds a flat stub DOM with no layout, so they cannot catch it.

**2. Approval mode no longer has a persistent indicator.**

It is now entirely behind an unlabeled gear. Mode is safety-relevant state, and it is treated as never-truncating in the TUI footer for exactly that reason. A user sitting in `auto-accept` or `yolo` in the VS Code panel now has nothing on screen telling them so, and finds out when a tool runs without asking.

Provider is a good candidate for tucking away, mode less so. Reflecting the current mode on the gear itself (short label or colour) or keeping a compact mode badge on the row would still satisfy the issue's "without making provider, model, or approval settings difficult to access" while preserving the signal. Happy to go either way on the exact treatment, but it should not be invisible.

**3. Two tests click disabled buttons.**

`CustomDropdown.setOptions` leaves `trigger.disabled = true` until it gets a non-empty option list. The two nested-open tests click `provider-trigger` / `mode-trigger` without posting `syncState` first, and the harness ignores `disabled`, so they pass on clicks that would be inert in a real webview. `provider and mode still post the existing extension messages` seeds state correctly, so it is just a case of doing the same in those two.

**Minor, non-blocking:**

- `aria-haspopup="true"` would be more accurate as `"dialog"` or `"menu"`.
- No Escape-to-close or focus handling on the popover.
- Outside-click dismissal stays untested because the harness stubs `document.addEventListener` to a no-op. Pre-existing, just flagging it as a known gap rather than something you introduced.
- `t.false(settings.includes('id="settings-view"'))` is a weak assertion but not wrong, `settings-view` is a real element at line 49 and the slice legitimately should not contain it.

Point 1 is the one that needs a code change. Point 2 is a design call, tell me what you think and I am not going to be precious about which option you pick.

will-lamerton:
All good to go! @Dhirenderchoudhary :)

PR #994, opened 2026-08-27
Fix: mcp development mode policy

### LINKED ISSUE #989: [Bug] MCP tools bypass development-mode policy: execute in plan mode, auto-denied in headless
## Description

MCP tools do not participate in the development-mode policy that governs every other tool. `mcp-client.ts` hand-rolls its own approval predicate instead of routing through the central mode system, and `MODE_EXCLUDED_TOOLS` only contains built-in tool names so MCP tools slip past mode filtering entirely.

This fails in **both directions** from a single root cause:

| Mode | Expected | Actual |
| --- | --- | --- |
| `plan` | Blocked — plan mode is documented as "show tool calls but don't execute" | An `alwaysAllow`-ed MCP tool **executes immediately, with no approval prompt** |
| `headless` | Runs unattended, like `execute_bash` and the file tools do | **Auto-denied** with `"Tool execution was denied by the user."` — no user is present |

**Root cause**

`source/mcp/mcp-client.ts:444-446` builds its own policy:

```ts
const isAutoApproved = this.isToolAutoApproved(toolName, serverName);
const approval: ToolApprovalPolicy = isAutoApproved
    ? false
    : (_args, mode) => mode !== 'auto-accept';
```

- `isAutoApproved` → `approval: false` → `resolveToolApproval` rule #3 (boolean short-circuit) returns "no approval" in **every** mode, including `plan`.
- Otherwise → `'headless' !== 'auto-accept'` → `true` → approval required in headless, where no approval handler exists (`tool-approval-queue.ts:23` defaults to `() => false`), so the call is silently denied.

Two supporting gaps let this reach the user:

1. `source/tools/tool-manager.ts:42-55` — `MODE_EXCLUDED_TOOLS.plan` lists only built-ins (`write_file`, `execute_bash`, `git_commit`, …). MCP tool names are never in it.
2. `source/tools/tool-manager.ts:194-203` — the plan/headless filter only consults `this.customTools`; `if (!meta) return true` waves MCP tools straight through.

Compare `source/tools/execute-bash.tsx:179` and `source/utils/tool-approval.ts:18` both explicitly special-case `headless`. MCP was never updated to match.

## Environment

- **OS**: Windows 11 (platform-independent,  this is pure policy logic)
- **Node version**: v22.22.3
- **Nanocoder version**: 1.30.0 (current `main`, tag `v1.30.0`)
- **Provider**: Any, reproduces regardless of provider
- **Model**: Any, reproduces regardless of model

## Steps to Reproduce

**A. Plan mode executes a mutation (no MCP server needed for the unit repro)**

1. Configure any MCP server in `.mcp.json` with a mutating tool in its `alwaysAllow` list the format documented in `source/commands/mcp.tsx:66`:
   ```json
   { "mcpServers": { "github": { "command": "...", "alwaysAllow": ["create_issue"] } } }
   ```
2. Start nanocoder and switch to **plan** mode (Shift+Tab).
3. Ask the model to do something that calls `create_issue`.
4. The issue is created. No approval prompt appears.

**B. Headless denies an ordinary MCP tool**

1. Configure any MCP server **without** the tool in `alwaysAllow`.
2. Add a skill with a `subscribe:` block on `file.changed` whose subagent uses that MCP tool.
3. Start the daemon: `nanocoder daemon start`.
4. Touch a watched file.
5. The run returns `Tool execution was denied by the user.` no user was ever prompted.

**Unit-level repro** (drop into `source/tools/`, no MCP server or daemon required):

```ts
import test from 'ava';
import {resolveToolApproval} from './approval-policy.js';

// Mirrors exactly what mcp-client.ts:444-446 constructs.
const autoApproved = {approval: false as const};
const ordinary = {approval: (_a: unknown, mode: string) => mode !== 'auto-accept'};

test('plan mode must not execute an alwaysAllow-ed MCP tool', async t => {
    t.true(await resolveToolApproval('create_issue', autoApproved, {}, {mode: 'plan'}));
});

test('headless must not require approval for an ordinary MCP tool', async t => {
    t.false(await resolveToolApproval('create_issue', ordinary as any, {}, {mode: 'headless'}));
});
```

Both assertions fail on current `main`.

## Expected Behavior

- **Plan mode**: no MCP tool that mutates state is reachable, and no `alwaysAllow` entry can override that. Plan mode is the safety mode users switch into precisely to inspect an untrusted model's intentions without side effects.
- **Headless mode**: MCP tools behave like `execute_bash` and the file tools they run unattended, since headless exists specifically for daemon-triggered runs with no foreground prompt available.

## Actual Behavior

- **Plan mode**: `alwaysAllow`-ed MCP tools execute silently. The documented contract ("Show tool calls but don't execute") is violated with no warning to the user.
- **Headless mode**: ordinary MCP tools are denied with a message claiming a user declined, when no user was present. Any triggered skill that depends on MCP silently fails, and the error message actively misdirects debugging.

## Logs/Screenshots

Output from the unit repro above against `v1.30.0`:

```
✔ PLAN MODE: alwaysAllow-ed MCP tool executes with NO approval
✘ [fail]: HEADLESS: ordinary MCP tool is auto-DENIED
  Value is not `false`:
  true
```

## Additional Context

**Suggested fix**  make MCP entries first-class participants in the existing policy rather than special-casing them:

1. Have `getToolEntries()` return `readOnly` alongside `approval`, derived from the MCP tool's own annotations (`readOnlyHint` where the server provides it), so `resolveToolApproval` rule #5 works for MCP the way it already does for built-ins.
2. Replace the hand-rolled predicate with the shared helper, so `headless` is handled once: `mode !== 'auto-accept' && mode !== 'headless'` matching `createFileToolApproval` in `source/utils/tool-approval.ts:18`.
3. Make plan-mode filtering in `tool-manager.ts` deny-by-default for any non-read-only tool instead of enumerating built-in names, so future tool sources cannot silently inherit this bug.
4. Consider whether a server-level `alwaysAllow` should be able to override a *mode* at all currently it short-circuits ahead of every other rule.

Happy to open a PR with the fix plus the regression tests above.

addyCooks:
Fixes #989

## Description

MCP tools never participated in the development-mode policy that governs every other
tool. `mcp-client.ts` hand-rolled its own approval predicate:

```ts
const isAutoApproved = this.isToolAutoApproved(toolName, serverName);
const approval: ToolApprovalPolicy = isAutoApproved
    ? false
    : (_args, mode) => mode !== 'auto-accept';
```

This failed in both directions from one root cause. An `alwaysAllow`-ed tool collapsed
to `approval: false`, which `resolveToolApproval` rule `3` (boolean short-circuit) reads
as "no approval" in every mode so plan mode executed mutating MCP tools with no prompt.
Anything else evaluated `'headless' !== 'auto-accept'` to `true` and demanded approval in
headless, where no approval handler exists (`tool-approval-queue.ts` defaults to
`() => false`), so daemon-triggered runs were denied with a message blaming a user who
was never asked.

Two supporting gaps let it reach users: `MODE_EXCLUDED_TOOLS.plan` can only ever hold
built-in names, and the plan/headless filter consulted only `customTools`, so
`if (!meta) return true` waved MCP tools straight through.

## The fix

- **`mcp-client.ts`** the predicate now takes the same posture as built-in tools:

  ```ts
  if (mode === 'auto-accept' || mode === 'headless') return false;
  if (mode === 'plan') return !readOnly;
  return !isAutoApproved; // normal: alwaysAllow only, as documented
  ```

  Auto-accept and headless run unattended, matching `createFileToolApproval`. Plan is
  gated on the server's annotation. Normal is unchanged from documented behaviour.

- **`types/mcp.ts` + `mcp-client.ts`** carry the server's `annotations.readOnlyHint`
  onto the tool entry. An absent hint fails safe to "may mutate". This decides plan-mode
  *availability* only: `readOnlyHint` is supplied by the very server being gated, so it
  never skips a confirmation prompt in normal mode the user's `alwaysAllow` list stays
  the only thing that does.

- **`tool-manager.ts`** plan mode gates MCP tools on that annotation instead of an
  enumeration it can never hold. Headless deliberately untouched.

- A server's `alwaysAllow` no longer short-circuits ahead of every other rule it applies
  in normal mode only, which is what `docs/configuration/mcp-configuration.md` already
  said. The code was violating its own documentation.

- **Docs** `mcp-configuration.md` gains a per-mode table for `alwaysAllow`, and records
  that headless runs every MCP tool unattended with no per-server opt-out (more permissive
  than custom tools, which need `approval: never`), with `"enabled": false` as the
  server-level gate. `development-modes.md` notes the plan-mode rule.

- `getToolMapping()` is memoised plan mode called it on every `getAvailableToolNames()`,
  both for prompt building and per turn at runtime.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Documentation update

## Changeset

### Automated Tests

- [x] New features include passing tests in `.spec.ts`/`tsx` files
- [x] All existing tests pass

Regression tests cover both directions of #989 (plan must not run an `alwaysAllow`-ed
mutation; headless must not prompt), the normal-mode boundary (a server's `readOnlyHint`
must not silence its own prompt, while `alwaysAllow` still does), and the annotation path
end to end driven through `connectToServer()` with a stubbed `listTools()` so the
mapping in `mcp-client.ts` actually executes, including absent annotations, annotations
without the hint, and a truthy-but-not-`true` hint.

### Manual Testing

- [ ] Tested with Ollama
- [ ] Tested with OpenRouter

## Checklist

- [x] Self-review completed


will-lamerton (changes requested):
Thanks for this - the diagnosis is right and the direction is correct. MCP shouldn't own a private copy of the approval policy. Two things to address before merge.

**1. `readOnlyHint` now skips approval in normal mode**

`if (readOnly) return false;` is the first branch, so it applies in normal mode too. Any tool whose server sets `readOnlyHint: true` now executes silently where it previously prompted. That contradicts `docs/configuration/mcp-configuration.md:220` ("Tools not listed still require approval"), the same doc the PR cites for the `alwaysAllow` change, and it trusts a server-supplied hint to gate the server itself. The MCP spec treats annotations as untrusted hints for exactly that reason.

None of the bugs being fixed need it. Scoping the short-circuit keeps every fix and leaves normal mode alone:

```ts
const approval: ToolApprovalPolicy = (_args, mode) => {
    if (mode === 'auto-accept' || mode === 'headless') return false;
    if (mode === 'plan') return !readOnly;
    return !isAutoApproved; // normal: alwaysAllow only, as documented
};
```

**2. The annotation plumbing has no coverage**

`MCPClient: readOnlyHint is read from server annotations` re-implements the mapping inside the test and writes the result straight into `serverTools`, so production code never runs. I replaced `mcp-client.ts:172` with `readOnly: undefined` and all 136 tests across both touched specs still passed. The one line that makes the feature work is untested. Could you drive it through `connectToServer` with a stub `listTools` returning `annotations: {readOnlyHint: true}`, or extract the mapping so a test can call it directly?

Smaller items:

- `tool-manager.ts:207` can use the existing `this.isReadOnly(n)` instead of reaching into `this.registry.getEntry(n)?.readOnly`.
- `getToolMapping()` rebuilds a Map on every `getAvailableToolNames` call in plan mode, which runs on both prompt building and per-turn runtime. Worth caching on the client.
- `docs/configuration/mcp-configuration.md` isn't touched, but its "Auto-Approve Tools" bullets are now wrong in two ways: `alwaysAllow` no longer applies in plan, and it doesn't apply in headless either. Worth a line there alongside the `development-modes.md` update.
- Headless now runs every MCP tool unattended with no per-server opt-out, while custom tools in the same filter still need `approval === 'never'`. That matches `execute_bash` and is what #989 asked for, but the asymmetry deserves a mention in the MCP docs.

Verified locally: both touched specs pass (51 + 85), `tsc --noEmit` clean, biome clean on the production files.

addyCooks:
Thanks @will-lamerton, 
all addressed in 0eec51d9.

**1. `readOnlyHint` in normal mode.** Took your predicate as written:

```ts
if (mode === 'auto-accept' || mode === 'headless') return false;
if (mode === 'plan') return !readOnly;
return !isAutoApproved; // normal: alwaysAllow only, as documented
```

That's the whole fix, not half of one: the entry still carries `readOnly`, but rule #4
(function policy) resolves ahead of #5, and MCP always supplies a function so #5 never
fired for MCP in the first place. The PR description was wrong to claim it did; corrected.
A comment now records why the order matters.

**2. Annotation coverage.** You were right that the old test proved nothing. Replaced with
one driven through `connectToServer()` and a stubbed `listTools()`, asserting the flag lands
on the discovered records *and* the registry entries plan mode filters on. Absent
annotations, annotations without the hint, and a truthy-but-not-`true` hint are covered too.
Your probe now fails it:

```
# readOnly: tool.annotations?.readOnlyHint === true  ->  readOnly: undefined
not ok 51 - MCPClient.connectToServer: carries annotations.readOnlyHint onto discovered tools
```

**Smaller items.** `isReadOnly(n)` done. `getToolMapping()` memoised, invalidated on
connect/disconnect; the Map is now shared by reference and documented read-only (all three
callers only `get`/`has`). `mcp-configuration.md` gains a per-mode table `alwaysAllow` is
normal-mode only and can't override plan plus the headless asymmetry, with `"enabled": false`
as the server-level gate.

**CI.** The three failures were the branch sitting 20 commits behind `main`, not the diff
(the VS Code job checks out `head.sha`; the labeler reads `labeler.yml` at `base.sha`; coverage
compared against a newer baseline). Merged `main`, no conflicts all 14 checks green, coverage
92.2% vs the 92.17% baseline.

will-lamerton:
Nice work again @addyCooks :)

addyCooks:
Thanks @will-lamerton !! : )

PR #887, opened 2026-08-17
feat(vscode): add agent action timeline undo

AryanNandanwar:
## Description

Adds an action timeline to the VS Code sidebar chat so users can click a prior mutating tool step and revert workspace files plus conversation history back to that checkpoint. Closes #809.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

## Changeset

- [x] Added a changeset (`pnpm changeset`) describing this change for the changelog

Docs-only or internal chores need no changeset (or run `pnpm changeset --empty` to note that intentionally).

### Automated Tests

- [x] New features include passing tests in `.spec.ts/tsx` files
- [ ] All existing tests pass (`pnpm test:all` completes successfully)
- [x] Tests cover both success and error scenarios

### Manual Testing

- [ ] Tested with Ollama
- [ ] Tested with OpenRouter
- [x] Tested with OpenAI-compatible API
- [ ] Tested MCP integration (if applicable)

## Checklist

- [x] If this was for an open issue, I was assigned to it
- [x] Code follows project style guidelines
- [x] Self-review completed
- [ ] Documentation updated (if needed)
- [x] No breaking changes (or clearly documented)
- [x] Appropriate logging added using structured logging (see [CONTRIBUTING.md](../CONTRIBUTING.md#logging))

PR #881, opened 2026-08-15
feat: add .nanocoderignore file support

A-S-Manoj:
Closes #755

## Description
Adds support for an optional `.nanocoderignore` file at the workspace root. Patterns in it are merged into the same `ignore` instance used for `.gitignore`, additively — nothing already ignored stops being ignored, and this doesn't replace `.gitignore` handling.

This lets you keep files tracked in git (e.g. `package-lock.json`, generated fixtures, `.env`) while still excluding them from what the AI reads, saving tokens and avoiding context bloat, without touching your actual git ignore rules.

`loadGitignore()` in `source/utils/gitignore-loader.ts` was extended to also look for `.nanocoderignore` in `cwd` and add its contents on top of the existing `.gitignore` + default ignores. Function name and return type are unchanged, so all 5 existing call sites work with no changes. Read failures on `.nanocoderignore` fail silently, matching existing `.gitignore` error handling.

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

## Changeset
- [x] Added a changeset (`pnpm changeset`) describing this change for the changelog

### Automated Tests
- [x] New features include passing tests in `.spec.ts/tsx` files
- [x] All existing tests pass (`pnpm test:all` completes successfully)
- [x] Tests cover both success and error scenarios

Added 3 new test cases to `gitignore-loader.spec.ts`: `.nanocoderignore` alone, `.gitignore` + `.nanocoderignore` together, and neither file present (no-op, no throw).

### Manual Testing
- [ ] Tested with Ollama
- [ ] Tested with OpenRouter
- [ ] Tested with OpenAI-compatible API
- [ ] Tested MCP integration (if applicable)

Not applicable — this change is isolated to file-ignore pattern matching and doesn't touch provider/model code paths.

## Checklist
- [x] If this was for an open issue, I was assigned to it
- [x] Code follows project style guidelines
- [x] Self-review completed
- [x] Documentation updated (if needed)
- [x] No breaking changes (or clearly documented)
- [x] Appropriate logging added using structured logging (see CONTRIBUTING.md#logging)

PR #1015, opened 2026-08-28
fix: snapshot files in repositories without commits

### LINKED ISSUE #1010: [Bug] git diff --name-only HEAD fails silently in new git repos
## Description

In a newly initialized Git repository with no commits, the file snapshot service fails silently when trying to capture untracked files.

## Environment

## Steps to Reproduce

1. `mkdir new-repo && cd new-repo`
2. `git init`
3. `touch file.txt`
4. Run nanocoder and perform an action that triggers a file snapshot.

## Expected Behavior

The file snapshot service should capture the state of `file.txt`.

## Actual Behavior

The command `git diff --name-only HEAD` fails (returns `[]` or errors) because `HEAD` doesn't exist yet, causing the snapshot service to silently miss untracked files in brand new projects.

## Additional Context

## Implementation Notes

In `source/services/file-snapshot.ts:75`, detect if the repo has commits (e.g., checking if HEAD exists) and fallback to listing all files in the directory (respecting `.gitignore`) if it's a completely fresh repo.

ethanhawkes-gif:
## Summary

- check work-tree validity separately from commit existence
- skip only the committed-file diff when `HEAD` is absent, then collect
  non-ignored untracked files with `git ls-files --others --exclude-standard`
- update `source/services/file-snapshot.ts` and add the regression case
  `FileSnapshotService returns untracked files from a repository without commits`
  in `source/services/file-snapshot.spec.ts`; the case initializes a temporary
  repository with no commits and verifies root, nested, and ignored files

## Test plan

- `pnpm run test:ava -- source/services/file-snapshot.spec.ts`
- `pnpm run test:types`

Closes #1010


will-lamerton (changes requested):
Thanks for this - the diagnosis is right and the error handling degrades safely. I merged it into `main` locally and confirmed it merges cleanly, typechecks, and passes all 30 tests in `file-snapshot.spec.ts`. A few things before it can go in.

### The fix misses staged files

On an unborn branch we set `modifiedOutput = ''` and rely only on `git ls-files --others --exclude-standard`. `--others` excludes anything in the index, so the fix stops working the moment a user stages. Running the merged code against `git init && git add .`:

```
{"files":[],"truncated":false,"available":true}
```

Zero files, for a repo where the whole tree is staged. `git init` then `git add .` is a very common opening move, and `git checkout --orphan` hits the same path with a fully populated index. That affects `checkpoint-manager.ts:125` and both scans in `acp-timeline.ts`.

`git diff` handles an unborn HEAD natively, so one word covers it:

```ts
modifiedOutput = execSync('git diff --name-only --cached', {...}).trim();
```

### The `show-ref` probe is redundant

`rev-parse --verify HEAD` already answers "does HEAD exist". The `show-ref` probe only exists to work out *which* command in the inner `try` threw, and it only needs to because `git diff` is inside that same `try`. `throw headError` on line 131 is also thrown inside the try whose own `catch (showRefError)` then catches and re-inspects it. It happens to rethrow correctly today (both commands exit 128), but only by accident - any git command that ever exits 1 there would be silently treated as an unborn branch.

Moving `git diff` out removes the whole nested-catch structure:

```ts
let hasHead = true;
try {
	execSync('git rev-parse --verify HEAD', {cwd: this.workspaceRoot, stdio: ['pipe', 'pipe', 'pipe']});
} catch {
	hasHead = false;
}

// An unborn branch has no HEAD to diff against, but its index can still be
// full (`git init && git add .`), so diff the index instead of skipping.
const modifiedOutput = execSync(
	hasHead ? 'git diff --name-only HEAD' : 'git diff --name-only --cached',
	{cwd: this.workspaceRoot, encoding: 'utf-8', stdio: ['pipe', 'pipe', 'pipe']},
).trim();
```

A real `git diff` failure then falls straight to the outer catch, which is the old behaviour the comment says it wants to preserve.

### Smaller points

- `git rev-parse --is-inside-work-tree` never checks its output - in a bare repo it prints `false` and exits 0, so the comment overstates what it does. With the simplification above it is fully redundant with the existing error handling, and dropping it takes the per-scan subprocess count back from 4 to 2 (`getModifiedFilesResult` runs twice per opaque tool call in `acp-timeline`).
- No changeset. `changeset-check` is deliberately non-blocking, so its green tick is not evidence one exists. This is a user-facing fix, so please run `pnpm changeset`.
- The new test is a genuine regression test (it fails without the fix), but please add the staged-file case above, and call `getModifiedFilesResult()` so it can assert `available: true` - that flag is what `acp-timeline.ts` branches on for unborn repos.
- Nit: `import {execSync} from 'child_process'` sits after the `fs` import; `file-snapshot.ts` itself orders `child_process` first.

ethanhawkes-gif:
All six points addressed in `1c72787`. The branch was 109 commits behind `main`, so I rebased first — that turned out to matter for both CI reds.

**Index-diff instead of the work-tree probe (1–3).** `getModifiedFilesResult` now checks `git rev-parse --verify HEAD` and diffs `git diff --name-only --cached` when there is no HEAD, in the shape you suggested. Two notes where the rebase moved things: the `--is-inside-work-tree` probe you asked me to drop was already removed on `main`, and `getModifiedFiles` has since been refactored into a thin wrapper, so the change sits in `getModifiedFilesResult`. The nested `show-ref` catch is gone, so a genuine `git diff` failure falls to the outer catch and still reports git as unavailable. Two subprocesses per scan plus the existing `ls-files`.

**Staged-file test (4).** Added alongside the untracked one, asserting `available` and the file list. I wrote it before the source change to confirm it bites: against the old code it fails with `available: false` and an empty list — your reported `{"files":[],...}` — and passes after. `--others` alone would not have covered this, since `git ls-files --others` excludes anything already in the index, which is exactly the `git init && git add .` case.

**Changeset (5)** is `.changeset/snapshot-files-without-commits.md`; `pnpm test:changesets` passes. **Import order (6)** fixed in the spec — `child_process` above `fs`.

Full suite: 7520 passed, 1 skipped, 0 failed; `test:types` and `test:format` clean.

**On the two red checks — both were stale-base, not this PR.** The Semgrep finding is `plugins/vscode/src/chat-webview-provider.ts`, which this PR never touches; it was fixed on `main` by [`e2770f1`](https://github.com/Nano-Collective/nanocoder/commit/e2770f1866e78279d5bb472bc351b55ea3193731) (#887) at 2026-08-28 10:33 +0100, and this branch was forked from [`b448f2b`](https://github.com/Nano-Collective/nanocoder/commit/b448f2b60e3e94295767b7d42ad7c34d35608e76) (#881) at 10:09 the same morning, 24 minutes earlier. The 51 unit-test failures were all in `source/vscode/chat-panel-*.spec.ts` with none in file-snapshot; those specs pass locally on the rebased branch. The rebase should clear both, but worth a fresh CI run to confirm rather than taking my word for it.

will-lamerton:
Thanks for this PR @ethanhawkes-gif - feel free to add yourself as a contributor to our website via a PR which I will approve :)

PR #1017, opened 2026-08-28
fix: derive fetch url truncation label

mrinalg297:
## Description

Fix the hardcoded `100KB` truncation warning in `fetch-url` by deriving the displayed size from `MAX_URL_CONTENT_BYTES` using the existing `formatSize` utility. Also updates the related formatter test to verify the dynamic value.

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Changeset

### Automated Tests

- [] New features include passing tests in `.spec.ts/tsx` files

### Manual Testing

- [ ] Tested with Ollama
- [ ] Tested with OpenRouter

## Checklist

- [x] Self-review completed

### DIFF
--- a/.changeset/big-states-wink.md
+++ b/.changeset/big-states-wink.md
@@ -1,0 +1,6 @@
+---
+"@nanocollective/nanocoder": patch
+---
+
+Fix fetch_url truncation warning to display the content limit in characters.
+  
--- a/source/tools/fetch-url.spec.tsx
+++ b/source/tools/fetch-url.spec.tsx
@@ -3,8 +3,6 @@ import {render} from 'ink-testing-library';
 import React from 'react';
 import {themes} from '../config/themes';
 import {ThemeContext} from '../hooks/useTheme';
-import {MAX_URL_CONTENT_BYTES} from '../constants';
-import {formatSize} from '../components/file-explorer/utils';
 
 console.log(`\nfetch-url.spec.tsx – ${React.version}`);
 
@@ -327,8 +325,7 @@ test('formatter shows truncation warning when content is truncated', t => {
 
 	const output = lastFrame();
 	t.truthy(output);
-	const expectedLabel = `Content was truncated to ${formatSize(MAX_URL_CONTENT_BYTES)}`;
-	t.regex(output!, new RegExp(expectedLabel.replace('.', '\\.')));
+	t.true(output!.includes('Content was truncated to 100,000 characters'));
 });
 
 test('formatter renders without result (before execution)', t => {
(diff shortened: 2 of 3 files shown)


will-lamerton (changes requested):
Thanks for this. The instinct is right, the label shouldn't be hardcoded separately from the constant, but the chosen formatter changes the user-visible string in a way I don't think we want.

**1. The rendered label becomes `97.7 KB`**

`MAX_URL_CONTENT_BYTES` is `100_000` (decimal, per the `// ~100 KB` comment), but `formatSize` divides by 1024, so the warning now reads `⚠ Content was truncated to 97.7 KB`. Users never configured 97.7 KB, and the one-decimal figure implies a precision we don't have: the truncation at `fetch-url.tsx:39` compares `content.length` (UTF-16 code units), not bytes.

Two options that both keep the value derived from the constant:

- Say what the code actually measures: `Content was truncated to ${MAX_URL_CONTENT_BYTES.toLocaleString()} characters`. This also lines up with the handler's own note (`original size was ${content.length} characters`), which currently disagrees with the formatter's "KB" wording.
- Or keep KB with the decimal divisor the constant was written in: `${MAX_URL_CONTENT_BYTES / 1000}KB`, which renders the same `100KB` as today.

**2. The updated test no longer asserts anything**

```ts
const expectedLabel = `Content was truncated to ${formatSize(MAX_URL_CONTENT_BYTES)}`;
t.regex(output!, new RegExp(expectedLabel.replace('.', '\\.')));
```

This builds the expectation from the same constant and the same function the source uses, so it passes for any value of either, including a broken `formatSize`. The old hardcoded assertion was the stronger test because it pinned the user-visible output. Please assert the literal expected string so a change to the constant surfaces as a failure someone has to consciously accept.

Also, `String.replace` with a string pattern only replaces the first match, so the escaping is incomplete by construction (it works here only because there's exactly one dot). Since no pattern matching is needed, `t.true(output!.includes(expectedLabel))` is simpler and correct.

**3. Missing changeset**

The PR body has the changeset box checked, but the diff is only the two source files. This is a user-visible string change, so it needs one.

Happy to merge once the label stays round and honest about its unit, the test pins the literal output, and the changeset is added.

mrinalg297:
Sorry for the delay,
I've pushed updates addressing all three points:

- Label now reflects what the code actually measures — `Content was truncated to 100,000 characters` (via `MAX_URL_CONTENT_BYTES.toLocaleString()`), dropping the misleading 97.7 KB.
- The regression test now pins the literal string (`t.true(output!.includes('Content was truncated to 100,000 characters'))`) instead of re-deriving it from `formatSize()`, so it'll actually catch a future regression.
- Added the changeset (`.changeset/big-states-wink.md`).

Format, lint, types, and the full `fetch-url.spec.tsx` suite (22 tests) all pass locally. Let me know if there's anything else you'd like changed!

PR #998, opened 2026-08-28
fix(input): stop placeholder ids from being recycled and clobbering live placeholders

arpan7sarkar:
## Description

Closes #969 

Paste placeholder ids were numbered from the count of live entries (`existingPasteCount + 1`), so deleting one freed its id for reuse.
Paste twice, delete the first, paste again: the third paste got id `2` and overwrote the second paste, destroying its content and leaving two chips labelled `#2` - one of which reached the model as literal text.

Ids are now namespaced by type (`paste_1`, `file_1`) and allocated from the highest id ever used, so a deletion can never free one.
Placeholder lookup matches on each entry's own display text instead of a paste-shaped regex, which also makes `@file` mentions delete atomically instead of leaving an orphan entry.

Note: the reported paste/`@file` collision was never possible - mentions have been keyed `file_${n}` since the feature shipped.
The symptom was real but came from reuse within the paste keyspace.

Also included: a 9-line fix for 51 pre-existing failing specs in `source/vscode/`, where the harness never loaded the new `uri-utils.js` / `slash-command-utils.js` webview helpers.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Changeset

- [x] Added a changeset (`pnpm changeset`) describing this change for the changelog

### Automated Tests

- [x] New features include passing tests in `.spec.ts/tsx` files
- [x] All existing tests pass (`pnpm test:all` completes successfully)
- [x] Tests cover both success and error scenarios

`pnpm test:ava` - 6664 passed, 0 failed. Types, lint, format and knip clean; `test:audit` and `test:security` need network and were left to CI.

### Manual Testing

- [ ] Tested with Ollama
- [ ] Tested with OpenRouter
- [x] Tested with OpenAI-compatible API
- [ ] Tested MCP integration (if applicable)

No provider round-trip is involved - the defect lives in input state, before anything is sent.
To check in the TUI: paste a multi-line block twice, backspace once into the first chip, then paste again. Before, two chips read `#2`; after, `#2` and `#3`.

## Checklist

- [x] If this was for an open issue, I was assigned to it
- [x] Code follows project style guidelines
- [x] Self-review completed
- [ ] Documentation updated (if needed)
- [x] No breaking changes (or clearly documented)
- [x] Appropriate logging added using structured logging (see [CONTRIBUTING.md](../CONTRIBUTING.md#logging))

PR #1020, opened 2026-08-28
Fix/atomic deletion boundary

### LINKED ISSUE #977: [Bug] Atomic-deletion boundary off-by-one
## Description
In `source/utils/atomic-deletion.ts`, boundaries are checked with inconsistent inclusiveness (`<=` vs `<`). This results in an off-by-one error at the boundary limits.

## Environment

## Steps to Reproduce
1. Trigger an atomic deletion on the exact boundary edge of a placeholder.
2. Observe incorrect character trimming.

## Expected Behavior
Boundaries should be checked consistently (both strict `<` or both `<=`).

## Actual Behavior
One side uses `<=` and the other strict `<`, causing subtle boundary bugs.

## Logs/Screenshots
N/A

## Additional Context

hiarun02:
## Description

Fixes #977

Consolidates atomic deletion overlap checks around an explicit half-open `[start, end)` range helper and clarifies cursor boundary behavior for placeholder lookup. The deletion behavior is preserved, while the cursor-position helper now treats the position before a placeholder as outside and the backspace position immediately after it as inside.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Changeset

### Automated Tests

Ran:

```bash
./node_modules/.bin/ava source/utils/atomic-deletion.spec.ts
./node_modules/.bin/biome check source/utils/atomic-deletion.ts source/utils/atomic-deletion.spec.ts
./node_modules/.bin/tsc --noEmit

### DIFF
--- a/source/utils/atomic-deletion.spec.ts
+++ b/source/utils/atomic-deletion.spec.ts
@@ -106,6 +106,42 @@ test('handleAtomicDeletion returns null for normal deletions', t => {
 	t.is(result, null);
 });
 
+test('handleAtomicDeletion ignores deletion immediately before placeholder', t => {
+	const previousState: InputState = {
+		displayValue: 'x[Paste #123: 100 chars] more',
+		placeholderContent: {
+			'123': {
+				type: PlaceholderType.PASTE,
+				displayText: '[Paste #123: 100 chars]',
+				content: 'content',
+				originalSize: 100,
+			} as PastePlaceholderContent,
+		},
+	};
+
+	const result = handleAtomicDeletion(previousState, '[Paste #123: 100 chars] more');
+
+	t.is(result, null);
+});
+
+test('handleAtomicDeletion ignores deletion immediately after placeholder', t => {
+	const previousState: InputState = {
+		displayValue: '[Paste #123: 100 chars]x more',
+		placeholderContent: {
+			'123': {
+				type: PlaceholderType.PASTE,
+				displayText: '[Paste #123: 100 chars]',
+				content: 'content',
+				originalSize: 100,
+			} as PastePlaceholderContent,
+		},
+	};
+
+	const result = handleAtomicDeletion(previousState, '[Paste #123: 100 chars] more');
+
+	t.is(result, null);
+});
+
 test('handleAtomicDeletion returns null for additions', t => {
 	const previousState: InputState = {
 		displayValue: 'Short text',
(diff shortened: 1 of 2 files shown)


will-lamerton (changes requested):
Thanks for picking this one up. The `rangesOverlap` consolidation is a good change and I'd like to take it, but a few things need sorting first.

**The branch is based on a stale `main` and currently conflicts.** `main` rewrote this file in #998 (cad13a0c): it now iterates `findPlaceholderOccurrences(text, placeholderContent)` and destructures `{id, start, end}` instead of using inline regex matches, and both `findPlaceholderAtPosition` and `wouldPartiallyDeletePlaceholder` take a third `placeholderContent` argument. The new tests call them with 2 and 3 args, so they won't typecheck after a rebase. CI is green only because it ran against the head commit's own base. Please rebase onto `main` and port the change to the post-#998 shape.

**Two of the three changes are behavioural no-ops.** I brute-forced the old and new predicates across all `(placeholderStart, placeholderEnd, deletionStart, deletionEnd)` combinations and got zero divergences for both the `handleAtomicDeletion` 3-clause check and the `wouldPartiallyDeletePlaceholder` 4-clause `hasOverlap`. The old forms were already correct half-open tests, just written verbosely (`deletionEnd > start && deletionEnd <= end` is only reachable when `deletionEnd > start`, so the `<=` never widened anything). The three tests covering those two functions therefore pass unchanged against `main` today. That's fine as a regression guard for the refactor, but the PR description claims deletions touching the character before/after a placeholder are "no longer" treated as overlapping, and that already held. Could you reword the description so it doesn't land in the changelog as a fix that wasn't there?

**The one real change needs a semantics decision.** The only line that alters behaviour is `position <= end` -> `position < end` in `findPlaceholderAtPosition`. That function and `wouldPartiallyDeletePlaceholder` have no production callers (grep shows only `useInputState.ts:72` calling `handleAtomicDeletion`; the other two appear only in the spec), so the change is currently unobservable and the direction should be argued rather than asserted. If `position` is a character index, `< end` is right. If it's a cursor position, which is what the doc comment says and what an Ink text input actually gives you, then `position === end` is the cursor sitting immediately after the placeholder, exactly where Backspace lands and exactly the case you'd want reported as "on placeholder X" - and `>= start` already treats the mirror-image position before the placeholder as inside. So `< end` arguably makes it asymmetric in the other direction rather than consistent.

Either keep `<= end` and fix the start side for cursor semantics, or land `< end` with a doc comment stating the index convention explicitly. Worth also considering whether these two exports should just be deleted; knip won't flag them because the spec imports them, and if they're vestigial that resolves #977 more decisively than adjusting the boundary math of code nothing calls.

Minor: `rangesOverlap` should get a one-line JSDoc stating the half-open `[start, end)` convention, since that convention is the whole point of the PR. And no changeset - `changeset-check` passed so it's clearly not enforced here, but if the boundary change is user-visible please add one, otherwise `pnpm changeset --empty` records that it's internal.

PR #1030, opened 2026-08-29
fix: keep custom-tool cwd inside the project

Dhirenderchoudhary:
## Description

Custom tool `cwd` was used as-is whenever the path existed. A symlink, an absolute path, or `${HOME}` could put the shell outside the project with no warning.

`resolveCwd` now realpath-checks against the project root, same idea as the file tools. If the real path is outside (or missing), it falls back to the project root.

Closes #1027.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Changeset

- [x] Added a changeset (`pnpm changeset`) describing this change for the changelog

Docs-only or internal chores need no changeset (or run `pnpm changeset --empty` to note that intentionally).

### Automated Tests

- [x] New features include passing tests in `.spec.ts/tsx` files
- [ ] All existing tests pass (`pnpm test:all` completes successfully)
- [x] Tests cover both success and error scenarios

### Manual Testing

- [ ] Tested with Ollama
- [ ] Tested with OpenRouter
- [ ] Tested with OpenAI-compatible API
- [ ] Tested MCP integration (if applicable)

N/A this is cwd resolution before spawn. Covered by the unit tests above.

## Checklist

- [ ] If this was for an open issue, I was assigned to it
- [x] Code follows project style guidelines
- [x] Self-review completed
- [ ] Documentation updated (if needed)
- [x] No breaking changes (or clearly documented)
- [ ] Appropriate logging added using structured logging (see [CONTRIBUTING.md](../CONTRIBUTING.md#logging))

No extra logging  missing cwd already fell back quietly, same for an out-of-project cwd.

PR #1031, opened 2026-08-29
fix: spawn cmd.exe with /c for custom tools on Windows

### LINKED ISSUE #1028: [Bug] Windows: custom tools spawn cmd.exe with -c; execute_bash already uses /c
## Description

Custom tools always spawn the shell with `-c`:

```ts
// source/custom-tools/handler.ts:56 (main @ 1d0cf103, v1.30.0)
const child = spawn(options.shell, ['-c', script], {
```

On Windows, `pickShell` in the same file (around 175-180) returns `process.env.ComSpec` / `cmd.exe` when `shell:` is not set.

`cmd.exe` does not take `-c`. It takes `/c`.

`execute_bash` already does this right in `source/services/bash-executor.ts:79-80`:

```ts
const proc = isWindows
	? spawn('cmd', ['/c', command], {cwd})
	: spawn('sh', ['-c', spawnCommand], {cwd, detached: true});
```

Custom tools never copied that branch.

`handler.spec.ts` always passes `shell: '/bin/sh'`, so Linux/mac CI never hits this.

`shell: bash` in frontmatter becomes `/bin/bash` on every OS. Fine on mac/Linux. Usually missing on Windows.

Docs say `shell: bash | sh`. They do not mention cmd, but that is what Windows actually gets.

## Environment

- **OS**: Windows 11 (the bug is the spawn argv; mac/Linux `-c` is fine)
- **Node version**: N/A
- **Nanocoder version**: 1.30.0 (main @ 1d0cf103)
- **Provider**: N/A
- **Model**: N/A

## Steps to Reproduce

1. On Windows, `/tools create hello` with body `echo hi` and no `shell:` key
2. Call the tool
3. Compare with `execute_bash` running `echo hi` on the same machine

## Expected Behavior

Same Windows spawn as execute_bash: `cmd /c <script>`. POSIX stays `-c`.

## Actual Behavior

Custom tools spawn `cmd.exe -c ...`. `-c` is not a cmd switch.

execute_bash on the same machine: `cmd /c echo hi`

## Logs/Screenshots

Did not run cmd.exe on mac. The argv is hardcoded at handler.ts:56 and pickShell returns cmd.exe on win32.

## Additional Context

Copy the `isWindows ? ['/c', script] : ['-c', script]` split from bash-executor.ts. No need for PowerShell in v1.

A table test would let Linux CI assert the Windows argv.

I can take the PR if this gets assigned.

Dhirenderchoudhary:
## Description

Custom tools always spawned the shell with `-c`. On Windows, `pickShell` returns ComSpec/`cmd.exe`, and cmd does not take `-c` — it takes `/c`. `execute_bash` already does this.

`shellArgs` now keys off the shell **basename**, not `platform === 'win32'`. Frontmatter `shell: bash|sh` still returns a POSIX path on Windows; a platform check would have handed `/c` to `/bin/sh`. `cmd` / `cmd.exe` (including a full Windows path) get `/d /s /c` (`/d` skips AutoRun, `/s` makes quote stripping deterministic). bash/sh keep `-c`.

`{{ }}` substitution stays POSIX-quoted and is **not** shell-safe under cmd. Per-shell quoting in `renderValue` is tracked in #1084.

`shell: bash` / `shell: sh` still spawn `/bin/bash` or `/bin/sh` on Windows and throw `Custom tool failed to start` if those binaries are missing. Pre-existing; #1028 is not fully closed for that path.

Closes #1028.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Documentation update

## Changeset

### Automated Tests

`shellArgs` plus a `runScript` spawn test that a binary named `cmd.exe` receives `/d /s /c`.

### Manual Testing

N/A spawn argv only.

## Checklist

- [x] Self-review completed
- [ ] Appropriate logging added using structured logging (see CONTRIBUTING.md)

No extra logging. POSIX spawn is unchanged (`-c`).


will-lamerton (changes requested):
Thanks for this - the fix itself is right, and branching on the shell basename instead of `platform === 'win32'` is a genuine improvement over what the issue proposed, since `shell: bash|sh` in frontmatter returns a POSIX path on Windows too and a platform check would have handed `/c` to `/bin/sh`. Worth calling that out in the description.

Requesting changes for the merge conflict plus two things that should land with it.

**1. Merge conflict (blocking).** The branch is `CONFLICTING` against `main`. The only real conflict is the import block at the top of `source/custom-tools/handler.spec.ts` - #1030 landed `symlinkSync`, `resolve`, and `type ExecutionContext` there. `handler.ts` merges clean. A rebase on `main` keeping both import sets resolves it.

**2. This makes a previously-inert path live, and the templating is POSIX-only.** `source/custom-tools/template.ts` quotes every substituted value with `shellQuote()`, which is explicitly POSIX single-quoting. cmd.exe does not treat `'` as a quote character at all. Before this PR `cmd.exe -c ...` did nothing, so the mismatch was unreachable; after it, on Windows:

- Correctness: a body like `type {{ file }}` reaches the program as `type 'notes.txt'` and fails to find the file.
- Security: LLM-supplied values are no longer neutralised. `&`, `|`, `>`, `^` in a value break out of the intended command under cmd, and `%VAR%` expands. That is exactly what `shellQuote` prevents on POSIX, and there is a spec asserting it.

Still a strict improvement over "custom tools silently do nothing on Windows", so I am not asking for cmd-flavoured quoting in this PR. But it should be a deliberate call rather than an unnoticed side effect: please add a line to the changeset and to `docs/features/custom-tools.md` noting that `{{ }}` substitution is POSIX-quoted and not shell-safe under cmd, and open a follow-up issue for per-shell quoting in `renderValue`.

**3. Docs line is now wrong.** `docs/features/custom-tools.md:93` says `shell: bash | sh  # default: bash if available, else sh`. On Windows the default is ComSpec/cmd.exe. This PR is the natural place to fix that one line.

Non-blocking notes:

- Consider `['/d', '/s', '/c', script]` rather than bare `/c`. `/d` suppresses the `AutoRun` registry key, which otherwise runs an arbitrary user-set command before every `cmd /c`, and `/s` makes quote stripping deterministic. Node does exactly this for its own `shell: true` path.
- Related: `spawn` applies MSVCRT quoting to `script` (wrapping in `"`, backslash-escaping inner quotes) and cmd does not understand `\"`, so a body containing `echo "hi"` reaches cmd with stray backslashes. `bash-executor.ts:80` has the same wart, so matching it is defensible here, but a shared helper used by both would beat a second copy.
- `shell: bash` on Windows still spawns `/bin/bash` and throws "Custom tool failed to start". Pre-existing and mentioned in #1028 - fine to leave, but say so in the PR so it is not assumed closed.
- Test covers the pure function only; `runScript` actually passing `shellArgs(...)` to `spawn` is not asserted. Acceptable, but worth one if the argv gains `/d /s` or `windowsVerbatimArguments`.
- Nit: the file groups exported helpers above the private `pickShell`; `shellArgs` landing after it reads slightly out of order.

Mechanics look good otherwise - `isWindowsCmd` correctly handles the full ComSpec path, `.exe`, and case, `replaceAll` is fine at ES2022, and the changeset is scoped correctly as a patch.

will-lamerton (changes requested):
Thanks for the follow-up commit - the code side of this is fully addressed. The conflict is resolved and the branch merges clean against current `main`, the `/d /s /c` argv landed with a test that actually asserts `runScript` forwards it, the docs and changeset state the POSIX-quoting caveat plainly, docs line 93 is corrected, and `shellArgs` is grouped with the other exported helpers. CI is green.

Two bits of bookkeeping left before I approve:

**1. Open the per-shell quoting follow-up issue.** `docs/features/custom-tools.md` now says "Per-shell quoting in `renderValue` is a follow-up", but nothing tracks it. That issue was the condition for me not asking for cmd-flavoured quoting in this PR - without it the concession quietly becomes a permanent gap. Please open it and link it from both the PR body and that docs line.

**2. Refresh the PR description.** It still says `shellArgs` passes `/c`, which is no longer accurate now that it passes `/d /s /c`. It also does not mention the basename-over-`platform === 'win32'` rationale, or that `shell: bash` on Windows still spawns `/bin/bash` and throws "Custom tool failed to start". Worth stating the latter so #1028 is not assumed fully closed once this merges.

No code changes needed.

will-lamerton:
Great work as always @Dhirenderchoudhary :)

Issue #1041, opened 2026-08-29
[Bug] message-compression.ts aggressively false-positives "resolved" errors


akramcodez:
## Description

When trying to determine if an error mentioned in the chat has been "resolved", the message compression logic searches the entire conversation content for words like `fixed|resolved|success|working`. Because it checks the whole content rather than just the text after the error occurred, unrelated text (e.g. "I fixed a typo earlier") will falsely flag the new error as resolved.

## Environment

- **OS**: Any
- **Node version**: Node 18+
- **Nanocoder version**: 1.30.0

## Steps to Reproduce

1. Have a conversation where you mention the word "success".
2. Trigger an unrelated terminal error later in the same message block.
3. The compression utility will falsely label the new error as "resolved".

## Expected Behavior

Resolution detection should only look at text *after* the error match index.

## Actual Behavior

It runs `/fixed|resolved|success|working/i.test(content)`, scanning the whole string.

## Additional Context

- [x] I have searched existing issues to ensure this is not a duplicate
- [x] I can reproduce this issue consistently
- [x] This issue occurs with the latest version of nanocoder

## Implementation Notes

Fix in `source/utils/message-compression.ts:306`. Scope the regex test to `content.slice(content.indexOf(match[0]) + match[0].length)`.

akramcodez:
sure @mrprohack

Issue #1040, opened 2026-08-29
[Bug] DaemonIpcClient.request leaks memory and hangs if socket.write throws synchronously


akramcodez:
## Description

When communicating with the background daemon over IPC, `DaemonIpcClient.request` adds the request promise to a pending map and then calls `socket.write()`. If the socket is already closed or in an error state, `write()` throws synchronously. Because there is no `try/catch`, the promise is never resolved/rejected and the slot leaks in the map forever.

## Environment

- **OS**: Any
- **Node version**: Node 18+
- **Nanocoder version**: 1.30.0

## Steps to Reproduce

1. Force a daemon crash or close the IPC socket prematurely.
2. Attempt an IPC request right as the socket closes.
3. `write()` throws, and the pending promise sits in memory forever.

## Expected Behavior

If `write()` fails synchronously, the promise should be rejected and removed from the `pending` map.

## Actual Behavior

The throw bypasses the promise resolution, leaking the map entry.

## Additional Context

- [x] I have searched existing issues to ensure this is not a duplicate
- [x] I can reproduce this issue consistently
- [x] This issue occurs with the latest version of nanocoder

## Implementation Notes

Wrap `this.socket?.write()` in a `try/catch` in `source/daemon/ipc.ts:248-251`. In the `catch` block, call `reject(error)` and `this.pending.delete(id)`.

AniketR10:
@akramcodez please assign

PR #1043, opened 2026-08-29
fix(daemon): stream the log tail instead of buffering the whole file

tisankanj:
## Description

`nanocoder daemon logs` returns the last 64KB of the daemon log, but it read the whole file first:

```ts
const start = Math.max(0, size - 64 * 1024);
const buf = await readFile(logPath, 'utf-8');
return {exitCode: 0, output: buf.slice(start)};
```

A daemon left running long enough to grow a 1GB log made the command allocate 1GB of heap to throw almost all of it away.

It now reads only the tail through `createReadStream(logPath, {start})`.

## A second defect this also fixes

`size` comes from `statSync`, so it is a **byte** count, but `slice(start)` ran on a **decoded string**, so it is a character offset. The two only agree for pure ASCII. Any multi-byte content in the log shifted the window and returned far less than the intended 64KB, and a large enough log could return almost nothing.

Reading from a byte offset removes the mismatch. Because a byte offset can land inside a character or mid-line, the tail resumes at the next newline, which keeps a partial character out of the output and starts on a whole log line.

## Type of Change

- [x] Bug fix

## Tests

Added `source/daemon/cli.spec.ts` with 5 AVA cases: missing log, a log smaller than the window, tail-only for a large log, multi-byte content, and line-boundary alignment.

Two of them fail against the previous implementation, which is how I checked they are worth having:

```text
# against the original readFile + slice
  ✔ logs reports when no daemon log exists
  ✔ logs returns the whole log when it is smaller than the tail window
  ✔ logs returns only the tail of a large log
  ✘ [fail]: logs keeps the tail intact when the log holds multi-byte characters
  ✘ [fail]: logs starts the tail on a line boundary
  2 tests failed
```

Worth noting: my first version of the multi-byte test passed against the buggy code, so it was not actually testing anything. It only became a real test once it asserted the returned tail is close to a full window rather than a sliver.

## Commands run

```text
$ pnpm run test:all
✅ AVA tests passed
✅ Knip check passed
✅ Audit passed
✅ Everything passes!
```

`pnpm run build` is needed before `test:all` on a fresh clone, otherwise the 7 `cli-integration.spec.ts` cases fail with `MODULE_NOT_FOUND` because they spawn `dist/cli.js`. That is not related to this change.

Not run: Semgrep, which is not installed on this machine, and the manual provider testing in CONTRIBUTING, since this path does not touch an LLM provider.

## Related

`#1040` and `#1041` are the other two open bugs in this batch and are still unclaimed. I left them alone to keep this PR to one problem.

---
<table><tr>
<td></td>
Developed and verified</td>
</tr></table>

### DIFF
--- a/.changeset/daemon-logs-stream-tail.md
+++ b/.changeset/daemon-logs-stream-tail.md
@@ -1,0 +1,5 @@
+---
+"@nanocollective/nanocoder": patch
+---
+
+Fixed `nanocoder daemon logs` reading the whole log file into memory to return its last 64KB, so a long-running daemon with a large log made the command allocate the entire file and stall. The tail is now streamed from a byte offset, which also corrects a byte offset applied to a decoded string: any multi-byte content in the log shifted the window and returned far less than the intended 64KB. Closes #1042.
--- a/source/daemon/cli.spec.ts
+++ b/source/daemon/cli.spec.ts
@@ -1,0 +1,93 @@
+import {mkdir, mkdtemp, rm, writeFile} from 'node:fs/promises';
+import {tmpdir} from 'node:os';
+import {join} from 'node:path';
+import test from 'ava';
+import {runDaemonCli} from './cli';
+
+console.log(`\ncli.spec.ts`);
+
+const TAIL_BYTES = 64 * 1024;
+
+async function tempProject(logContent?: string): Promise<string> {
+	const root = await mkdtemp(join(tmpdir(), 'daemon-cli-'));
+	await mkdir(join(root, '.nanocoder'), {recursive: true});
+	if (logContent !== undefined) {
+		await writeFile(join(root, '.nanocoder', 'daemon.log'), logContent, 'utf-8');
+	}
+	return root;
+}
+
+test.serial('logs reports when no daemon log exists', async t => {
+	const root = await tempProject();
+	try {
+		const result = await runDaemonCli('logs', {projectRoot: root});
+		t.is(result.exitCode, 0);
+		t.is(result.output, 'No daemon log yet.');
+	} finally {
+		await rm(root, {recursive: true, force: true});
+	}
+});
+
+test.serial('logs returns the whole log when it is smaller than the tail window', async t => {
+	const content = 'first line\nsecond line\n';
+	const root = await tempProject(content);
+	try {
+		const result = await runDaemonCli('logs', {projectRoot: root});
+		t.is(result.exitCode, 0);
+		t.is(result.output, content);
+	} finally {
+		await rm(root, {recursive: true, force: true});
+	}
+});
+
+test.serial('logs returns only the tail of a large log', async t => {
+	const line = `${'x'.repeat(99)}\n`;
+	const lines = Math.ceil((TAIL_BYTES * 3) / line.length);
+	const content = `${Array.from({length: lines}, (_, i) => `${i} ${line}`).join('')}last line\n`;
+	const root = await tempProject(content);
+	try {
+		const result = await runDaemonCli('logs', {projectRoot: root});
+		t.is(result.exitCode, 0);
+		t.true(result.output.endsWith('last line\n'));
+		t.true(Buffer.byteLength(result.output, 'utf-8') <= TAIL_BYTES);
+		t.true(result.output.length < content.length);
+		t.false(result.output.startsWith('0 '));
+	} finally {
+		await rm(root, {recursive: true, force: true});
+	}
+});
+
+test.serial('logs keeps the tail intact when the log holds multi-byte characters', async t => {
+	const line = `${'é'.repeat(60)}\n`;
+	const lines = Math.ceil((TAIL_BYTES * 2) / Buffer.byteLength(line, 'utf-8'));
+	const content = `${Array.from({length: lines}, () => line).join('')}last line\n`;
+	const root = await tempProject(content);
+	try {
+		const result = await runDaemonCli('logs', {projectRoot: root});
+		t.is(result.exitCode, 0);
+		t.true(result.output.endsWith('last line\n'));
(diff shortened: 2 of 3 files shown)


will-lamerton (changes requested):
Thanks for this - both defects are real and the streaming fix is the right direction. Two things to sort before merge.

**1. The newline realignment can return an empty string**

`tail.indexOf('\n')` finds the first newline anywhere in the 64KB, not one near the start. If the window has no interior newline, the only match is the trailing one and `slice(newline + 1)` returns `''`. `cli.tsx` guards with `if (result.output) console.log(result.output)`, so `daemon logs` prints nothing at all - worse than the sliver the old code returned.

Reproduced against this branch's implementation:

```
# log = 'é'.repeat(100000) + '\n'   (200001 bytes)
output is '' (first codepoint: undefined)
```

Not exotic for a daemon log, which is the process's raw stdout/stderr: one long serialized payload or a single-line JSON error is enough. Any case where the first newline sits late in the window truncates proportionally.

Suggested shape - align to a line boundary only when one is actually near the start, otherwise trim just the partial UTF-8 character. Working on the Buffer also avoids materializing U+FFFD only to slice it away:

```ts
let buf = Buffer.concat(chunks);
if (start === 0) return {exitCode: 0, output: buf.toString('utf-8')};
const nl = buf.indexOf(0x0a);
if (nl !== -1 && nl < buf.length - 1) {
	buf = buf.subarray(nl + 1);
} else {
	// No usable line break: drop only the partial leading character.
	let i = 0;
	while (i < buf.length && i < 4 && (buf[i] & 0xc0) === 0x80) i++;
	buf = buf.subarray(i);
}
return {exitCode: 0, output: buf.toString('utf-8')};
```

That returns the full 65535-byte window on the same input instead of nothing.

Worth a test, since the existing upper bounds (`Buffer.byteLength(result.output) <= TAIL_BYTES`) are trivially satisfied by an empty string. Pairing each one with a lower bound, as the multi-byte case already does with `bytes > TAIL_BYTES / 2`, would have caught this.

**2. `biome check` fails on the new spec**

`pnpm run test:format` will be red in CI:

```
$ pnpm exec biome format --stdin-file-path=source/daemon/cli.spec.ts < cli.spec.ts
# 3 hunks: two long test titles need the wrapped test.serial(title, fn) form,
# plus the writeFile(...) call in tempProject
```

`source/daemon/ipc.spec.ts` exits 0 through the same command, so it is specific to the new file rather than a config quirk. `pnpm run test:lint:fix` handles it. Flagging because it contradicts the `test:all` output in the description, so it may be worth a re-run.

**Minor, take or leave**

- Off-by-one: if the byte at `start - 1` is already `\n`, the window opens on a complete line and the code still drops it. A log of 64-byte lines returns 1023 lines where 1024 fit.
- `existsSync`/`statSync` are now the only blocking calls left in an otherwise fully async function. Switching to `stat` from `node:fs/promises` and letting `ENOENT` cover the missing-file case would drop the `existsSync` call and its TOCTOU window too.
- `{start, end: size - 1}` would pin the read to the size you measured, in case the daemon appends between the `stat` and the stream open.

Nice catch on the byte-versus-character mismatch, and good instinct calling out that your first multi-byte test passed against the buggy code. Happy to re-review once the empty-output case has a test behind it.

rascal-sl:
Thanks, this was a good catch. Fixed in 6b94212.

## 1. Empty output: confirmed, and it was worse than you described

I wrote your reproduction as a test before touching the code, and it failed exactly as you said. It now passes, along with a second one for the off-by-one.

Against the implementation you reviewed:

```
✘ [fail]: logs keeps the tail when the window holds no line break
✘ [fail]: logs keeps every line when the window opens on a line boundary

  - 1023
  + 1024
```

Your `1023 lines where 1024 fit` was precise, so I took that one too rather than leaving it.

I adopted your Buffer shape. One adjustment: reading the byte before the window is what fixes the off-by-one, since a newline there means the window already opens on a whole line and skipping past it keeps that line. But that extra byte then has to be dropped in the no-break branch, or the output comes back at 65537 bytes and breaks the upper bound. My first attempt did exactly that and the new test caught it, which is a fair advertisement for your point about pairing bounds.

Also took `{start, end: size - 1}`.

## 2. Your point about my bounds was the real lesson

You are right that `bytes <= TAIL_BYTES` is trivially satisfied by an empty string, and that is precisely why my suite went green over a bug that returned nothing. Every size assertion now has a lower bound, including the two existing tests. That is what would have caught this, not more tests.

## 3. `existsSync` / `statSync`: not taken, and here is why

Agreed on the substance. I left it out because it changes the missing-file path's error semantics rather than the tail logic this PR is about, and I would rather not fold an unrelated behaviour change into a fix that is already growing. Happy to do it here if you would prefer it in one go, or as a follow-up.

## 4. `biome check` does not reproduce for me

This one I cannot confirm, and the file has not changed since the commit you reviewed (`6ee87f8`).

Your exact command, on this branch:

```
$ pnpm exec biome format --stdin-file-path=source/daemon/cli.spec.ts < source/daemon/cli.spec.ts
exit=0, output byte-identical to the file

$ pnpm run test:format
$ biome check --no-errors-on-unmatched .
Checked 479 files in 136ms. No fixes applied.
```

Biome 2.5.0, Node 24, macOS. `test:all` is green end to end, so the description was accurate at the time rather than reconstructed. Something differs between our environments. If you still see the three hunks, could you share your `biome --version`? Happy to apply whatever it wants.

## Verification

```
$ npx ava source/daemon/cli.spec.ts
  7 tests passed

$ pnpm run test:all
✅ AVA tests passed
✅ Knip check passed
✅ Audit passed
✅ Everything passes!
```

Semgrep is still not installed here, so that step is unverified on my side.

will-lamerton:
Good to go @rascal-sl :)

Issue #1042, opened 2026-08-29
[Bug] nanocoder logs buffers the entire log file into memory


akramcodez:
## Description

The `logs()` command in the daemon CLI is designed to return the last 64KB of the background daemon log. However, to do this, it calls `fs.readFile()` on the entire file, pulling hundreds of megabytes into memory just to run `.slice(start)` on it.

## Environment

- **OS**: Any
- **Node version**: Node 18+
- **Nanocoder version**: 1.30.0

## Steps to Reproduce

1. Let the daemon run for a long time until the log file is 1GB.
2. Run the command to fetch logs.
3. The Node process will attempt to allocate 1GB of heap memory and crash/stall.

## Expected Behavior

The command should stream only the tail end of the file.

## Actual Behavior

`readFile(logPath, 'utf-8')` buffers the whole file into RAM.

## Additional Context

- [x] I have searched existing issues to ensure this is not a duplicate
- [x] I can reproduce this issue consistently
- [x] This issue occurs with the latest version of nanocoder

## Implementation Notes

Fix in `source/daemon/cli.ts:251-253`. Use `fs.createReadStream(logPath, { start })` or child process `tail -c 65536`.

AniketR10:
@akramcodez please assign me this.

PR #1045, opened 2026-08-29
fix(daemon): release the pending IPC slot when write throws

tisankanj:
## Description

`DaemonIpcClient.request` registers a pending slot and then writes to the socket:

```ts
this.pending.set(id, {resolve, reject});
this.socket?.write(...);
```

When the socket is already closed, `write()` throws synchronously and the slot stays in `pending` for the lifetime of the client. Every failed request adds one more entry that nothing ever removes.

Now the synchronous throw is caught, the slot deleted, and the promise rejected explicitly.

## One correction to the issue

The issue says the promise is never resolved or rejected and that the request hangs. That part does not reproduce. A throw inside a Promise executor rejects that promise, so the caller already got a rejection before this change.

I wrote the regression test first to check, and it separates the two claims cleanly. Against unfixed `main`:

```text
✘ [fail]: request rejects and releases the pending slot when write throws

  - 1
  + 0
```

The rejection assertion in that same test passed. Only `pending.size` was wrong, `1` where `0` was expected. So the leak is real and worth fixing, the hang is not. I kept the explicit `reject` anyway rather than relying on executor semantics, since depending on an implicit throw for control flow is not obvious to the next reader.

## Type of Change

- [x] Bug fix

## Tests

Added one case to `source/daemon/ipc.spec.ts`. It stubs `write` on the connected socket to throw, then asserts the call rejects and that the pending map is empty. It fails on `main` and passes here.

Reading the private `pending` map through a cast is deliberate: the leak has no other observable effect from outside the class.

## Commands run

```text
$ npx ava source/daemon/ipc.spec.ts
  8 tests passed

$ pnpm run test:all
✅ AVA tests passed
✅ Knip check passed
✅ Audit passed
✅ Everything passes!
```

Not run: Semgrep, which is not installed here. `pnpm run build` is needed before `test:all` on a fresh clone, otherwise the `cli-integration.spec.ts` cases fail on `MODULE_NOT_FOUND` because they spawn `dist/cli.js`.

## Note

This branches from `main` and is independent of #1043, which fixes `#1042` in `source/daemon/cli.ts`. The two can merge in either order.

---
<table><tr>
<td></td>
Developed and verified</td>
</tr></table>

### DIFF
--- a/.changeset/ipc-pending-slot-release.md
+++ b/.changeset/ipc-pending-slot-release.md
@@ -1,0 +1,5 @@
+---
+"@nanocollective/nanocoder": patch
+---
+
+Fixed a leaked pending slot in the daemon IPC client. When the socket was already closed, `socket.write()` threw synchronously and the request's entry stayed in the pending map for the lifetime of the client, one per failed request. The slot is now released and the request rejected explicitly. Closes #1040.
--- a/.changeset/ipc-pending-slot-release.md
+++ b/.changeset/ipc-pending-slot-release.md
@@ -2,4 +2,4 @@
 "@nanocollective/nanocoder": patch
 ---
 
-Fixed a leaked pending slot in the daemon IPC client. When the socket was already closed, `socket.write()` threw synchronously and the request's entry stayed in the pending map for the lifetime of the client, one per failed request. The slot is now released and the request rejected explicitly. Closes #1040.
+Fixed a leaked pending slot in the daemon IPC client. If serializing or writing a request threw synchronously, the request's entry stayed in the pending map for the lifetime of the client, one per failed request. The slot is now released and the request rejected explicitly. Closes #1040.
--- a/source/daemon/ipc.spec.ts
+++ b/source/daemon/ipc.spec.ts
@@ -39,6 +39,32 @@ test.serial('ping/pong round-trips through the socket', async t => {
 	}
 });
 
+test.serial('request rejects and releases the pending slot when write throws', async t => {
+	const path = await makeSocketPath();
+	const server = new DaemonIpcServer(path, {
+		listSubscriptions: () => [],
+	});
+	await server.start();
+	const client = new DaemonIpcClient(path);
+	await client.connect();
+	try {
+		const internals = client as unknown as {
+			socket: {write: (payload: string) => boolean};
+			pending: Map<number, unknown>;
+		};
+		internals.socket.write = () => {
+			throw new Error('socket already closed');
+		};
+
+		await t.throwsAsync(client.ping(), {message: 'socket already closed'});
+		t.is(internals.pending.size, 0);
+	} finally {
+		await client.disconnect();
+		await server.stop();
+		await rm(join(path, '..'), {recursive: true, force: true});
+	}
+});
+
 test.serial('listSubscriptions returns server-side list', async t => {
 	const path = await makeSocketPath();
 	const server = new DaemonIpcServer(path, {
--- a/source/daemon/ipc.spec.ts
+++ b/source/daemon/ipc.spec.ts
@@ -52,14 +52,43 @@ test.serial('request rejects and releases the pending slot when write throws', a
 			socket: {write: (payload: string) => boolean};
 			pending: Map<number, unknown>;
 		};
+		const write = internals.socket.write;
 		internals.socket.write = () => {
-			throw new Error('socket already closed');
+			throw new Error('serialization failed');
 		};
 
-		await t.throwsAsync(client.ping(), {message: 'socket already closed'});
-		t.is(internals.pending.size, 0);
+		try {
+			await t.throwsAsync(client.ping(), {message: 'serialization failed'});
+			t.is(internals.pending.size, 0);
+		} finally {
+			internals.socket.write = write;
+		}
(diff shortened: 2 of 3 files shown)


will-lamerton (changes requested):
Thanks for this, and especially for writing the regression test first and correcting the "hang" claim in the issue. I verified it: the test fails on `main` with exactly the `- 1 / + 0` you reported and passes on the branch. The fix itself is correct and idiomatic.

One thing needs changing before merge.

## The changeset states a mechanism that isn't real

Both the issue and the changeset say `write()` throws synchronously when the socket is already closed. It doesn't. On Node 22.15:

- write after `destroy()`: returns `false`, no throw, no error event
- write after `end()`: returns `false`, no throw, async `ERR_STREAM_WRITE_AFTER_END`

`net.Socket.prototype.write` only throws synchronously when the chunk isn't a string or ArrayBufferView. So the repro steps in #1040 never hit this path. The closed-socket case was already safe via `handleClose()`, which rejects and clears every pending slot on the socket's `close` event.

The guard is still worth having, but the realistic sync-throw source in that block is `JSON.stringify` on circular `params`, not `write`. Since the changeset ships verbatim into the published changelog, I'd rather it didn't assert the wrong cause. Suggested wording:

> Fixed a leaked pending slot in the daemon IPC client. If serializing or writing a request threw synchronously, the request's entry stayed in the pending map for the lifetime of the client, one per failed request. The slot is now released and the request rejected explicitly. Closes #1040.

## Suggestions (non-blocking)

**`source/daemon/ipc.ts:248` - the `?.` is now dead and hides the same bug class.** `request()` already throws at line 244 if `!this.socket`, and the executor runs synchronously, so it can't be null here. But if it ever were, `?.` would skip the write and leave the slot pending with no rejection, which is exactly the leak you're fixing. Capturing it first makes the guard total:

```ts
const socket = this.socket;
const id = this.nextId++;
return new Promise((resolve, reject) => {
  this.pending.set(id, {resolve, reject});
  try {
    socket.write(`${JSON.stringify({id, method, params} satisfies IpcRequest)}\n`);
  } catch (error) { ... }
});
```

**Test stub restoration.** Overwriting `internals.socket.write` survives the `finally` only because `Writable.end()` with no chunk doesn't route through the public `write`. Restoring the original in `finally` would decouple the test from that internal.

**Optional second case.** A test that destroys the socket, fires a request, and asserts the promise rejects and `pending` drains via `handleClose` would cover the path the issue actually describes. Nothing in the suite pins that today.

## Adjacent, please don't fix here

Two things I noticed while reading, both pre-existing and better as separate issues if you're interested:

- No request timeout. If the daemon accepts the connection but never responds and never closes the socket, the slot leaks and the promise never settles. That's the real route to the "hangs" symptom in the issue title.
- `connect()` (`source/daemon/ipc.ts:204`) never removes its `s.once('error', reject)` on success. Post-connect, the first socket error is swallowed by a no-op reject and consumes the listener, so a second `error` has no listener and takes the process down.

rascal-sl:
You are right, and I reproduced it rather than taking it on faith. Pushed in 3494f8e.

## The mechanism does not exist

Probed the four cases directly on Node 24.18, so it is not version-specific to your 22.15:

```
write after destroy():        NO SYNC THROW, write() returned false
write after end():            NO SYNC THROW, write() returned false
                              [async error] ERR_STREAM_WRITE_AFTER_END
non-string chunk:             SYNC THROW TypeError ERR_INVALID_ARG_TYPE
JSON.stringify(circular):     SYNC THROW TypeError
```

So the repro steps in #1040 never reach the `catch`, and `handleClose()` already rejects and clears every slot on the socket's `close` event. I took your changeset wording verbatim, since "serializing or writing" is the accurate description and `JSON.stringify` on circular `params` is the realistic source.

That makes this the second wrong claim in the issue after the "hang", and I should have checked the mechanism as carefully as I checked the leak itself. The leak is real, the stated cause was not.

## Suggestions, all three taken

**The dead `?.`.** Captured the socket first. Your reasoning is the same argument as the fix itself: had it been null, the write would be skipped and the slot left pending with no rejection.

**Stub restoration.** Restored in a `finally`, so the test no longer depends on `Writable.end()` bypassing the public `write`.

**The close-path test.** Added: destroy the socket with a request in flight, assert it rejects with `IPC connection closed` and the pending map drains. It passes, which pins the behaviour that made the issue's premise wrong in the first place. Nine tests in the file now.

## Verification

```
$ npx ava source/daemon/ipc.spec.ts
  9 tests passed

# guard removed
✘ [fail]: request rejects and releases the pending slot when write throws
  - 1
  + 0

$ pnpm run test:all
✅ Everything passes!
```

Semgrep still not installed here, so that step is unverified on my side.

## The two adjacent ones

Left both alone as you asked. Happy to file them with what I found if useful:

- no request timeout, which is the actual route to the "hangs" symptom in the issue title
- `connect()` never removing its `once('error', reject)` on success, so the first post-connect socket error is swallowed by a no-op reject and the second has no listener

Say the word, or take them yourself since you found them.

will-lamerton:
Thanks for this PR @rascal-sl - feel free to add yourself as a contributor to our website via a PR which I will approve :)

Issue #1009, opened 2026-08-28
[Feature] Auto-Format on Save Hook


akramcodez:
## Description

Run a configured code formatter automatically after the agent modifies a file.

## Use Case

Agents often generate code with slightly off indentation or formatting that doesn't match the project's style (e.g., Prettier, `gofmt`). Auto-formatting keeps the git diffs clean without the agent having to spend tokens thinking about whitespace.

## Proposed Solution

- Leverage the existing custom-tool primitive for formatters (`source/custom-tools/formatter.tsx`).
- Wire it to run automatically on any file touched by a file modification tool.
- Similar wiring path to the Auto-Lint loop, but failures are non-fatal.

## Alternatives Considered

- Asking the agent to strictly follow style guides (they fail at whitespace often).

## Additional Context

- [x] I have searched existing issues to ensure this is not a duplicate
- [x] This feature aligns with the project's goals (local-first AI assistance)

AniketR10:
@akramcodez would like to work on this please assign

Issue #1004, opened 2026-08-28
[Feature] Audit Log


akramcodez:
## Description

Implement a persistent audit log that records all tool executions and agent actions.

## Use Case

Essential for enterprise users and security compliance. Users need to be able to answer "what did the agent actually do on my machine while I stepped away?".

## Proposed Solution

- Create `source/services/audit-log.ts`.
- Hook into `source/hooks/chat-handler/conversation/conversation-loop.tsx` to log every `tool.execute` event.
- Write to `~/.nanocoder/audit.log` (JSONL format: timestamp, session.id, toolName, argsHash).
- Ensure sensitive data can be redacted.

## Alternatives Considered

- Relying on the session chat history (hard to parse programmatically, UI-focused).

## Additional Context

- [x] I have searched existing issues to ensure this is not a duplicate
- [x] This feature aligns with the project's goals (local-first AI assistance)

addyCooks:
Hey @akramcodez, 
I'd like to pick this one up.. 
I already worked on the lifecycle hooks (#1047), 
so I know the tool execution path pretty well..
could you assign it to me?

**Quick plan:**
1. Add `source/services/audit-log.ts`, which appends one JSONL entry per tool call (`timestamp`, `sessionId`, `toolName`, `argsHash`, `status`) to `~/.nanocoder/audit.log`. It'll be called from the tool execution step in `conversation-loop.tsx`.
2. Arguments are hashed by default, with config-driven redaction (key patterns like `token`/`secret`/`password`) and an opt-in for logging full args. Writes won't block the loop, and AVA specs will cover redaction and file output.

Happy to change the scope if you'd rather go a different way

PR #1047, opened 2026-08-29
feat(hooks): add lifecycle hooks for agent events

### LINKED ISSUE #1032: [Feature] Lifecycle hooks for agent events
## Description

Add a lifecycle hooks system: user-defined shell commands that Nanocoder runs at
defined points in the agent loop before/after a tool runs, on session start/end,
on prompt submit, before compaction.

Hooks are deterministic and free. No model is involved, no tokens are spent, and
they fire every time rather than when the model remembers to.

## Use Case

Nanocoder currently has no way to run user code in response to its own behaviour.

The closest existing mechanism is a skill `subscribe:` block, but it is a different
shape and does not cover this:

| | Skill subscriptions (today) | Hooks (proposed) |
|---|---|---|
| Triggered by | The outside world (file changed, cron fired) | Nanocoder itself |
| What runs | An AI subagent | A shell command |
| Cost | LLM tokens on every fire | Free, instant |
| Deterministic | No | Yes |
| Requires the daemon | Yes, the TUI never starts event sources | No |
| Available moments | 2 (`file.changed`, `schedule.cron`) | Any lifecycle point |
| Can veto an action | No | Yes |

So today you can answer *"the world changed, should an AI look at it?"* but not
*"the agent is about to do X, should I allow it, and what should happen after?"*

Concrete things this unlocks:

- **Auto-format after every edit.** Run `biome check --write` on the file the agent
  just wrote. Today the options are an `AGENTS.md` instruction the model forgets, or
  spending a full LLM turn on it. (This is #1009 one specific instance of the
  general mechanism.)
- **Deterministic policy.** "Never touch `.env`." "Never push to main." "No
  `execute_bash` while the working tree is dirty." `approval-policy.ts` currently
  decides *whether the user is asked*; it cannot run a user-supplied rule and refuse.
- **Automatic context injection.** On session start, run `git log -5` / `docker ps`
  and put the output in front of the model without the user typing `!cmd` each time.
- **Custom notifications.** Post to Slack when a long run completes, alongside the
  built-in desktop notification.
- **Audit logging.** Append every tool call to a file would make #1004 a few lines
  of config rather than a feature.

This is also one of the four weaknesses `docs/battlemap.md` concedes in writing:

> Claude Code's Hooks system (process-level event hooks) also has no direct Nanocoder
> equivalent; Nanocoder's event story runs through Skill `subscribe:` blocks via the
> daemon, not arbitrary user-defined shell hooks on the agent lifecycle.

The other three admitted gaps (community size, surface breadth, distribution polish)
are resourcing problems. This one is a feature someone can build.

Secondary benefit: it lowers the bar for extending Nanocoder from "write a skill in
markdown, or TypeScript" to "add a shell one-liner to a config file" a much larger
group of potential contributors for a 51-contributor project.

## Proposed Solution

Hook definitions in `agents.config.json`, keyed by lifecycle point, optionally
filtered by tool name:

```json
{
  "hooks": {
    "post-tool-use": [
      { "matchTools": ["write_file", "string_replace"], "command": "biome check --write $NANOCODER_FILE" }
    ],
    "pre-tool-use": [
      { "matchTools": ["execute_bash"], "command": ".nanocoder/hooks/guard.sh" }
    ],
    "session-start": [
      { "command": "git log --oneline -5" }
    ]
  }
}
```

Proposed lifecycle points:

| Hook | Fires |
|---|---|
| `session-start` / `session-end` | App init / shutdown |
| `user-prompt-submit` | Before a prompt is sent |
| `pre-tool-use` | Before a tool executes  **can block** |
| `post-tool-use` | After a tool returns |
| `pre-compact` | Before context compaction |

Contract:

- Hook context passed via environment variables (tool name, arguments, changed file
  path, session id, working directory).
- A non-zero exit from `pre-tool-use` denies the tool call, and stdout is returned to
  the model as the denial reason so it can adapt rather than retry blindly.
- stdout from `session-start` / `post-tool-use` is optionally injected as context.
- Every hook has a timeout, so a hanging script cannot wedge the session.
- Hooks are project-local config and therefore a code-execution surface they should
  respect the existing directory-trust boundary, and probably be listed by `/doctor`
  so a user can see what is wired up.

## Alternatives Considered

- **Skill `subscribe:` blocks.** Already exist, but only fire on `file.changed` and
  `schedule.cron`, only run through the daemon, only dispatch to an AI subagent, and
  cannot veto anything. Wrong shape for this problem.
- **Instructions in `AGENTS.md`.** "Always run the formatter after editing." Works
  until the model forgets, which is often, and costs tokens every turn.
- **Custom tools.** The model has to *choose* to call them. Not deterministic, and
  useless for enforcing a rule the model is trying to break.
- **An MCP server.** Far too heavyweight for "run this shell command after an edit",
  and MCP has no hook into Nanocoder's own lifecycle.
- **Adding more one-off features.** #1009 (auto-format) and #1004 (audit log) are both
  individual cases of this mechanism. Building it once is cheaper than building each
  separately, and covers the next N requests too.

## Additional Context

The insertion points are already centralised, which keeps the change small:

- **`pre-tool-use` / `post-tool-use`** → `source/tools/approval-policy.ts`.
  `resolveToolApproval()` is already documented as the single authority every path
  routes through (interactive loop, subagents, plain shell), so there is exactly one
  place to add the gate rather than one per surface.
- **`session-start` / `session-end`** → `source/app/App.tsx` init and the shutdown
  path in `source/utils/shutdown/`.
- **`pre-compact`** → `source/utils/auto-compact.ts`, in `performAutoCompact()`.
- **`user-prompt-submit`** → the input submit handler.

`source/services/bash-executor.ts` already handles spawning with timeout and cwd/env
resolution and can be reused for hook execution.

### Open design question

Should `pre-tool-use` be able to **block**, or only observe?

Blocking is what makes the feature genuinely useful  an observe-only hook cannot
enforce "never touch `.env`". The cost is that a buggy hook can wedge the agent, and
it introduces a second denial path alongside the approval policy.

My suggestion: allow blocking, but keep it strictly inside the existing approval
model a hook returns a decision that feeds `resolveToolApproval()` rather than
short-circuiting it, so there is still one authority and one place to reason about
tool gating. Timeouts and a clear "denied by hook X" message in the transcript would
cover the failure modes.

Happy to adjust the scope if the maintainers would rather start observe-only and add
blocking later.

addyCooks:
## Description

Closes #1032.

Nanocoder had no way to run user code in response to its own behaviour. Skill
`subscribe:` blocks answer "the world changed, should an AI look at it?" — they
fire on file/cron events, dispatch a subagent, cost tokens, need the daemon, and
cannot refuse anything. They are the wrong shape for "the agent is about to do X,
should I allow it, and what happens after?".

Hooks are user-defined shell commands run at fixed points in the agent loop. No
model, no tokens, and they fire every time rather than when the model remembers
to — which is what makes "format after every edit" and "never touch .env"
enforceable instead of merely requested.

Configured under `nanocoder.hooks` in `agents.config.json`, keyed by event,
optionally scoped with `matchTools` and given a `timeout` and a `name`:

```json
{"nanocoder": {"hooks": {
  "post-tool-use": [
    {"matchTools": ["write_file", "string_replace"], "command": "biome check --write \"$NANOCODER_FILE\""}
  ],
  "pre-tool-use": [{"name": "no-env", "command": ".nanocoder/hooks/guard.sh"}],
  "session-start": [{"command": "git log --oneline -5"}]
}}}
```

Six points: `session-start`, `session-end`, `user-prompt-submit`, `pre-tool-use`,
`post-tool-use`, `pre-compact`.

## Contract

- Context arrives as `NANOCODER_*` environment variables (event, cwd, session
  cwd, session id, tool name, tool args, changed file, bash command, tool result,
  prompt, message count). Hook commands are deliberately left out of config-time
  env substitution so `$NANOCODER_FILE` survives to the shell that runs them.
- On `pre-tool-use` and `user-prompt-submit` a non-zero exit denies the action and
  its stdout goes back to the model as the reason, so it can adapt rather than
  retry blindly. **The tool gate runs ahead of the approval decision**, so a
  vetoed tool never renders a confirmation prompt — you see the denial, not a
  diff preview you have to approve first.
- Only a deliberate non-zero exit vetoes. A hook that times out or fails to spawn
  is logged and skipped, so a broken script degrades to "no hook" instead of
  wedging the session. A timeout reaps the hook's whole process tree, not just
  the shell it was spawned under.
- `post-tool-use` stdout is folded into the tool result in a `<hook-output>`
  block — including when the tool failed, so an audit hook sees the errors too.
  `session-start` and `user-prompt-submit` stdout is buffered and prepended to
  the next chat prompt in a `<hook-context>` block, with the transcript still
  showing what the user typed. `/clear` drops undelivered context.
- Hooks run from the project root, so a relative command like
  `.nanocoder/hooks/guard.sh` keeps resolving after the model has `cd`-ed the
  shell elsewhere. `NANOCODER_SESSION_CWD` still reports where the shell is.

## On the open design question (blocking vs. observe-only)

I went with blocking, but alongside `resolveToolApproval()` rather than feeding
into it. Routing a hook decision through the approval policy would be strictly
weaker than the feature needs: yolo mode short-circuits to "no approval" before
any policy runs, so a "never touch .env" rule expressed that way would evaporate
in the mode where you most want it. Keeping them separate leaves two
clearly-scoped authorities — the gate answers "may this run at all?", the
approval policy answers "must a human confirm?" — rather than two overlapping
ones. The denial is attributed by hook name in the transcript.

## Security

Hooks execute a shell command, so `spawn(..., {shell: true})` trips semgrep's
`spawn-shell-true` and `detect-child-process` rules. The suppression is scoped to
those two rule ids and carries the reasoning, because the invariant behind it is
worth protecting explicitly:

> `hook.command` is the only value that reaches the shell. Everything the model
> influences — tool arguments, file paths, bash commands, tool results, prompts —
> is passed through the environment in `buildEnv()` and is never interpolated
> into the command string, so a model-chosen path like `a.ts; rm -rf /` is inert.
> Making that line a template literal would turn a config string into an
> injection sink, so the comment says so.

The residual risk is the hook author's own quoting, since the values they expand
are model-influenced; the docs call that out and the examples use
`"$NANOCODER_FILE"` rather than bare `$NANOCODER_FILE`.

Hooks also inherit Nanocoder's full environment, provider API keys included —
which is what makes `git`, `biome`, and `docker` work, and is the same trust an
`mcpServers` entry in the same file already carries. The docs say so explicitly
rather than leaving it implied.

Beyond that, hooks are project-local shell commands and carry the same
code-execution weight as `mcpServers` in the same file, gated by the same
directory-trust prompt.

## Coverage

The tool hooks are applied on every surface that runs tools: the shared
`processToolUse` path (TUI, `run`, `--plain`, ACP), the streaming bash path the
interactive TUI uses for `execute_bash`, and the subagent executor's own loop.
The session and prompt hooks cover the TUI and `run`/`--plain`; `pre-compact`
covers both automatic compaction and `/compact`.

The gate is applied at more than one layer on purpose — the TUI gates before the
confirmation prompt, and each execution boundary gates again so no surface can
reach a handler ungated. `runPreToolUseGate` keys a `WeakSet` on the tool-call
object, so the hook itself still fires exactly once per call: an audit-log hook
records one line per tool call, not one per layer.

Invalid entries are dropped with a logged error rather than taking the session
down, and `/doctor` lists what a project has wired up.

One deliberate exclusion worth a reviewer's eye: `user-prompt-submit` does not
fire for slash commands or `!` bash passthroughs. Both are local actions that
never reach the model, and prefixing either would break the routing that
recognises it. Neither consumes buffered context — it stays queued for the next
prompt that actually goes to the model. This is documented and tested.

This also closes out one of the four gaps `docs/battlemap.md` concedes in
writing, and subsumes #1009 (auto-format) and #1004 (audit log) as config rather
than as separate features.

## Review follow-ups

All three blocking items and every "worth fixing" item from @will-lamerton's
review are addressed.

**Blocking**

1. **`!bash` passthrough no longer breaks when hook context is pending.** The
   local-action exclusion now covers `message.trim().startsWith('!')` as well as
   `/`, matching `parseInput`'s own trimmed check, and leaves the buffer
   undrained so the context reaches the next real prompt.
2. **The gate is hoisted above `signalToolConfirm`.** It now runs in the
   partition loop in `conversation-loop.tsx`, before the approval decision, and
   likewise above `signalToolApproval` in the subagent executor. A vetoed tool
   never renders a confirmation prompt. The execution-boundary gates stay as
   defence in depth; the once-per-call `WeakSet` keeps the hook itself from
   firing twice.
3. **`post-tool-use` now fires when a tool throws** — in `message-handler.ts`,
   `subagent-executor.ts`, and on the streaming-bash validation-failure path. A
   call refused by `pre-tool-use` is the one exception: it never ran, so there is
   nothing to observe.

**Worth fixing**

- `session-end` now defaults to a **2s** timeout rather than 30s, sized to fit
  inside the shutdown manager's shared 5s budget. Documented, along with the fact
  that raising the hook's own `timeout` alone does nothing.
- A timeout now reaps the **whole process tree**: hooks are spawned `detached` on
  POSIX and signalled by process group; Windows uses `taskkill /T`.
- The full-environment inheritance is now stated in the Security section.
- Hook cwd is `getProjectRoot()`, not `getSafeSessionCwd()`, so a relative
  command survives a model `cd`. `NANOCODER_SESSION_CWD` exposes the shell's
  location for hooks that want it.
- `session-start` is still fire-and-forget so it cannot delay the UI, but the
  first prompt now **awaits the in-flight run** before draining, so a fast typist
  gets the context on prompt one. A `/clear` mid-flight drops that output via a
  generation check rather than grafting it onto the new conversation.

**Nits**

- `HOOK_EVENTS` doc comment now names both vetoing events.
- Hook output is re-truncated after being appended, so a chatty hook can no
  longer push a result past the cap.
- `SESSION_END_HOOK_HANDLER` is a shared exported const, used by both surfaces.
- The project-wins/no-merge resolution is documented, with the warning that a
  project defining any hook disables every global one.
- The veto path echoes the user's prompt before the denial, so the scrollback
  shows a cause.
- `docs/features/hooks.md` moved to `sidebar_order: 15` (no collision).
- Removed a duplicate `register`/`unregister` pair the PR had introduced in
  `shell.spec.ts`.

**Tests** — the wiring sites outside `processToolUse` are now covered, including
the two that hid the blocking bugs. 31 in `lifecycle-hooks.spec.ts` (up from 19),
plus new tests in `useAppHandlers.spec.tsx` (the `!` regression, the
leading-whitespace variant, slash commands, and injection into a real chat
prompt), `shell.spec.ts` (session-start injection, a prompt veto exiting 1 before
any model call, session-end registration), `conversation-loop.spec.ts` (a veto
never reaching the confirmation prompt; a pass still prompting),
`subagent-executor.spec.ts` (veto, veto-before-approval, post-tool-use output,
post-tool-use on throw), and `streaming-bash-tool.spec.tsx` (veto,
`NANOCODER_COMMAND`, post-tool-use folding).

Also merged `upstream/main` and resolved the three conflicts
(`docs/battlemap.md`, `source/config/index.ts`,
`source/subagents/subagent-executor.ts`).

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Documentation update

## Changeset

### Automated Tests

- [x] All existing tests pass

### Manual Testing

- [ ] Tested with Ollama
- [ ] Tested with OpenRouter

## Checklist

- [x] Self-review completed
- [x] Appropriate logging added using structured logging (see CONTRIBUTING.md)


will-lamerton (changes requested):
Really nice piece of work. The design is right, the engine is well factored (one `runLifecycleHooks`, one `appendPostToolUseOutput` fanned into all four execution paths), and the security reasoning around the semgrep suppression is the strongest part of the PR. I ran `lifecycle-hooks.spec.ts` on the branch, 19/19 pass.

Three things need addressing before merge.

**1. `!bash` passthrough breaks whenever hook context is pending.** `useAppHandlers.tsx` prepends `<hook-context>…` to `submittedMessage`, then `handleMessageSubmission` calls `parseInput(message)`, which detects bash via `trimmed.startsWith('!')` (`source/command-parser.ts:6`). With the prefix in front that check fails, so the user's `!ls` is sent to the model as a chat message instead of running locally. The trigger is the headline example: any `session-start` hook that prints (`git log --oneline -5`) buffers context, and the next non-slash input eats it. Slash commands are excluded via `isSlashCommand`, `!` is not. Either extend the exclusion to `message.trim().startsWith('!')` and leave the buffer undrained, or move injection down into `onHandleChatMessage` where the local-action branches are already behind you.

**2. "A denied tool never reaches the confirmation prompt" is not true in the interactive TUI.** In `conversation-loop.tsx` the confirm loop awaits `signalToolConfirm({toolCall})` and only then calls `processToolUse`, where the gate lives. The streaming bash gate is inside `runStreamingBashTool`, which is likewise only reached from `executeApprovedTool`, i.e. post approval. So in normal mode a "never touch `.env`" hook still renders the full confirmation prompt with its diff preview, the user approves, and only then does the veto land. The tool is still blocked so this is not a safety hole, but both the docs and the UX are wrong. Either hoist the gate above `signalToolConfirm`, or correct `docs/features/hooks.md` and the PR body.

**3. `post-tool-use` never fires when a tool throws.** The catch block in `message-handler.ts` returns the error result without going through `appendPostToolUseOutput`, so validation failures and handler exceptions are invisible to `post-tool-use`. Same gap in `subagent-executor.ts`. The docs say "after a tool returns", and an audit-log hook that silently drops every failed call does not really subsume #1004.

Worth fixing or at least documenting:

- `session-end` is bounded by the 5s shutdown budget, not the hook's 30s timeout. `gracefulShutdown` races every handler against `timeoutMs` (default 5000) then calls `process.exit()`, and that budget is shared with the autosave flush. The advertised 30000ms default is unreachable for this event.
- The timeout kills the shell, not its children. With `shell: true`, `proc.kill('SIGTERM')` on a compound command leaves `sh`'s grandchildren running after the agent has moved on. `detached: true` plus `process.kill(-pid)` on POSIX would actually reap the group.
- Hooks inherit the full `process.env`, including provider API keys. Defensible given the `mcpServers` comparison, but the security section documents the injection invariant in detail and says nothing about this. One sentence would close it.
- Hook cwd is `getSafeSessionCwd()`, which the model can move with a bash `cd`. That breaks the relative `.nanocoder/hooks/guard.sh` example in the docs. It fails closed on `pre-tool-use` (your missing-binary test pins exactly this), so it is a self-DoS rather than a bypass, but a `post-tool-use` formatter would just silently stop working. `getProjectRoot()` is the more predictable cwd for a hook defined in project config.
- `session-start` is fire-and-forget, so a slow hook plus a fast typist puts the context on prompt two rather than prompt one.

Nits:

- `HOOK_EVENTS` doc comment in `types/config.ts` says `pre-tool-use` is the only vetoing point, but `VETOING_EVENTS` includes `user-prompt-submit`.
- Hook output is appended after `truncateToolResult`, so final content can exceed the cap by up to 16KB.
- `'lifecycle-hooks:session-end'` is a named const in `useAppInitialization.tsx` but a bare literal in `plain/shell.ts`.
- Hooks resolve project-wins with no merge. Worth a line in the docs, since a global "never touch .env" is silently disabled by any project that defines any hook at all.
- The veto path in `useAppHandlers` never echoes the user's prompt, so scrollback shows a denial with no visible cause.
- `docs/features/hooks.md` has `sidebar_order: 5`, colliding with `commands.md` and `task-management.md`.

On coverage: the 19 tests are well chosen, the timeout, missing-binary and unexpanded-`$VAR` cases each pin a real subtlety. The gap is that every wiring site outside `processToolUse` is untested (the prompt gate and injection, plain shell, the subagent gate, the streaming bash gate), and issues 1 and 2 both live in exactly that space. A test asserting that pending hook context does not swallow a `!` command would have caught the first.

addyCooks:
Thanks @will-lamerton  : ) 

this was a really useful review. All three blocking items are fixed, plus every "worth fixing" point and every nit. Also merged `upstream/main` and resolved the three conflicts.

**1. `!bash` passthrough.** The local-action check is now `isSlashCommand || message.trim().startsWith('!')`, with `.trim()` mirroring `parseInput`'s own check so `  !ls` can't slip through either. Both branches leave the buffer undrained. I went this way rather than moving injection into `onHandleChatMessage` because that's also called by `create-handler.ts` and the processed-prompt path in `app-util.ts:229` gating there would fire `user-prompt-submit` for AI prompts dispatched *by* slash commands like `/create`, reversing the documented contract.

**2. Gate hoisted above the prompt.** It now runs in the tool-partition loop ahead of `resolveToolApproval`, and above `signalToolApproval` in the subagent executor, so a vetoed tool never enters `confirmTools` no prompt, no diff preview. The execution-boundary gates stay as defence in depth, so I added `runPreToolUseGate`, which keys a `WeakSet` on the tool-call object (ids are provider-supplied and not guaranteed unique) to keep the hook itself firing exactly once per call. An audit-log hook records one line per tool call, not one per layer. Docs and PR body corrected.

**3. `post-tool-use` on failure.** Fires from the `catch` in `message-handler.ts` and `subagent-executor.ts`, and the streaming-bash validation path. Only exception, now documented: a call refused by `pre-tool-use` never ran.

**Worth fixing** all five done: `session-end` defaults to 2s inside the shutdown budget; timeouts reap the whole tree (`detached` + group signal on POSIX, `taskkill /T` on Windows); env inheritance documented; cwd is `getProjectRoot()` with `NANOCODER_SESSION_CWD` added; the first prompt awaits an in-flight `session-start`.

**Nits**,  all six done.

**Coverage** every wiring site outside `processToolUse` is now tested: the prompt gate and injection (including the `!` regression that would have caught #1), plain shell, the subagent gate, the streaming bash gate, and veto-before-confirm ordering. `lifecycle-hooks.spec.ts` goes 19 → 31. 232 tests pass across every suite this touches, CI green on all 14 checks.

One thing worth flagging: `shell.spec.ts` had a duplicate `register`/`unregister` pair in `makeFakeShutdownManager` that this PR introduced removed. And on the suppression you singled out, adding the `detached` explanation had pushed the `nosemgrep` comment several lines off the `spawn` call; it's back directly above it, and semgrep reports 0 findings on the file.

Ready for another look.

PR #1056, opened 2026-08-29
fix(tools): keep custom tools registered when MCP disconnects

tisankanj:
## Description

`disconnectMCP()` removed the MCP tools correctly, then threw away everything else:

```ts
this.registry.unregisterMany(mcpToolNames);
await this.mcpClient.disconnect();

// Reset registry to only static tools
this.registry = ToolRegistry.fromToolExports(allToolExports);
this.customTools.clear();
```

`unregisterMany(mcpToolNames)` on the line above already removes exactly the MCP set, so the rebuild was pure redundancy. It is gone. Nothing else changes.

## What the rebuild actually destroyed

My first description named only custom tools and claimed they stayed gone until a restart. @will-lamerton corrected both halves in review, and he is right on each.

**It was not reachable.** `disconnectMCP()` has exactly one production call site: the shutdown handler registered in `initializeMCP` (`tool-manager.ts:98-104`). `reinitializeMCPServers` lives in `useAppInitialization.tsx` and calls `initializeMCP` again without going through it. At shutdown, discarding state is harmless, so no user has hit this. The method is public, so a future caller would.

**It threw away more than I said.** Four things, not one:

| Lost on rebuild | Why it matters |
| --- | --- |
| workspace custom tools | plus the `approval` and `read_only` metadata that plan and headless filtering read |
| skill and bundle tools | `registerSkillTool()` writes into the same registry |
| the `web_search` gate | the constructor unregisters it without a Brave key, but `allToolExports` includes it, so the rebuild resurrected an unusable tool |
| the MCP tools themselves | only if `unregisterMany` were ever dropped, which is what the test now pins |

## Type of Change

- [x] Bug fix

## Tests

The first version of the test could not fail for the right reason, which @will-lamerton caught: the stub reported an empty native registry, so `unregisterMany` ran over an empty array.

It now registers a tool, has the stub report that name, and asserts the MCP tool is gone while the custom tool stays. Two checks that it earns its place:

```text
# delete the unregisterMany call entirely
✘ [fail]: custom tools survive an MCP disconnect
  1 test failed

# restore the original rebuild
✘ [fail]: custom-tools › end-to-end › custom tools survive an MCP disconnect
✘ [fail]: tools › tool-manager › disconnectMCP - keeps skill tools and leaves the web_search gate alone
  2 tests failed
```

The mode filtering is asserted too, since the `approval` and `read_only` metadata lives in the map that was cleared, so `getAvailableToolNames(undefined, 'plan')` and `'headless'` are what actually prove it survived rather than the tool merely being present.

The `web_search` assertion compares against the pre-disconnect state rather than a fixed value, so it holds whether or not a Brave key is configured locally.

## Commands run

```text
$ npx ava source/custom-tools/end-to-end.spec.ts source/tools/tool-manager.spec.ts
  92 tests passed

$ pnpm run test:all
✅ AVA tests passed
✅ Knip check passed
✅ Audit passed
✅ Everything passes!
```

Not run: Semgrep, not installed here. `pnpm run build` is needed before `test:all` on a fresh clone, or the `cli-integration.spec.ts` cases fail on `MODULE_NOT_FOUND`.

## Note

Branches from `main`, independent of #1043 and #1045.

---
<table><tr>
<td></td>
Developed and verified</td>
</tr></table>


will-lamerton (changes requested):
Thanks for this - the fix itself is correct. `getToolEntries()` derives its keys from `getNativeToolsRegistry()`, so `unregisterMany(mcpToolNames)` already removes exactly the MCP set and the rebuild was pure redundancy. I applied the diff locally: the spec file passes 7/7, tool-manager + tool-registry + mcp-client pass 168/168, `tsc --noEmit` clean.

Three things before merge.

**1. The test can't catch the regression it's for.** The stub returns `getNativeToolsRegistry: () => ({})`, so `unregisterMany` gets an empty array and is a no-op. A change that deleted the `unregisterMany` call entirely would still pass this test, leaving MCP tools registered against a dead client. Please register a fake MCP entry, have the stub return that name, and assert the MCP tool is gone *and* the custom tool remains.

**2. The metadata claim isn't asserted.** The PR body and changeset both lean on `approval`/`readOnly` surviving for plan and headless filtering, but the asserts only cover `hasTool` / `isCustomTool` / `getCustomToolNames` / `getAllTools`. A `getAvailableToolNames(undefined, 'plan')` and `'headless'` check after the disconnect is one line each and locks in the actual behaviour.

**3. The changeset needs adjusting in both directions.**

It overstates the symptom: `disconnectMCP()` has exactly one production call site, the shutdown handler registered in `initializeMCP` (`tool-manager.ts:112-118`). `reinitializeMCPServers` doesn't go through it, it just calls `initializeMCP` again. So "custom tools disappeared until Nanocoder was restarted" isn't reachable today - at shutdown, clearing state is harmless. Still worth fixing since the method is public and the daemon is a plausible future caller, but let's describe the defect in the method rather than a user-facing symptom.

It also undersells the blast radius. The rebuild destroyed two more things worth naming:

- Skill-registered tools. `registerSkillTool()` writes into the same registry, so bundle and skill tools vanished on disconnect too.
- The `web_search` gate. The constructor unregisters `web_search` when there's no Brave key, but `allToolExports` includes it, so the rebuild silently resurrected an unusable tool after every disconnect.

Both are cheap to cover with a test each if you're willing.

Unrelated and out of scope, but I noticed it while tracing the call sites: `reinitializeMCPServers` overwrites `this.mcpClient` without disconnecting the old one, which orphans child processes and re-registers the `mcp-client` shutdown handler each time. Worth a separate issue.

rascal-sl:
All three taken. Pushed in d1c24da.

## 1. The test could not fail for the right reason

Correct, and this is the second time in these PRs that I have written a test that passes against the bug. The stub reported an empty registry, so `unregisterMany` ran over an empty array.

It now registers a tool, has the stub report that name, and asserts the MCP tool is gone while the custom tool stays. Two checks that it earns its place:

```
# delete the unregisterMany call entirely
✘ [fail]: custom tools survive an MCP disconnect

# restore the original rebuild
✘ [fail]: custom-tools › end-to-end › custom tools survive an MCP disconnect
✘ [fail]: tools › tool-manager › disconnectMCP - keeps skill tools and leaves the web_search gate alone
```

The first of those is the case you named, and it passed before this change.

## 2. Metadata now asserted

Added `getAvailableToolNames(undefined, 'plan')` and `'headless'` after the disconnect. You are right that this is the assertion that carries the claim: presence in the registry says nothing about the `approval` and `read_only` metadata, and that map is the thing that was cleared. The fixture needed `read_only: true` to be meaningful for plan mode.

## 3. Changeset corrected in both directions

I verified both halves rather than taking them on faith.

**Overstated.** Confirmed: `disconnectMCP()` has one production call site, the shutdown handler at `tool-manager.ts:98-104`. `reinitializeMCPServers` is in `useAppInitialization.tsx` and calls `initializeMCP` again without routing through it. So the "until restart" symptom was not reachable and I have removed it. The changeset now describes the defect in the method and says plainly that nothing user-facing changes today.

**Undersold.** Confirmed both, and both now have a test:

- `registerSkillTool()` writes into the same registry, so skill and bundle tools went with it
- the constructor unregisters `web_search` without a Brave key while `allToolExports` includes it, so the rebuild resurrected an unusable tool

The `web_search` assertion compares against the pre-disconnect state rather than a fixed value, so it holds whether or not a key is configured locally.

The PR description had the same overstatement and has been rewritten too, rather than leaving a claim I now know is wrong sitting at the top of the PR.

## On `reinitializeMCPServers`

I looked while verifying the call sites and agree it is a real problem, and separate from this. Happy to file the issue with what I found if that is useful, or leave it to you since you spotted it.

## Verification

```
$ npx ava source/custom-tools/end-to-end.spec.ts source/tools/tool-manager.spec.ts
  92 tests passed

$ pnpm run test:all
✅ Everything passes!
```

Semgrep is still not installed here, so that step remains unverified on my side.

will-lamerton:
Hey @rascal-sl - all looking good. Please could you resolve the conflict and then good to go :)

rascal-sl:
Conflict resolved in e07fb4d8, `main` merged in. Shows `MERGEABLE` now.

The conflict was additive: we both appended a test to the end of `tool-manager.spec.ts`, and the two blocks ended up sharing one closing brace. Kept both, yours (`plan mode does not mutate the shared tool profile array`) alongside mine.

`tool-manager.ts` auto-merged, so I re-ran the fail-first against the merged code rather than assuming it survived:

```
# original rebuild restored, on merged main
✘ [fail]: tools › tool-manager › disconnectMCP - keeps skill tools and leaves the web_search gate alone
✘ [fail]: custom-tools › end-to-end › custom tools survive an MCP disconnect
  2 tests failed
```

Both still catch it.

```
$ npx ava source/tools/tool-manager.spec.ts source/custom-tools/end-to-end.spec.ts
  94 tests passed

$ pnpm run test:all
✅ Everything passes!
```

The PR diff is still the same four files, so the merge did not drag anything else in.

One note in case it matters for CI: my clone was shallow and `upstream/main` had been force-updated, so git refused the merge as unrelated histories until I unshallowed. Purely local, nothing in the branch.

PR #1061, opened 2026-08-29
feat: add /whoami and /auth commands

### LINKED ISSUE #937: [Feature] /whoami Debug Command
## Description

Add a `/whoami` (or `/auth`) utility command that prints the currently loaded provider, the active API keys (masked for security), and the base URLs.

## Use Case

When API calls fail or users switch between local models (Ollama) and cloud providers (OpenRouter/OpenAI), it can be confusing to remember which configuration is currently active. A quick diagnostic command prevents the need to manually hunt through `.env` or config files.

## Proposed Solution

- Open for suggestions

## Alternatives Considered

- Printing this information on startup. This creates unnecessary noise for daily use.

## Additional Context

- Ensure the API key masking logic strictly hides the middle portion of the key to prevent accidental leaks when users share screenshots.

awhite0030:
## Description

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Changeset

### Automated Tests

### Manual Testing

- [ ] Tested with Ollama
- [ ] Tested with OpenRouter

## Checklist

- [ ] Self-review completed

Root cause: Users have no quick way to check their active provider, model, masked API key, and base URLs.
Fix: Created the `/whoami` command (and `/auth` alias) in `source/commands/whoami.tsx` that extracts this info from `metadata` and `loadProviderConfigs`. Includes a secure API key masking helper with unit tests.
Validation: Ran pnpm run build, pnpm test:format, pnpm test:lint, pnpm test:types, pnpm test:knip, and pnpm test:ava which all pass successfully.
Fixes https://github.com/Nano-Collective/nanocoder/issues/937

Fixes #937

### DIFF
--- a/.changeset/whoami-command.md
+++ b/.changeset/whoami-command.md
@@ -1,0 +1,5 @@
+---
+"@nanocollective/nanocoder": minor
+---
+
+Add `/whoami` and `/auth` diagnostic commands to show active provider, model, masked API key and base URL.
--- a/source/commands/lazy-registry.ts
+++ b/source/commands/lazy-registry.ts
@@ -139,6 +139,16 @@ export const lazyCommands: LazyCommand[] = [
 		load: () => import('@/commands/status').then(m => m.statusCommand),
 	},
 	{
+		name: 'whoami',
+		description: 'Show active provider configuration, API keys, and base URLs',
+		load: () => import('@/commands/whoami').then(m => m.whoamiCommand),
+	},
+	{
+		name: 'auth',
+		description: 'Show active provider configuration, API keys, and base URLs',
+		load: () => import('@/commands/whoami').then(m => m.authCommand),
+	},
+	{
 		name: 'setup-config',
 		description: 'Open a configuration file in your editor',
 		load: () =>
--- a/source/commands/whoami.spec.tsx
+++ b/source/commands/whoami.spec.tsx
@@ -1,0 +1,9 @@
+import test from 'ava';
+import {maskApiKey} from './whoami';
+
+test('maskApiKey hides middle characters', t => {
+	t.is(maskApiKey('sk-1234567890abcdef'), 'sk-1...cdef');
+	t.is(maskApiKey('short'), '********');
+	t.is(maskApiKey(), 'Not set');
+	t.is(maskApiKey('dummy-key'), 'Not set');
+});
--- a/source/commands/whoami.tsx
+++ b/source/commands/whoami.tsx
@@ -1,0 +1,51 @@
+import {Box, Text} from 'ink';
+import {loadProviderConfigs} from '@/client-factory';
+import type {Command} from '@/types/index';
+
+export function maskApiKey(key?: string): string {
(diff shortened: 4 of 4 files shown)


will-lamerton (changes requested):
Thanks for picking this up, the command is a good addition and the masking helper is correct for the cases it covers. A few things to fix before merge.

**Blocking**

1. OAuth providers will report `API Key: Not set` while fully authenticated. `loadProviderConfigs()` substitutes the literal `'dummy-key'` when a provider has no `apiKey` (`source/client-factory.ts:226`), and `github-copilot` / `chatgpt-codex` authenticate through stored device-flow credentials instead (see `validateProviderCredentials`, `source/client-factory.ts:239`). Someone who has run `/copilot-login` will see "Not set" and think they are broken, which is the worst outcome for a command aliased `/auth`. Please branch on `currentProvider.sdkProvider` and use `loadCopilotCredential` / `loadCodexCredential`. Local providers like Ollama would also read better as "Not required (local)" than "Not set".

2. Masking leaks nearly the whole key at lengths 9 to 11: the `<= 8` threshold means a 9 character key renders 8 of its 9 characters. #937 specifically asked for strict middle-hiding so screenshots stay safe. Raise the threshold (12 or so) and consider a fixed-width mask so the output does not disclose key length either.

**Should fix**

3. Hardcoded `borderColor="blue"` / `color="red"` bypass theming. Every other command view pulls `colors.primary` / `colors.error` from `useTheme()`.

4. Please follow the command view convention: export a component (see `source/commands/lsp.tsx`) that uses `useTheme()` and `useTerminalWidth()` inside `TitledBoxWithPreferences` with `title="/whoami"`, and have the handler return `React.createElement(...)` with `generateKey('whoami')`. Right now the result has no React key, so the chat queue falls back to an unstable index key (`source/components/chat-queue.tsx:15`), and without `useTerminalWidth` the box will not match the width of other command output. This also makes the view testable.

5. No render test. The security-critical assertion is missing: nothing proves the raw key never reaches the output. Following `doctor.spec.tsx` with `renderWithTheme` plus `strip-ansi`, assert that a config with `apiKey: 'sk-verysecretvalue1234'` renders without `verysecretvalue`, and cover the unknown-provider branch and the new boundary lengths.

6. `docs/features/commands.md` lists every slash command, `/whoami` and `/auth` should be added there.

**Minor**

7. Provider lookup is exact-match, but `resolveProviderName` (`source/client-factory.ts:81`) resolves case-insensitively, and `currentProvider` can come from a restored session. Lowercasing both sides removes a spurious "Unknown provider" path.

8. The `/auth` entry repeats the `/whoami` description verbatim, so the picker shows two identical rows. Something like "Alias for /whoami" reads better.

Worth a line in the description on how this relates to `/status` (provider, model, theme) and `/doctor` (providers with base URLs), since the genuinely new information here is the masked key. Showing the config file the provider resolved from would also serve the issue's "no hunting through config files" goal nicely.

PR #1062, opened 2026-08-30
fix: correct read_file metadata_only check

### LINKED ISSUE #970: [Bug] isMetadataOnly check for read_file never matches

awhite0030:
## Description

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Changeset

### Automated Tests

### Manual Testing

- [ ] Tested with Ollama
- [ ] Tested with OpenRouter

## Checklist

- [ ] Self-review completed

Root cause: The check to determine if a file read was metadata-only in `source/tools/read-file.tsx` was expecting the result string to start with `"File:"`, but the actual format starts with `"File Information for"`. Furthermore, the original check ignored the `metadata_only` flag entirely and instead incorrectly assumed a large file triggered a metadata-only read.

Fix: I updated the `isMetadataOnly` condition in `source/tools/read-file.tsx` to properly check that `result` starts with `"File Information for"`. I also updated the rest of the condition to include `args.metadata_only` as a valid trigger, while still supporting the auto-fallback behavior for large files. I also added a test and a changeset to confirm correct functionality.

Validation: 
- `pnpm run build`
- `pnpm test:format`
- `pnpm test:lint`
- `pnpm test:types`
- `pnpm test:knip`
- `pnpm test:ava`
All passing.

Fixes https://github.com/Nano-Collective/nanocoder/issues/970

Fixes #970

### DIFF
--- a/.changeset/read-file-metadata-fix.md
+++ b/.changeset/read-file-metadata-fix.md
@@ -1,0 +1,5 @@
+---
+"nanocoder": patch
+---
+
+Fix read_file metadata_only check to correctly parse the 'File Information for' prefix and respect the metadata_only argument.
--- a/source/tools/read-file.spec.tsx
+++ b/source/tools/read-file.spec.tsx
@@ -1120,3 +1120,33 @@ test.serial('read_file metadata_only shows converted binary encoding for DOCX', 
 		rmSync(testDir, {recursive: true, force: true});
 	}
 });
+
+test('ReadFileFormatter correctly identifies metadata_only results', async t => {
+	const testDir = join(process.cwd(), 'test-metadata-formatter-temp');
+	try {
+		mkdirSync(testDir, { recursive: true });
+		const content = Array.from({ length: 100 }, (_, index) => `line-${index + 1}`).join('\n');
+		writeFileSync(join(testDir, 'small.ts'), content);
+
+		const formatter = readFileTool.formatter;
+		if (!formatter) {
+			t.fail('Formatter is not defined');
+			return;
+		}
+
+		const element = await formatter(
+			{ path: join(testDir, 'small.ts'), metadata_only: true },
+			`File Information for "test-metadata-formatter-temp/small.ts"\n==================================================\n\nType: file\nSize: 1000 bytes\nLast Modified: 2023-01-01T00:00:00.000Z\nLines: 100`
+		);
+		const { lastFrame } = render(
+			<TestThemeProvider>{element}</TestThemeProvider>
+		);
+
+		const output = lastFrame();
+		t.truthy(output);
+		t.regex(stripAnsi(output!), /metadata only/);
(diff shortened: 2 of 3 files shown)


will-lamerton (changes requested):
Thanks for tracking this down. The root cause is right and the new test is genuine regression coverage (it fails on `main` as written). I checked the branch out and ran the spec: the new test passes. Two things I'd like changed before merging.

**1. Drop the `totalLines > FILE_READ_PREVIEW_THRESHOLD_LINES` disjunct.**

There is no auto-fallback to metadata for large files. `File Information for` is emitted at `read-file.tsx:48` only, inside `if (args.metadata_only)`. A large file read without the flag returns a 250-line preview plus a `[Truncated at line ...]` marker, never metadata. So that branch is unreachable except by content collision, which I confirmed: a 2001-line file whose first line happens to be `File Information for "something"`, read with no flag and no line range, now renders `(metadata only)` / `Total lines: 2,001` and swallows the truncation notice.

```ts
const isMetadataOnly =
	(args.metadata_only ?? false) &&
	(result?.startsWith('File Information for') ?? false);
```

That also makes the `!args.start_line && !args.end_line` guard unnecessary. The handler checks `metadata_only` first and ignores line ranges, so `{metadata_only: true, start_line: 5}` currently renders `Lines: 5 - 100` / `Tokens: ~107`, where the token count is of the metadata blob rather than the file.

**2. #970 is only half fixed: directory metadata reads still render as content reads.**

`read_file` with `metadata_only: true` on a directory passes the validator and the handler returns valid metadata, but `getCachedFileContent` throws `EISDIR` in the formatter, the `catch` at `read-file.tsx:365` resets `fileInfo` to defaults, and the user sees:

```
Lines: 1 - 0
Tokens: ~0
```

`isMetadataOnly` is still always false there, which is the symptom the issue reports. Computing `isMetadataOnly` from `args` outside the try block that needs file content fixes it, leaving `totalLines`/`tokens` at zero when the content read fails. Same applies to files the handler reports as `Readable: no`.

Minor, non-blocking:

- A test for the directory case and one asserting that a plain large-file read still shows the truncation display would lock both of the above in.
- The commit subject (`Hi, Jules here! ...`) does not follow the repo's conventional commit format. Happy to squash-merge with a rewritten subject if you'd rather not amend.

Issue #1039, opened 2026-08-29
[Bug] paste-utils string replacement only swaps the first occurrence of pasted text


akramcodez:
## Description

When pasting text into the CLI, the paste detector attempts to swap out the raw text for a placeholder so that the terminal isn't overwhelmed. However, because it uses standard `String.prototype.replace()`, it only swaps the very first occurrence. If the same block of text is pasted twice, or exists elsewhere in the input, the second block ships raw to the model.

## Environment

- **OS**: Any
- **Node version**: Node 18+
- **Nanocoder version**: 1.30.0

## Steps to Reproduce

1. Paste a block of code into the CLI input.
2. Paste the exact same block of code again.
3. The first paste gets replaced by a placeholder, the second paste remains raw.

## Expected Behavior

All occurrences of the exact pasted string should be replaced with the placeholder ID.

## Actual Behavior

Only the first match is replaced due to the non-global `.replace()`.

## Additional Context

- [x] I have searched existing issues to ensure this is not a duplicate
- [x] I can reproduce this issue consistently
- [x] This issue occurs with the latest version of nanocoder

## Implementation Notes

Fix in `source/utils/paste-utils.ts` line 64. Change `.replace(pastedText, placeholder)` to `.replaceAll(pastedText, placeholder)`.

Zer0codestuff:
I would like to work on this. I will update the replacement logic, add a regression test covering repeated pasted text, and run the relevant checks before opening a PR.

nightcityblade:
Hi, I'd like to work on this. I'll submit a focused fix with a regression test shortly.

PR #995, opened 2026-08-27
fix: ignore non-positive paste deltas

alexsmolya:
## Summary

- return no added text when paste detection sees a deletion or unchanged input
- preserve paste detection for positive character deltas
- add regression coverage for deletion and non-positive deltas

Fixes #979

## Verification

- `pnpm exec ava source/utils/paste-detection.spec.ts`
- `pnpm test:types`
- `pnpm exec biome check source/utils/paste-detection.ts source/utils/paste-detection.spec.ts`

PR #1063, opened 2026-08-30
Fix paste detection handling for deleted characters

saksham-rathore:
## Description

Fixes an issue in paste detection where deleting characters could result in an incorrect negative character count.

For example, when a user had 11 characters and deleted 5, the character count could incorrectly become `-6` instead of the expected `6`.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Changeset

### Automated Tests

### Manual Testing

- [ ] Tested with Ollama
- [ ] Tested with OpenRouter

## Checklist

- [ ] Self-review completed
- [x] No breaking changes


saksham-rathore:
hey guys, recently I start contributing on open source can you check my PR and tell me is there any kind of mistake and how could I fix that @akramcodez

akramcodez:
@saksham-rathore Thanks for contributing! Could you add a changeset so we can review it and get it ready for merging?

saksham-rathore:
Thanks for your Guidence, Recently I added changeset so now you could merge.

will-lamerton:
Thank you for this, @saksham-rathore, and welcome! You spotted a real bug: the paste detector really was producing a negative `charsAdded` on deletions, and your instinct to short-circuit before running the detection heuristics is exactly the right fix.

The only reason I'm closing this is timing. That same fix landed on `main` earlier today in #995, and a follow-up (`2c278a91`) then reworked the detector to diff the buffer rather than assume appends. So `source/utils/paste-detection.ts` on `main` now has effectively the guard you wrote:

```ts
// Deletions and unchanged input do not contain added text to inspect.
if (charsAdded <= 0) {
  return {isPaste: false, method: 'none', addedText: '', details};
}
```

That's a frustrating way to have your first contribution overtaken, and it's genuinely not a reflection on the work. You independently arrived at the same solution the maintainers did, which is a good sign.

One thing worth knowing for next time, purely as a practical tip: this branch was created from a slightly older `main`, and the diff picked up some unrelated changes along the way (an older copy of the boot summary component reverting the config path work from #855, and the removal of the default branch marker in `source/tools/git/utils.ts`). That kind of thing happens to everyone. Running `git fetch origin && git rebase origin/main` before opening a PR usually keeps the diff to just your own change and makes review much easier. Also worth checking the diff on the "Files changed" tab before submitting, to confirm it only contains what you meant to send.

Two other small habits that will serve you well here:

- One changeset per change. There were two in this PR describing the same fix, which would produce duplicate changelog entries.
- `pnpm run test:format` before pushing. A couple of files lost their trailing newline, and Biome enforces one.

Please do come back. If you'd like something to pick up next, #1039 is right next door to the code you were just in: `paste-utils` only replaces the first occurrence of pasted text when it should replace all of them. You already have the context for it. There are other issues labelled `good first issue` too, and I'm happy to give early feedback on an approach before you invest time in it if you ever want a sanity check.

Thanks again for taking the time to dig into this one.

PR #1064, opened 2026-08-30
fix(skills): reject unsupported skill subscription targets

### LINKED ISSUE #1011: [Bug] TARGET_REGEX misses skill:<name> subscriptions
## Description

The regex used to validate subscription targets in the skill registrar does not match `skill:<name>` targets, even though they are supported and documented.

## Environment

## Steps to Reproduce

1. Create a skill bundle with a subscription target of `skill:some-other-skill`.
2. Load the skill.

## Expected Behavior

The subscription is successfully parsed and registered.

## Actual Behavior

The parsing silently fails or throws an error because the regex rejects the `skill:` prefix.

## Additional Context

## Implementation Notes

In `source/skills/registrar.ts:62` (and `manifest-parser.ts`, `bundle-loader.ts`), `TARGET_REGEX` is defined as `/^(command|agent|tool):([a-z][a-z0-9_-]*)$/`. It needs to be updated to `/^(command|agent|tool|skill):([a-z][a-z0-9_-]*)$/`.

1cbyc:
## Summary

- Parse `skill:<name>` subscription targets at the manifest boundary for forward compatibility.
- Reject every skill target during bundle loading and registration with an explicit unsupported-target error, so no dead subscription can be registered.
- Keep the dispatcher defensive with one reachable skill branch whose message reflects the registration invariant.
- Document the currently supported target kinds and include a patch changeset.

Fixes #1011

## Validation

Rebased onto current `main` (`bf456ad0`).

- Hosted CI: 7,663 tests passed and 11 skipped.
- Hosted line coverage: 92.46%, above the 92.24% `main` baseline.
- TypeScript, formatting, linting, Knip, build, changeset validation, config schema freshness, VS Code packaging, dependency audit, Semgrep, and CodeQL all passed.
- Local focused skill/event suite: 53 tests passed.

Co-authored-by: insisong <emmanuelisaacnsisong@gmail.com>

### DIFF
--- a/.changeset/fix-skill-subscription-targets.md
+++ b/.changeset/fix-skill-subscription-targets.md
@@ -2,4 +2,4 @@
 "@nanocollective/nanocoder": patch
 ---
 
-Allowed `skill:<name>` subscription targets in skill manifests. This fixes manifest parsing, bundle loading, and event-router registration for cross-skill subscriptions while keeping unresolved command, agent, and tool targets rejected. Thanks to @1cbyc. Closes #1011.
+Reject unsupported `skill:<name>` subscription targets with a clear error instead of registering subscriptions that cannot dispatch. Thanks to @1cbyc. Closes #1011.
--- a/.changeset/fix-skill-subscription-targets.md
+++ b/.changeset/fix-skill-subscription-targets.md
@@ -1,0 +1,5 @@
+---
+"@nanocollective/nanocoder": patch
+---
+
+Allowed `skill:<name>` subscription targets in skill manifests. This fixes manifest parsing, bundle loading, and event-router registration for cross-skill subscriptions while keeping unresolved command, agent, and tool targets rejected. Thanks to @1cbyc. Closes #1011.
--- a/docs/features/skills.md
+++ b/docs/features/skills.md
@@ -128,6 +128,11 @@ A multi-verb bundle is natural:
 Subscriptions can declare on member frontmatter (target is implicit
 `self`) or on the bundle manifest (target is explicit `kind:name`):
 
+Accepted manifest target kinds are `command:`, `agent:`, and `tool:`.
+`skill:` is parsed for forward compatibility, but registering it today
+raises a clear "not supported yet" error instead of loading a dead
+subscription.
+
 ```yaml
 # bundle manifest, multiple targets in one place
 subscribe:
--- a/docs/features/skills.md
+++ b/docs/features/skills.md
@@ -128,11 +128,6 @@ A multi-verb bundle is natural:
 Subscriptions can declare on member frontmatter (target is implicit
 `self`) or on the bundle manifest (target is explicit `kind:name`):
 
-Accepted manifest target kinds are `command:`, `agent:`, and `tool:`.
-`skill:` is parsed for forward compatibility, but registering it today
-raises a clear "not supported yet" error instead of loading a dead
-subscription.
-
 ```yaml
 # bundle manifest, multiple targets in one place
 subscribe:
@@ -144,6 +139,11 @@ subscribe:
     target: command:weekly-report
     cron: "0 9 * * MON"
 ```
+
+Accepted manifest target kinds are `command:`, `agent:`, and `tool:`.
+`skill:` is parsed for forward compatibility, but registering it today
+raises a clear "not supported yet" error instead of loading a dead
+subscription.
 
 v1 event kinds: `file.changed` (filter: `paths`, `eventKinds`) and
 `schedule.cron` (filter: `cron`).
--- a/source/skills/bundle-loader.spec.ts
+++ b/source/skills/bundle-loader.spec.ts
@@ -178,6 +178,31 @@ subscribe:
 );
 
 test.serial(
+	'manifest subscription accepts skill targets without local member resolution',
+	async t => {
+		await withTempEnv(async ({projectRoot}) => {
+			const bundleRoot = join(projectRoot, '.nanocoder', 'skills');
+			await makeBundle(bundleRoot, 'cross-skill-sub', {
+				'skill.yaml': `
+name: cross-skill-sub
+description: Cross-skill subscription.
+subscribe:
+  - kind: file.changed
+    target: skill:some-other-skill
+    paths: ["src/**"]
+`,
+			});
+
+			const loader = new BundleLoader(projectRoot);
+			const {skills, errors} = await loader.load();
+			t.deepEqual(errors, []);
+			t.is(skills.length, 1);
+			t.is(skills[0]?.subscribe?.[0]?.target, 'skill:some-other-skill');
+		});
+	},
+);
+
+test.serial(
 	'frontmatter subscription target is resolved to the owning member',
 	async t => {
 		await withTempEnv(async ({projectRoot}) => {
--- a/source/skills/bundle-loader.spec.ts
+++ b/source/skills/bundle-loader.spec.ts
@@ -178,26 +178,27 @@ subscribe:
 );
 
 test.serial(
-	'manifest subscription accepts skill targets without local member resolution',
+	'manifest subscription with skill target is rejected with a clear error',
 	async t => {
 		await withTempEnv(async ({projectRoot}) => {
 			const bundleRoot = join(projectRoot, '.nanocoder', 'skills');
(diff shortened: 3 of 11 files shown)


will-lamerton (changes requested):
Thanks for picking this up. The regex change itself is right and the three regression tests are well placed, but I don't think we should merge it in this shape.

**The main problem: this swaps a loud failure for a silent no-op.**

Issue #1011 says `skill:` targets are "supported and documented" - they aren't. `docs/features/skills.md` and `docs/features/scheduler.md` only ever show `agent:`, `command:`, and `tool:`, and no dispatch path has ever existed for a skill-level target. So this isn't restoring lost behaviour, it's accepting a new target form nothing can act on.

End to end, a `skill:foo` subscription now parses, loads, registers a live `Subscription`, gets matched by the daemon, and then hits the new `dispatcher.ts` branch which logs "deferred" and returns. The user sees a subscription that loads cleanly and never runs. Before this PR they got an immediate error naming the bad target.

Two knock-ons:

- It isn't free. `BackpressureDispatcher` wraps the dispatcher, so a `skill:` subscription with `paths: ["src/**"]` takes a concurrency slot and writes a `Skipped triggered run: ...` line to the daemon log on every debounced save.
- Nothing validates the skill exists. `mergeSubscriptions` short-circuits the `memberRefs` check for `skill:` targets, and the registrar doesn't check either, so `skill:tpyo` is accepted with zero validation while `agent:tpyo` is still correctly rejected.

I'd like one of these instead:

1. Preferred: implement skill-level dispatch. Resolve `skill:foo` to that skill's implicit target using the `resolveImplicitTarget` logic already in `registrar.ts` - it's a small addition given that function exists.
2. If deferring is deliberate: reject `skill:` at registration with a clear "not yet supported" collision rather than registering a dead subscription, and say so in the changeset and docs.

Either way the registrar should validate the target against the loaded skill names - it already has the full `skills` array in `registerSkills`.

**Other things to address:**

- Type modelling. Widening `SkillMemberKind` to include `'skill'` contradicts its own contract ("reference to a single skill member"), and leaks into `SkillCollision.kind` and `resolveImplicitTarget`'s return type, both of which can now express invalid states. Prefer keeping `SkillMemberKind` narrow and adding `export type SkillTargetKind = SkillMemberKind | 'skill'` used only by `SkillMemberRef`.
- Missing dispatcher test. `dispatcher.spec.ts` already covers "command target is reported as unsupported" and "tool target is reported as unsupported"; the new `skill` branch has no equivalent and the test plan doesn't run that spec. It's a copy-paste of the existing pattern and it's the test that documents the deferral decision.
- Docs. `docs/features/skills.md` is where the target grammar is taught. If `skill:` is accepted it belongs there with whatever caveat is true at merge.
- Changeset wording overstates the result. "Allowed `skill:<name>` subscription targets" reads as "this now works". Conventions (attribution, `Closes #`) are otherwise correct.

Minor, not blocking: `TARGET_REGEX` is duplicated in `bundle-loader.ts`, `registrar.ts`, and `manifest-parser.ts`. You had to edit all three in lockstep here and got them all, but the next change may not. Worth hoisting to one shared constant plus a `parseTargetRef` helper, which would also replace the raw `target.startsWith('skill:')` in `bundle-loader.ts`.

No security concerns - the regex stays anchored with the same character class, so no new injection or ReDoS surface.

1cbyc:
Addressed the requested changes in 270c6264:

- `skill:` targets are now rejected during bundle loading and registration with a clear “not supported yet” error, so no dead subscription reaches backpressure or dispatch.
- `SkillMemberKind` remains limited to command/agent/tool; `SkillTargetKind` models the wider parsed target boundary.
- Added registrar, bundle-loader, and dispatcher coverage for the rejection/unsupported paths.
- Documented the current target grammar and corrected the changeset wording so it no longer claims cross-skill dispatch works.

Validation completed locally:

- `pnpm build`
- `pnpm test:all` — 6,879 AVA tests passed, 1 skipped; formatting, TypeScript, lint, Knip, audit, and Semgrep all passed
- GitHub changeset check passed

You can review again.

will-lamerton (changes requested):
Thanks for the quick turnaround on this. Option 2 is the right call and the substance of my last review is addressed. I checked out `270c6264` in a clean worktree, ran `tsc --noEmit`, the seven skills specs, and `biome check` over the touched directories, all clean, 75 tests passing.

**What's addressed:**

- Rejection at registration. `bundle-loader.ts:450` and `registrar.ts:270` both refuse `skill:` with a clear "not supported yet". Nothing reaches the router now, so no backpressure slot and no per-save daemon log line. This was the main thing.
- Type modelling. `SkillMemberKind` stays narrow, `SkillTargetKind` is added, `SkillCollision.kind` is untouched.
- The dispatcher test is there and mirrors the existing command/tool pattern.
- Docs and changeset wording both fixed. The changeset no longer reads as "this now works".
- Validating against loaded skill names is moot under option 2, since `skill:` targets never reach registration. Happy to close that one.

**Three things before I approve:**

**1. Duplicate unreachable branch in `dispatcher.ts`.** `dispatch()` now has two `if (target.kind === 'skill')` blocks, at `source/skills/dispatcher.ts:97` and `:118`. The first returns, so the second is dead. Neither `tsc` nor Biome flags it, which is why it got through.

The surviving message is also inaccurate: `'skill targets are resolved earlier by the registrar'`. The registrar rejects them, it does not resolve them. The spec at `dispatcher.spec.ts:170` only asserts `/skill targets/`, so it passes against either message and doesn't pin the semantics down.

Please collapse to one branch and say what actually happens, something like `'skill targets are rejected at registration and should never reach dispatch'`, and tighten the assertion to match.

While you're in there, worth a thought: under option 2, `Subscription.target` can never carry `kind: 'skill'`. That branch only type-checks because `SkillMemberRef.kind` was widened to `SkillTargetKind`, which flows through to `events/types.ts:68`. I asked for that widening last time, so I'm not going to bounce it, but the tighter shape is to keep `SkillMemberRef` at `SkillMemberKind`, confine `SkillTargetKind` to the three parse sites, and drop the dispatcher branches entirely. Your call, either is fine by me as long as the duplicate goes.

**2. Rebase on main.** The branch is 99 commits behind (merge base `7b46740c`, 30 Aug), and both red checks come from that rather than from your diff:

- Changeset Validation fails with `Cannot find module scripts/validate-changesets.js`. That script landed on main after your merge base, so the check is running against a file the branch has never had.
- Unit Tests & Coverage fails on drift, not on threshold: 91.69% against the 92.17% main baseline, with the 80% threshold passing comfortably. Your added code is covered, and a 0.48% whole-repo drop cannot come from eleven files of this diff.

A merge or rebase from main should clear both. Please confirm coverage actually goes green afterwards rather than assuming it.

**3. Refresh the PR description.** It still describes the pre-fix revision: it reports the coverage failure, says changeset validation failed because the script is absent, and states that loaded-skill-name validation "remains outstanding on the pushed revision". That contradicts your follow-up comment and would leave a misleading merge record.

**Still open from last time, still not blocking:** `TARGET_REGEX` is duplicated in `bundle-loader.ts:63`, `registrar.ts:63`, and `manifest-parser.ts:23`. You edited all three in lockstep again and got them all again, and `bundle-loader.ts:450` still carries the raw `target.startsWith('skill:')`. I'd take the hoist plus a `parseTargetRef` helper if you want to do it here, but I won't hold the PR on it.

**Nit:** the new docs paragraph sits between "...(target is explicit `kind:name`):" and the code block that sentence introduces, so the colon dangles. Moving it below the fence reads better.

1cbyc:
Addressed the requested changes in `b1a0ad96`:

- removed the duplicate unreachable `skill` dispatcher branch;
- made the surviving defensive message match the registration invariant and tightened its test to an exact assertion;
- rebased the branch onto current `main` (`bf456ad0`);
- moved the docs explanation below the YAML example;
- refreshed the PR description and validation results.

All hosted checks are green: 7,663 tests passed (11 skipped), and line coverage is 92.46% versus the 92.24% `main` baseline. The pushed commit is signed and GitHub reports it as Verified.

I kept the suggested regex centralization out of this focused fix because it was marked non-blocking and would broaden the change.

@will-lamerton, this is ready for re-review.

will-lamerton:
Thanks for this PR @1cbyc - feel free to add yourself as a contributor to our website via a PR which I will approve :)

PR #1066, opened 2026-08-30
feat: run auto-compact on --plain, ACP, and subagent loops

### LINKED ISSUE #1048: [Feature] Auto-compact on headless, ACP, and subagent loops
## Description

`autoCompact` is on by default (60% of the model limit, `source/config/index.ts` `loadAutoCompactConfig`). Docs (`docs/features/context-compression.md`) talk like it is a product-wide behaviour.

It only runs in the interactive TUI loop. `performAutoCompact` is called from `processAssistantResponse` in `source/hooks/chat-handler/conversation/conversation-loop.tsx` (around 551), after the assistant message is committed.

These paths never call it:

- `source/plain/conversation.ts` (~168) `client.chat` for `--plain` / `nanocoder run`
- `source/acp/acp-conversation.ts` (~183) ACP / VS Code / Zed
- `source/subagents/subagent-executor.ts` (~436) `client.chat(messages)` with no cap and no compact

Plain and ACP do use `capMessagesForModel` (count, default 1000). Subagents do not even do that. Token budget is ignored everywhere except the TUI.

Local Ollama / 8k models in `nanocoder run`, editor ACP, and delegated subagents will keep stuffing tool results until the provider 400s. The setting in `agents.config.json` does nothing there.

## Use Case

CI / `--plain` job on a small local model. Daemon skill that launches a subagent. VS Code ACP session. Same `autoCompact` config, none of those loops honour it.

## Proposed Solution

Call the existing helper. Do not write a second compressor.

1. After each assistant turn on the three loops, call `performAutoCompact` the same way the TUI does (messages, system message, tokenizer / `getModelContextLimit`, native tool defs).
2. Subagents also pass through `capMessagesForModel` before `client.chat`. Plain and ACP already do.
3. Mechanical strategy if LLM compact is too heavy for a tiny subagent. Reuse `autoCompact.strategy`. No new config key required for v1.

Tests: fixture that is over threshold on the plain loop compact; subagent history longer than `maxMessages` is sliced without orphaning tool rows (`message-capping.ts` already walks back).

## Alternatives Considered

- Leave it as TUI-only. Then the default config is a lie for `--plain`, ACP, and subagents, which is most of the local-first story.
- Fold this into #793. 793 is send-time stubs / a compiler. This is wiring a function that already exists. Smaller and shippable alone.
- Truncate harder per tool (`truncateToolResult`). Already done. Does not replace history compact on long loops.

## Additional Context

- [x] I have searched existing issues to ensure this is not a duplicate (#13 shipped TUI auto-compact; #525 was context-limit detection, closed)

Dhirenderchoudhary:
## Description

`autoCompact` only ran in the TUI. `--plain` / `nanocoder run`, ACP, and subagents ignored the token budget and kept stuffing tool results until the provider 400'd.

Those three loops now call the existing `performAutoCompact` helper after each assistant turn (same gate as the TUI, no chat-queue notify). Subagents also pass history through `capMessagesForModel` and keep the system row out of the compact input so system tokens are not counted twice. No second compressor, no new config key.

Closes #1048

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

## Changeset

### Automated Tests

### Manual Testing

- [ ] Tested with Ollama
- [ ] Tested with OpenRouter

## Checklist

- [x] Self-review completed


will-lamerton (changes requested):
Thanks for this, the core design is right. Insertion points mirror the TUI exactly (after the assistant message, before tool results), so the tool-call pairing invariant is preserved by the same `summariseWithLLM` boundary walk. I ran the branch locally: `tsc --noEmit` is clean and the four touched spec files pass.

Two things I'd like changed before merge, plus a coverage gap.

**1. Compaction is uncancellable in the two loops where cancel matters.**

`--plain` checks `abortSignal.aborted` at `source/plain/conversation.ts:206` and ACP at `source/acp/acp-conversation.ts:156` / 233. Both new `maybeAutoCompact` calls land past those checks. With the default `strategy: 'llm'` that's a full extra model round trip nothing can interrupt, so an ACP `session/cancel` from the editor now stalls for a summariser turn. `summariseWithLLM` already accepts a `signal`; `performAutoCompact` just never passes one. Please thread an optional signal through `maybeAutoCompact` -> `performAutoCompact` -> `summariseWithLLM`, plus an early `if (signal?.aborted) return messages;`.

**2. The subagent gate measures history the model never receives.**

`subagent-executor.ts` caps the payload into `modelMessages` but calls `maybeAutoCompact(messages.slice(1), ...)` on the full uncapped array. Once a run exceeds `sessions.maxMessages`, the gate counts messages already dropped from the request, so compaction fires on a context the provider never sees. Either compact the capped view, or cap after compacting, so both operate on the same set.

**3. Subagent auto-compact has no test.**

The new subagent test covers `capMessagesForModel` only. The headline claim for subagents is untested, and it's the one path that differs structurally from the TUI (system row spliced out and back in). Worth a test.

Smaller points, happy to take these as follow-ups:

- Two divergent system-message guards in the subagent: the capping block handles a non-system `messages[0]`, the compaction block silently no-ops on it. Derive the split once and reuse it.
- `messages = [systemMessage, ...compacted]` rebinds a function parameter, so the caller's `const messages` (line 166) and `initSubagentSession` (441) keep the pre-compaction array. Benign today, easy to trip later.
- The bare `catch { return messages; }` in `maybeAutoCompact` logs nothing. Given #1048 is "compaction wasn't running and nobody noticed", a `logger.debug` in the catch seems worth it, and the structured-logging box is unticked on the checklist.
- No user-visible signal during compaction. `--plain` already has `writeStatus` for empty-turn retries; an LLM-strategy compaction currently reads as a hang.
- `conversation-loop.tsx:595-633` still carries its own inline copy of the gate. This PR creates the shared helper and leaves the TUI as the one caller not using it, which will drift. It needs only an optional `onNotify` to absorb the TUI too.
- The plain call sits inside `if (hasAssistantPayload)` while the TUI runs the gate unconditionally, so a run alternating empty turns and nudges never compacts.

Test hygiene:

- The subagent test writes `getAppConfig().sessions!.maxMessages = 3` in place and restores only when `originalMaxMessages !== undefined`. `reloadAppConfig()` in `finally` is safer.
- Neither new test in `auto-compact.spec.ts` resets its session overrides. They're last in the file so it passes today, but the next test appended inherits `enabled: false` / `strategy: 'mechanical'`. `t.teardown(resetAutoCompactSession)` fixes it.
- `t.true(secondHistory.length < filler.length)` and `t.true(last.length < 8)` pass for reasons other than the one under test. Asserting the compacted shape would fail for the right reason.
- Uncovered branches: the `catch` fallback in `maybeAutoCompact`, and the `toolsDisabled ? undefined : tools` gating.

Nit: new spec imports use `.js` extensions while their neighbours in the same files are extensionless, and they reach into `models-dev-client` directly rather than the `@/models/index` barrel that `auto-compact.ts` uses. Biome excludes `*.spec.ts` so it won't auto-fix.

Changeset and docs are accurate, including the "`/compact` is still TUI-only" caveat.

PR #1068, opened 2026-08-30
chore: label PRs by changed path

### LINKED ISSUE #1051: [Feature] Label PRs by changed path
## Description

Reviewers are sorting a large PR queue by hand. CI already says if tests pass. It does not say what kind of change it is.

There is no `actions/labeler` (or equivalent) under `.github/`. A tools fix, a TUI tweak, and a workflow change all look the same in the list.

This is leftover from closed #155 (PR management). Not merge queue, not extra test jobs, not auto-merge.

Current related automation: `pr-checks.yml`, `changeset-check.yml`, `stale-prs.yml`, `.github/CODEOWNERS`. Same comment/label style as changeset-check. Skip `changeset-release/main`.

## Use Case

Maintainer filters the PR list by `area:tools` / `area:tui` / `area:ci` / `area:docs` instead of opening every title.

## Proposed Solution

1. `.github/labeler.yml`, for example:
   - `source/tools/**`, `source/custom-tools/**` → `area:tools`
   - `source/hooks/**`, `source/components/**`, `source/app/**` → `area:tui`
   - `.github/**` → `area:ci`
   - `docs/**` → `area:docs`
   - `plugins/vscode/**` → `area:vscode`
2. `.github/workflows/pr-labeler.yml` on `pull_request` using `actions/labeler`.
3. Create the `area:*` labels in the repo (or let the action create them).
4. `pnpm changeset --empty` (chore).

## Alternatives Considered

- Redo #155. Too big, already closed.
- Extra CI jobs. The bottleneck is review time, not lint.

## Additional Context

- [x] I have searched existing issues (#155 closed)

Dhirenderchoudhary:
## Description

Adds `actions/labeler` so PRs get `area:*` labels from changed paths (`tools`, `tui`, `ci`, `docs`, `vscode`). Same `pull_request_target` trigger and `changeset-release/main` skip as changeset-check. Labels are created if missing. `sync-labels` is off so existing labels stay.

Closes #1051

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

## Changeset

Empty changeset (chore, no package bump).

### Automated Tests

### Manual Testing

- [ ] Tested with Ollama
- [ ] Tested with OpenRouter

## Checklist

- [x] Self-review completed


will-lamerton (changes requested):
Thanks for picking this up. The core of it is right: the SHA pins all resolve to the versions claimed, the config uses the correct v5+ `changed-files` schema for labeler v7, every globbed path exists, and the `pull_request_target` + `changeset-release/main` skip matches `changeset-check.yml`. Security posture is fine too, no PR-authored code is executed and the config is read from base.

Four things I'd like changed before merge.

**1. Drop the "Ensure area labels exist" step.** `actions/labeler` v7 creates missing labels itself when granted `issues: write`, which this workflow already grants. From the v7 README: "Creating New Labels - Requires: `issues: write` - This is necessary if the action needs to create labels that do not already exist." The manual path is documented as the alternative for people who *don't* want that permission. As written we do both, and the cost is a hardcoded label list duplicating `labeler.yml`. Add `area:mcp` later and forget the script, and it gets a random colour. If you want deterministic colours, keep the step but derive the list from the config rather than restating it.

**2. Drop the checkout step.** The `configuration-path` docs cover this: "If the file doesn't exist at the specified path on the runner, action will read from the source repository via the GitHub API." `github-script` doesn't need a working tree either. Removing it also drops the `pull_request_target` + `actions/checkout` pairing that scanners flag (it's why `changeset-check.yml` carries a `# nosemgrep`), and makes `contents: read` unnecessary. This is coupled to point 1: if you keep the ensure-labels step and have it read the config off disk, checkout stays justified.

**3. `source/vscode/**` gets no label.** We have VS Code code in two places. `plugins/vscode/**` is the extension package, but `source/vscode/**` holds the extension-host integration (`chat-panel-*.spec.ts`, `cli-path-discovery.ts`, `extension-installer.ts`). A PR touching only that matches nothing today. Please add it to `area:vscode`.

Worth flagging the wider gap, though I'm happy to leave it out of scope since the issue only listed five: `source/` has ~40 subdirectories and 5 are mapped, so `ai-sdk-client`, `mcp`, `commands`, `skills`, `daemon` and `session` PRs all land unlabeled. That's a good chunk of our traffic.

**4. Give the labels distinct colours and descriptions.** All five are `0e8a16` with no description. Our core labels (`bug`, `documentation`, `enhancement`) all have descriptions, and identical colours undercut the visual scanning this exists for.

Minor: `sync-labels: false` is already the v7 default, so it's redundant, but it documents intent and the PR body calls it out. Keep it if you like.

Testing note: because of `pull_request_target` this workflow won't run on this PR. The labeler README is explicit that it "needs to be committed blindly". If you want to verify before merge, flip the trigger to `pull_request` in one commit, confirm the labels land, then flip it back.

will-lamerton (changes requested):
Thanks for turning this around quickly. Re-reviewed at `1acbe968`.

Three of the four are done:

- **Point 2, checkout.** Gone, along with the `pull_request_target` + `actions/checkout` pairing. Good.
- **Point 3, `source/vscode/**`.** Added. Verified the path exists (`cli-path-discovery.ts`, `extension-installer.ts`, the `chat-panel-*.spec.ts` files).
- **Point 4, colours and descriptions.** All five now carry both.

Two things left, and the second one is my mistake rather than yours.

### 1. The label list is still restated rather than derived

My ask was to drop the step, or keep it and derive the list from `labeler.yml`. The step was kept and the list is still hardcoded, so it now duplicates *more* of the config than before: name, colour and description instead of just name.

The failure mode I was worried about is unchanged. Add `area:mcp` to `labeler.yml`, forget the script, and labeler v7 creates it with a random colour. Renaming the step to "Color area labels" describes the new behaviour accurately, but it does not close that gap.

Concretely, what would close it: read `.github/labeler.yml` in the `github-script` step via `github.rest.repos.getContent`, take the label names from it, and look each one up in a colour map. A name in the config with no entry in the map should fail the step loudly rather than silently getting a random colour. That keeps your deterministic colours and removes the drift.

### 2. Dropping `contents: read` looks like a regression, and that one is on me

My point 2 said removing the checkout "makes `contents: read` unnecessary". That premise was wrong, and you correctly did what I asked, so this is mine to fix.

With no working tree, labeler v7 reads its config over the API rather than off disk. `src/api/get-label-configs.ts:70-75` falls back to `getContent`, which calls `client.rest.repos.getContent`. The v7 README's "Example Workflow Permissions" section is explicit:

> To ensure the action works correctly, include the following permissions in your workflow file:
> ```yml
> permissions:
>   contents: read
>   pull-requests: write
>   issues: write
> ```

Every workflow example in that README includes `contents: read`. The workflow now grants only `issues: write` and `pull-requests: write`, so `contents` resolves to `none`.

Being straight about the uncertainty: the repo is public, so this may well work in practice. But it is undocumented behaviour to lean on, `pull_request_target` means we cannot test it before merge, and restoring one line costs nothing. It does not bring back the checkout pairing you just removed. Please add `contents: read` back.

### Minor

- The step now calls `updateLabel` unconditionally on every `opened` / `synchronize` / `reopened`. If a maintainer tweaks a label's colour or description in the UI, the next PR event silently reverts it. It is also five API writes per event where the previous version was one paginated read. Create-if-missing would avoid both.
- The bare `catch {}` swallows every failure class, not just the 404 that means "does not exist yet". A 403 falls through to `createLabel` and surfaces as a confusing create error.
- `area:docs` is `0075ca`, byte-identical to our existing `documentation` label, and `area:ci` is `d93f0b`, identical to the existing `extension` label. Worth nudging both, since distinct colours were the point of item 4.

Nothing here is large. `contents: read` is the one that has to happen before merge.

will-lamerton:

PR #1069, opened 2026-08-30
fix: resume queued prompts after slash commands

### LINKED ISSUE #1060: [Bug] After compaction, queued messages do not fire and cannot be edited/deleted
## Description

After usage of the `/compact` command, messages queued during the compaction phase do not auto-start after the compaction is complete, remaining in the queue.

## Environment

- **OS**: Fedora Linux 44 (Workstation Edition)
- **Node version**: v26.7.0
- **Provider**: Ollama 0.33.0
- **Model**: laguna-xs-2.1:q8_0 (not related; happens with other models)

## Steps to Reproduce

1. Run `/compact`
2. While compaction is running, add a new prompt and hit "Enter" so it goes into the queue
3. Wait for compact to finish and observe the queued message being stuck

## Expected Behavior

The first message in the queue would start to run (e.g. a request in plan mode, 'Your next task is...', etc.

## Actual Behavior

The compaction finishes, prompt returns and can be typed in, but queued messages stay in the queue; the method of selecting them for edit/delete also no longer works (down does nothing, up rolls back through old prompts). In order to get them to work, you have to provide a new prompt, cancel it, then select the next message to start it going.

## Logs/Screenshots

Image

Quick "write me a an overview of 6502 emulation" -> compact -> queue "Write me lorem ipsum" -> Prompt doesn't fire; hitting down arrow does nothing, hitting up arrow shows me last message (that I typed to queue).

## Additional Context

- [x] I can reproduce this issue consistently (happens all the time)

mikemikimike:
## Summary

- forward slash-command completion through the app handler lifecycle
- drain queued prompts only after generation, tool execution, cancellation, decision prompts, and modal modes are idle
- make queue draining idempotent across delayed command completion and retried turns
- keep queued-message editing and removal available while idle
- add regression coverage for generation and modal guards, delayed completion, multiple queued prompts, and `/retry`

Closes #1060.

## Validation

- `pnpm exec ava source/app/sections/interactive-app.spec.tsx source/hooks/useAppHandlers.spec.tsx source/app/utils/handlers/retry-handler.spec.ts` — 42 passed
- `pnpm run test:format` — passed
- `pnpm run test:lint` — passed
- `pnpm run test:types` and `pnpm run test:types:vscode` — passed
- `pnpm exec tsc`, `pnpm exec tsc-alias`, and the Windows-equivalent asset copy step — passed
- `pnpm run test:knip` — passed
- `git diff --cached --check` — passed

The full AVA suite was attempted. It reported 79 unrelated Windows/path/network/platform failures and 292 tests pending after timeouts; the changed queue and retry suites passed. The configured `pnpm audit` endpoint is unavailable on the repository's npm mirror, and the package `build` script uses Unix `cp`, so those two checks were documented with their Windows limitations.


will-lamerton (changes requested):
Right diagnosis and the right seam: `/compact` never goes through the chat handler, so the queue drain was unreachable. The problem is that `onCommandComplete` is a shared terminal callback for essentially every slash command, and the drain has no busy/mode guard (`App.tsx:252-279` only checks `client` and `toolManager`).

Blocking issues:

- **Concurrent turns.** Slash commands intentionally bypass the queue while busy (`user-input.tsx:448`) and the input stays enabled during generation (`chat-input.tsx:240`). Queue a prompt mid-turn, type `/status`, and its completion drains the queue into a second `handleChatMessage` on top of the live one. There is no re-entrancy guard, so `setAbortController` (`useChatHandler.tsx:374`) clobbers the in-flight turn's controller and Escape only reaches the newer one.
- **Double drain on `/retry`.** `retry-handler.ts:85-89` awaits `onHandleChatMessage` and then calls `onCommandComplete`. The awaited turn already drains via `onConversationComplete`, so the handler's own call drains a second message while the first is starting. `create-handler.ts` has the same await-then-complete shape on some paths.
- **Drain behind a modal mode.** `/model`, `/explorer`, `/ide`, `/tune` (`app-util.ts:271-276`) and `/settings` (`:309-311`) enter a mode and complete immediately, so a queued prompt starts generating underneath the picker.

Suggested fix: guard the drain on "nothing in flight" (a generating ref plus `activeMode === null`), or drop the callback plumbing and drain from an effect on the busy to idle transition. `cancellable` in `interactive-app.tsx:173-181` already computes that signal, it covers `/compact` naturally, and it cannot fire mid-generation or behind a modal by construction.

Test coverage: `setup()` defaults to `client: null, messages: []`, so `/compact` takes the "No messages to compact" branch (`compact-handler.ts:147-151`) which completes synchronously. The new JSDoc says "including delayed completions", but no `DELAY_COMMAND_COMPLETE_MS` path is exercised, and those are what every real compaction uses (`:226`, `:265`). Please add a delayed-completion case and a negative case asserting no drain while a turn is generating.

Smaller notes:

- `props.onCommandComplete` is missing from `handleMessageSubmit`'s dependency array (`useAppHandlers.tsx:718-751`). It lints clean only because bare `props` is listed; every other prop is enumerated.
- The changeset says "after manual context compaction completes", but the change affects all slash commands. The title is accurate, worth matching them up.
- "Fixes #1060" is only half of the issue. The second symptom, queued messages not being selectable or deletable once idle, is untouched: `user-input.tsx:610`, `:638` and `:669` gate the key handling on `isBusy` while the list plus its "↑/↓ select, Enter edit, Del remove" hint render whenever the queue is non-empty (`:1077`). Any queue that fails to drain is still stranded and still advertises dead controls. Either fix that gating or downgrade to "Refs #1060".

Verified `pnpm exec ava source/hooks/useAppHandlers.spec.tsx` passes on the branch (15 tests). No security or performance concerns, and the style matches the codebase.

mikemikimike:
Thanks for the detailed review. I addressed the requested issues in `5642f223`:

- Queue draining now waits until generation, tool execution, cancellation, abort-controller, confirmation/question, subagent approval, and modal/settings state are idle. The drain is guarded against re-entry and schedules another pass after each dispatched turn settles.
- `/retry` no longer emits a second completion signal after its awaited chat turn, so it cannot start a second queued turn.
- Queued messages can be selected, edited, or removed while idle.
- Added regression coverage for generating and modal states, delayed command completion, multiple queued prompts, and `/retry` completion ownership.

The PR is rebased through a normal merge commit onto the current upstream `main`; no history was force-pushed. Focused validation passes with 42 tests, and format, lint, type checks, the Windows-equivalent build steps, and knip pass. The full AVA and audit limitations are recorded in the PR description.

will-lamerton (commented):
Re-reviewed at `6c5d556d`, diffing the real change set against the merge base (`4143e0ef`) rather than the PR page's 134 files / +10,592, which is merge-commit noise. The actual change is 11 files, +285/-58.

Six of the seven points are addressed. One gap remains, and the branch no longer merges cleanly.

## Addressed

- **Concurrent turns.** The callback drain is gone; `App.tsx` no longer holds `queuedUserSubmitRef` or `drainQueuedUserMessage`. The drain is now an effect (`interactive-app.tsx:199`) gated on `queueDrainBlocked` (`:191`: cancelling, generating, tool-executing, live abortController, tool-confirmation mode, question mode, subagent approval, tool confirmation) plus `isConversationComplete`. Because `handleMessageSubmit` sets `setIsConversationComplete(false)` on entry (`useAppHandlers.tsx:656`), an in-flight slash command blocks the drain too. Re-entrancy is held by `drainInProgressRef`, with a cleanup that only releases the ref when the scheduled timeout has not yet fired. This is the seam I suggested and it holds.
- **Double drain on `/retry`.** `options.onCommandComplete?.()` removed from `retry-handler.ts:89` with an explanatory comment and a new spec asserting zero completion calls.
- **Delayed completion and negative coverage.** `interactive-app.spec.tsx` now has a `DELAY_COMMAND_COMPLETE_MS` harness, a "does not drain while generating" case, a "does not drain while a modal mode is active" case, and a two-item sequential drain case. `app-util.spec.ts` adds a delayed `/status` completion assertion.
- **Dependency array.** Moot under the new design: `onCommandComplete` is an inline lambda over `props.setIsConversationComplete`, which is enumerated in the deps.
- **Changeset wording.** Now reads "after slash commands and manual context compaction complete", which matches the title and the scope of the change.
- **Second half of #1060.** `!isBusy` dropped from `handleQueueNavigation`, `loadSelectedQueuedMessage`, and `removeSelectedQueuedMessage` in `user-input.tsx`, so select / edit / remove work while idle. Test updated to render idle.

## Remaining: the plan review bar is still a modal the drain ignores

`queueDrainBlocked` covers `activeMode` and `isSettingsMode`, but not `appState.planReviewState?.show`. That bar hides the entire `ChatInput` (`interactive-app.tsx:485`), so it is the same class of blocking modal as `/model` or `/settings`.

Reachable path: queue a prompt during a plan-mode turn (`user-input.tsx:473` queues it because `isBusy`), the turn completes uninterrupted, `planTurnCompleted` raises the review bar on the same render that `onConversationComplete` sets `isConversationComplete`, and the drain effect then starts a new turn underneath the plan review prompt.

I confirmed this rather than inferring it. Cloning the existing "does not drain queued prompts while a modal mode is active" test and swapping `activeMode: 'model'` for `planReviewState: {show: true, originalMessage: ''}` fails, and it fails by reaching the dispatch:

```
TypeError: userMessageQueue.drainNextMessage(...).then is not a function
    at Timeout._onTimeout (source/app/sections/interactive-app.tsx:230:6)
```

The `activeMode` variant of that same test passes cleanly, which is the contrast that makes it conclusive. The fix is one clause in `queueDrainBlocked` plus that negative test. Worth also considering `appState.pendingPlanProceed`, since the proceed effect (`:157`) dispatches its own turn a render later.

## Blocking on mechanics

The branch conflicts with `main`. `git merge-tree` reports a single content conflict, in `source/app/sections/interactive-app.spec.tsx`.

## Smaller notes

- `create-handler.ts:416-449` still has the await-then-complete shape I flagged. It is harmless under the new design, since a second signal only re-sets an idempotent boolean and the drain is idle-gated, which means the `/retry` change was defensive rather than strictly required. That is fine, but the asymmetry between the two handlers now reads as arbitrary and deserves a word in the comment.
- Dropping `!isBusy` from `handleQueueNavigation` means Up/Down now select queued messages instead of navigating prompt history whenever the queue is non-empty and the input is empty, including while idle (`user-input.tsx:921`, `:946` sit above `handleHistoryNavigation`). With the drain working, idle-with-a-non-empty-queue should be rare, so this is a note rather than an objection.
- The effect calls `.then()` on `drainNextMessage`'s return unconditionally (`:230`). Production always returns a promise, so this is not a runtime bug, but it does mean the two "does not drain" tests only pass because the guard short-circuits before the mock's non-promise return is touched.
- Two whitespace-only hunks in `user-input.spec.tsx` (blank line near 1108, trailing blank at EOF) are unrelated to the change.

## Verification run locally on the branch

- `pnpm exec ava source/app/sections/interactive-app.spec.tsx source/hooks/useAppHandlers.spec.tsx source/app/utils/handlers/retry-handler.spec.ts source/components/user-input.spec.tsx source/app/utils/app-util.spec.ts` - 160 passed.
- `pnpm run test:lint`, `pnpm run test:format`, `pnpm run test:types` - all clean.

will-lamerton:
Closing in favour of #1210

PR #1070, opened 2026-08-30
fix: standardize tool-approval error kind to single source of truth

### LINKED ISSUE #1014: [Bug] Inconsistent error kind naming (tool-approval-required vs tool-approval)
## Description

There is a typo/inconsistency in how the tool approval error kind is referenced across the codebase, which introduces a silent risk of breakage.

## Environment

## Steps to Reproduce

N/A - Codebase inspection issue.

## Expected Behavior

Error kinds should use consistent string literals across the codebase.

## Actual Behavior

`source/plain/conversation.ts:53` uses `tool-approval-required`, but `app/types.ts:66` uses `tool-approval`. If logic is added that attempts to grep or match on these specific strings, it will fail.

## Additional Context

## Implementation Notes

Standardize the error kind string. Create a single source of truth (e.g., an `enum` or a `const` exported type) and use it in both files.

soumojit-D48:
## Description

Standardize the tool-approval error kind string literal across the codebase by introducing a single source of truth constant `TOOL_APPROVAL_REQUIRED_KIND` in `source/constants.ts` and replacing all bare `'tool-approval-required'` / `'tool-approval'` string literals with it.

Fixes #1014

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Changeset

### Automated Tests

### Manual Testing

- [ ] Tested with Ollama
- [ ] Tested with OpenRouter

## Checklist

- [x] Self-review completed

### DIFF
--- a/source/app.spec.tsx
+++ b/source/app.spec.tsx
@@ -1,4 +1,5 @@
 import test from 'ava';
+import {TOOL_APPROVAL_REQUIRED_KIND} from '@/constants';
 import {isNonInteractiveModeComplete} from './app/helpers';
 import {VALID_MODES} from './app/types';
 
@@ -219,7 +220,7 @@ test('Non-interactive mode: exits with tool-approval reason when tool approval r
 	);
 
 	t.true(shouldExit, 'Should exit when tool approval is required');
-	t.is(reason, 'tool-approval', 'Exit reason should be tool-approval');
+	t.is(reason, TOOL_APPROVAL_REQUIRED_KIND, 'Exit reason should be tool-approval-required');
 });
 
 test('Non-interactive mode: CLI parsing without run command', t => {
--- a/source/app.spec.tsx
+++ b/source/app.spec.tsx
@@ -197,8 +197,8 @@ test('Non-interactive mode: CLI parsing with complex prompt', t => {
 	t.is(prompt, 'create a new file with content');
 });
 
-test('Non-interactive mode: exits with tool-approval reason when tool approval required', t => {
-	// Test that we exit with tool-approval reason when a message indicates tool approval is required
+test('Non-interactive mode: exits with tool-approval-required reason when tool approval required', t => {
+	// Test that we exit with tool-approval-required reason when a message indicates tool approval is required
 	const appStateToolApprovalRequired = {
 		isThinking: false,
 		isToolExecuting: false,
@@ -220,7 +220,11 @@ test('Non-interactive mode: exits with tool-approval reason when tool approval r
 	);
 
 	t.true(shouldExit, 'Should exit when tool approval is required');
-	t.is(reason, TOOL_APPROVAL_REQUIRED_KIND, 'Exit reason should be tool-approval-required');
+	t.is(
+		reason,
+		TOOL_APPROVAL_REQUIRED_KIND,
+		'Exit reason should be tool-approval-required',
+	);
 });
 
 test('Non-interactive mode: CLI parsing without run command', t => {
--- a/source/app/helpers.spec.ts
+++ b/source/app/helpers.spec.ts
@@ -19,7 +19,7 @@ test('isNonInteractiveModeComplete returns timeout when time exceeded', t => {
 	t.is(result.reason, 'timeout');
 });
 
-test('isNonInteractiveModeComplete returns tool-approval when tool approval required', t => {
+test('isNonInteractiveModeComplete returns tool-approval-required when tool approval required', t => {
 	const state: NonInteractiveModeState = {
 		isToolExecuting: false,
 		isToolConfirmationMode: false,
--- a/source/app/helpers.spec.ts
+++ b/source/app/helpers.spec.ts
@@ -1,5 +1,5 @@
 import test from 'ava';
-import {TOOL_APPROVAL_REQUIRED_PREFIX} from '@/constants';
+import {TOOL_APPROVAL_REQUIRED_KIND, TOOL_APPROVAL_REQUIRED_PREFIX} from '@/constants';
 import {isNonInteractiveModeComplete} from './helpers';
 import type {NonInteractiveModeState} from './types';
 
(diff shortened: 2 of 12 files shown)


will-lamerton (changes requested):
Thanks for picking this up! The core idea is right, and I appreciate that you left `'tool-approval'` alone in `settings-selector.tsx` / `settings-tabs.tsx` - that's a settings-panel id in an unrelated domain, and a blind grep-replace would have broken it.

I checked out the branch: `test:types` passes and all 117 tests in the touched specs pass. A few things I'd like changed before merging.

### 1. Drop the `typeof` indirection in the type declarations

In `source/app/types.ts:67` and `source/plain/conversation.ts:76`, `| typeof TOOL_APPROVAL_REQUIRED_KIND` sits next to bare literals (`'complete'`, `'timeout'`, `'error'`, and `'success'` in the plain outcome). That's inconsistent, and it makes a discriminated union harder to read at a glance.

It also isn't solving anything. The bug in #1014 is a *runtime* comparison risk. Type-level drift was never possible - tsc already rejected `helpers.ts` returning a string the union didn't contain. Please revert those two declarations to plain string literals and keep the constant for runtime comparisons and key generation. That keeps the whole fix with none of the legibility cost.

### 2. `cli-harness.spec.ts` still has the bug you're fixing

Lines 278-281 hand-copy production types and logic:

```ts
type ExitReason = 'complete' | 'timeout' | 'error' | typeof TOOL_APPROVAL_REQUIRED_KIND | null;
function getExitCodeForReason(reason: ExitReason): number {
	return reason === 'error' || reason === TOOL_APPROVAL_REQUIRED_KIND ? 1 : 0;
}
```

This tests a function defined in the test file, not `useNonInteractiveMode.ts`, so the union and the mapping can still silently drift from production. Pre-existing, but this PR is the right place for it. Minimum: `import type {NonInteractiveExitReason} from '@/app/types'`. Better: lift the exit-code mapping into `app/helpers.ts` so the test exercises real code.

### 3. Empty `else if`

`source/hooks/useNonInteractiveMode.ts:127-130` is a branch whose entire body is two comments. You're editing that exact line - please delete the branch, the `code` expression below already says it.

### Nits

- `as const` on `constants.ts:117` is redundant; `export const X = 'literal'` already infers the literal type. No other single-value constant in that file uses it.
- The constant sits under the `=== NON-INTERACTIVE NOTICES ===` banner, but `plain/conversation.ts` uses it as an outcome discriminant, which isn't a notice. Own banner or a wider heading.
- `plain/conversation.spec.ts:3` puts `@/constants` before `@/config/index`, breaking the surrounding sort order. Biome excludes spec files so CI won't flag it.
- `app.spec.tsx:223` and `cli-harness.spec.ts:278` now run past the repo's 80-column width. Also biome-excluded, but they read wider than their neighbours.
- Test titles are half-renamed: `cli-harness.spec.ts:295` says "tool-approval-required" now, but `helpers.spec.ts:22` and `app.spec.tsx:199` still say "tool-approval".

### One note on the description

This isn't purely a naming cleanup - it changes the *value* of `NonInteractiveExitReason` from `'tool-approval'` to `'tool-approval-required'`. I confirmed that's safe (only `useNonInteractiveMode.ts` reads it, for exit-code selection, and it's never serialized), but the PR body should say so. Also, the checked "New features include passing tests" / "Tests cover both success and error scenarios" boxes don't describe this change; no new tests are needed for a rename.

No changeset needed for an internal chore - `changeset-check` is non-blocking. `pnpm changeset --empty` will silence the bot if it's bothering you.

soumojit-D48:
@will-lamerton Can you chcek the letest commit then let me know if anything wrong or something ? 
thank u..

will-lamerton (changes requested):
Thanks, this is close. Items 1-3 and all five nits are properly addressed, and lifting `getExitCodeForReason` into `app/helpers.ts` is the better of the two options I offered - the test now exercises real production code. I checked out the branch: `test:types` passes and all 117 tests in the touched specs pass.

Two things before I can merge, both one-liners.

### 1. `source/app/helpers.ts:65` uses a bare literal

```ts
return reason === 'error' || reason === 'tool-approval-required' ? 1 : 0;
```

This is the one production runtime comparison the PR newly authors, and it hand-types the string rather than using the constant. That's exactly the pattern #1014 is about. Please import `TOOL_APPROVAL_REQUIRED_KIND` and use it here.

### 2. The JSDoc above it is inaccurate

It says the function is "Used by both the Ink runtime (`useNonInteractiveMode`) and the headless plain shell (`runPlainShell`) to stay in sync." `runPlainShell` does not call it. `plain/shell.ts:294-300` has its own switch that exits **2** for tool-approval-required and **1** for error, while this function returns **1** for tool-approval-required. Those mappings are deliberately different, so the comment claims a sync that doesn't exist and would send the next person "fixing" one of them. Drop that sentence, or state that the plain shell has its own intentionally distinct mapping.

### Still outstanding: the PR description

From my last pass - please note in the body that this changes the runtime *value* of `NonInteractiveExitReason` from `'tool-approval'` to `'tool-approval-required'` (safe, but it isn't purely naming), and uncheck "New features include passing tests" / "Tests cover both success and error scenarios". No new tests are needed for a rename.

### Nits, optional

- `cli-harness.spec.ts:2-3` puts `@/app/types` before `@/app/helpers`, inverting the sort order. Biome excludes specs so CI won't flag it.
- `app/helpers.spec.ts:2` is a new 87-column import line.

soumojit-D48:
@will-lamerton hi, Check the letest commit, thanku...

will-lamerton:
Thanks for this PR @soumojit-D48 - feel free to add yourself as a contributor to our website via a PR which I will approve :)

soumojit-D48:
@will-lamerton Thanks for merging the PR! I’ll definitely keep contributing to the project and the org.

PR #1071, opened 2026-08-30
fix(acp): claim turns before async prompt setup

### LINKED ISSUE #1038: [Bug] AcpSession.beginTurn() race condition can corrupt session state
## Description

In the ACP session handler, `beginTurn()` creates a new `AbortController` but the `turnActive` boolean flag is set much later in the caller function. If two concurrent prompts arrive in that tiny window, they can both pass the concurrency check and overwrite each other's abort controllers, leading to corrupted UI state.

## Environment

- **OS**: Any
- **Node version**: Node 18+

## Steps to Reproduce

1. Trigger two rapid, concurrent prompts to the ACP interface.
2. The race condition occurs, corrupting the AbortController references.

## Expected Behavior

`turnActive` should be set synchronously inside `beginTurn()` the moment the turn is locked.

## Actual Behavior

`this.turnActive = true` is set in `acp-agent.ts:314` after awaiting other async functions, rather than inside `acp-session.ts:40`.

## Additional Context

## Implementation Notes

Move `this.turnActive = true` directly into `beginTurn()` in `source/acp/acp-session.ts`.

mikemikimike:
## Summary

Fixes #1038.

`AcpAgent.prompt()` now claims a session turn synchronously when `beginTurn()` is called, before content conversion or any other asynchronous setup. This makes the existing overlap check effective for concurrent prompts and prevents the second prompt from replacing the first turn's abort controller.

The prompt cleanup boundary now also covers setup and built-in command paths, so conversion failures and early returns release the active-turn state.

## Tests

- `pnpm exec ava source/acp/acp-session.spec.ts source/acp/acp-agent.spec.ts`
- `pnpm run test:types`
- `pnpm run test:types:vscode`
- `pnpm run test:format`
- `pnpm run test:lint`
- `pnpm run test:knip`
- `git diff --check`

The repository's aggregate `test:all` script cannot run on Windows because it invokes the Bash-only `./scripts/test.sh`; the direct full AVA run also has pre-existing CLI/ACP integration failures and timeout outside this change.

### DIFF
--- a/.changeset/quiet-turns-claim.md
+++ b/.changeset/quiet-turns-claim.md
@@ -2,4 +2,4 @@
 "@nanocollective/nanocoder": patch
 ---
 
-Fixed ACP prompts racing with each other by claiming session turns before asynchronous setup begins, preventing overlapping prompts from corrupting turn state. Closes #1038.
+Fixed ACP prompts racing with each other by claiming session turns before asynchronous setup begins, preventing overlapping prompts from corrupting turn state. Built-in slash-command replies are kept in persisted session history for replay, while command-only sessions are not saved. Closes #1038.
--- a/.changeset/quiet-turns-claim.md
+++ b/.changeset/quiet-turns-claim.md
@@ -1,0 +1,5 @@
+---
+"@nanocollective/nanocoder": patch
+---
+
+Fixed ACP prompts racing with each other by claiming session turns before asynchronous setup begins, preventing overlapping prompts from corrupting turn state. Closes #1038.
--- a/source/acp/acp-agent.spec.ts
+++ b/source/acp/acp-agent.spec.ts
@@ -470,15 +470,12 @@ test('AcpAgent.prompt - rejects an overlapping prompt before the first async bou
 	await t.throwsAsync(
 		agent.prompt({
 			sessionId: session.sessionId,
-		prompt: [{type: 'text', text: 'second'}],
+			prompt: [{type: 'text', text: 'second'}],
 		}),
 		{message: `Prompt already in progress for session: ${session.sessionId}`},
 	);
 
-	t.is(
-		agent['sessions'].get(session.sessionId)!.abortController,
-		controller,
-	);
+	t.is(agent['sessions'].get(session.sessionId)!.abortController, controller);
 	t.is((await first).stopReason, 'end_turn');
 });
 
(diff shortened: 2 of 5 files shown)


will-lamerton (changes requested):
Thanks for this. The core fix is correct and minimal: `git diff -w` reduces the 285-line diff to exactly the two intended hunks, and I confirmed the new test fails against `main`'s implementation, so it is real regression coverage. `tsc`, `biome`, and the 66 ACP specs all pass on the branch.

A few things before merge:

**1. Fix the indentation in the new test.** `source/acp/acp-agent.spec.ts:468` (`prompt: [{type: 'text', text: 'second'}],`) is two tabs short. CI will not catch it because `biome.json` excludes `**/*.spec.ts` from the formatter. There are also two blank lines left before the next test.

**2. Assert the abort controller survives.** The issue is specifically about controller corruption, but the test only asserts the second prompt rejects. Please capture `session.abortController` after the first call and assert identity after the rejection:

```ts
const controller = agent['sessions'].get(session.sessionId)!.abortController;
// ... overlapping prompt rejects ...
t.is(agent['sessions'].get(session.sessionId)!.abortController, controller);
t.is((await first).stopReason, 'end_turn');
```

**3. Built-in slash commands now persist to disk.** Moving `try` above the command interception means `/help`, `/copy`, `/model`, `/provider`, `/settings` etc. now hit `saveAcpSessionToDisk` in the `finally`, where they previously returned before it. `sendBuiltinReply` appends `displayOnly` user and assistant messages, and `saveAcpSessionToDisk` filters on `role` only, so a session whose only interaction is `/help` now gets written to disk titled `/help` and shows up in the resume list. Either filter `displayOnly` out of `saveableMessages` or call this out as intentional in the description.

**4. Optional: cover the conversion-failure path.** A throw from `acpContentToUserMessage` now lands in the `catch` and emits an inline error notice instead of propagating raw. That is an improvement and effectively the second half of the fix, but it is untested. A spec asserting `turnActive === false` plus the error chunk would lock it in.

One non-blocking note: `/clear` still does not persist, because `saveAcpSessionToDisk` early-returns on an empty message list. Pre-existing, but routing `/clear` through the save path now makes it read as fixed when it is not. Worth a follow-up issue rather than something for this PR.

mikemikimike:
verified

will-lamerton (changes requested):
Thanks for the follow-up. I verified `bf96e6fc` in a clean worktree: `tsc --noEmit` is clean and all 181 ACP specs pass. Point 2 is fully met, and the new persistence test is real coverage - I reverted the filter and confirmed `AcpAgent.prompt - built-in replies are not persisted` fails without it. Three things still outstanding.

**1. The indentation from point 1 is still there.** `source/acp/acp-agent.spec.ts:472` is still `prompt: [{type: 'text', text: 'second'}],` at two tabs while its sibling `sessionId:` is at three. The duplicate blank line was fixed, but this half was not, and `biome.json:19` excludes `**/*.spec.ts` from the formatter so CI will not flag it. Small extra while you are in there: the new `t.is(...)` at 477-480 is split across three lines where biome would put it on one (73 columns with the leading tab).

**2. The `displayOnly` filter went a step too far and regresses resumed sessions.** Two filters landed, not one:

- `source/acp/acp-agent.ts:859` filters `displayOnly` out of `saveableMessages`. This is what I asked for and it alone fixes the bug: a `/help`-only session now early-returns and never reaches disk.
- `source/acp/acp-agent.ts:890` additionally filters `displayOnly` out of the persisted `messages` array. This one changes documented behavior. `source/utils/message-visibility.ts:6` states that display-only chrome "renders in chat and persists with session history", and the unreleased `.changeset/fix-display-only-notices.md` ships exactly that guarantee: "they still render in the chat and replay with session history". `replaySessionHistory` (`acp-agent.ts:704`) deliberately does not filter `displayOnly`. With this second filter, a session resumed from disk silently loses its `_Cancelled by user._` notices, its inline `**Error:** ...` banners, and the timeline-revert notice added in `.changeset/fix-display-only-followups.md`. Nothing caught it because no spec covers that round-trip.

Please drop the second filter and keep only the `saveableMessages` one:

```ts
messages: session.messages.map(m => {
	if (m.role === 'user' && typeof m.content === 'string') {
		return {
			...m,
			content: m.content.replace(/^\[Active file: [^\]]+\]\n\n/, ''),
		};
	}
	return m;
}), // We save the raw AcpSession messages with UI prefix stripped
```

**3. The changeset was not updated.** `quiet-turns-claim.md` still describes only the turn-locking fix. My original point 3 asked for either the filter or a note in the description, and the persistence behavior changed either way, so it should be mentioned once you settle on the final shape.

Still open from the first pass, neither blocking: the conversion-failure path (a throw from `acpContentToUserMessage` landing in the `catch`) is untested, and no follow-up issue exists yet for `/clear` not persisting.

Two small changes clear this: fix the tab on `acp-agent.spec.ts:472`, and remove the `.filter(m => !m.displayOnly)` on `acp-agent.ts:890`.

mikemikimike:
Both review rounds have now been addressed, including the requested test formatting and controller-identity assertion, the corrected `displayOnly` persistence behavior, and the changeset update. The merge conflict was resolved by rebasing onto the current `main` while preserving the semantic-memory logic already present there. Tests pass, and all 14 CI checks on `ed96832b` are green.

@will-lamerton, could you please re-review when convenient?

PR #1072, opened 2026-08-30
ci: adopt the shared PR checks workflow; keep the VS Code extension local

will-lamerton:
Seventh and last repo onto the shared `pr-checks` workflow in `Nano-Collective/.github`, after get-md, sentinel, prompt-scrubber, nanoterm, json-up and nanotune.

This closes the loop: nanocoder's 300-line `pr-checks.yml` was the file the shared workflow was derived from. The definition now lives in one place and every repo calls it.

### What stays local

The VS Code extension, as its own job. It is the only genuinely nanocoder-specific work:

- **`test:types:vscode`** — the extension has its own tsconfig, so the root `test:types` does not reach it.
- **`build:vscode`** + verification that `assets/nanocoder-vscode.vsix` was produced, since the release consumes it.

### What is deliberately not carried over

Three things the local workflow did that the shared one does not, each checked rather than assumed:

**`build:credits` before each build.** `source/commands/contributors.json` is committed, and `pnpm build` copies it — verified the build succeeds without regenerating it. `release.yml` still runs `build:credits`, which is where freshness actually matters.

**The Nerd Fonts install.** The `powerline-*` references in `useTitleShape.spec.ts` are shape-name strings in a `TitleShape` union, not font rendering, and Ink tests assert on output strings. Installed fonts cannot change a string comparison. Test verified passing locally.

**`set-safe-directory: false` on the coverage checkout.** Added for `file-snapshot.ts`, which shells out to git. The test it was added for creates its own temp non-git directory and asserts only `t.true(Array.isArray(files))`, with a comment explicitly tolerating either environment. Verified passing locally.

If any of these turn out to be load-bearing, CI will say so on this PR rather than after merge.

### What it gains

Coverage **fail-on-drop**, which the local workflow declared as `FAIL_ON_COVERAGE_DROP: 'true'` but never actually read — the variable appears on line 10 and nowhere else. The baseline is now read from the committed coverage badge on `main`, so coverage cannot regress. Same for `ENABLE_SECURITY_SCAN`, also declared and never read.

It also picks up the semgrep fix from the shared workflow: semgrep runs in the `semgrep/semgrep` container rather than being invoked as `pnpm test:security` on a bare runner, where it is not installed.

### Check names

Blocking: `pr-checks / Linting`, `Type Checks`, `Format Checks`, `Unused Dependencies`, `Unit Tests & Coverage Analysis`, `Verify Build`, plus `VS Code Extension` from the local job.

Advisory: `pr-checks / Package Audit Analysis`, `Semgrep Security Scan`, `CodeQL Security Analysis` — these report but never fail the PR.

### Note on the queue

nanocoder has ~65 open pull requests. They will pick up the new checks when next synchronised. The job names are unchanged from the current workflow for all six shared checks, so nothing that references them by name breaks.


will-lamerton:
Superseded by a clean branch. This PR was accidentally opened from a diverged local `main` — 17 unrelated commits and 393 behind origin/main — which made it CONFLICTING, and that is why only `changeset-check` ran: GitHub cannot compute a merge ref for a conflicting PR, so `pull_request` workflows never fire (`pull_request_target` ones still do, from base). Reopened with the single workflow change on top of current main.

PR #1074, opened 2026-08-30
fix(subagents): preserve object tool output

### LINKED ISSUE #1033: [Bug] Subagents crash when a custom tool returns an object without llmContent
## Description

When a subagent executes a custom tool that returns an object without an `llmContent` field, the subagent executor assigns `undefined` to the content. This literal `undefined` gets sent to the AI SDK, which crashes the current turn.

## Environment

- **OS**: Any
- **Node version**: Node 18+
- **Provider**: Any
- **Model**: Any

## Steps to Reproduce

1. Create a custom tool that returns a JSON object without an `llmContent` property (e.g., `{ someField: "x" }`).
2. Have a subagent execute this custom tool.
3. Observe the subagent execution crash because `content` becomes `undefined`.

## Expected Behavior

If `llmContent` is missing from an object response, the tool executor should stringify the entire object using `JSON.stringify()` as a fallback so the LLM receives the data.

## Actual Behavior

The tool executor blindly checks `typeof result === 'string' ? result : result.llmContent`, passing `undefined` up the stack and crashing the turn.

## Additional Context

## Implementation Notes

Fix in `source/subagents/subagent-executor.ts` lines 625-626:
```typescript
const content = typeof result === 'string' ? result : (result.llmContent ?? JSON.stringify(result));

mikemikimike:
## Summary

Fixes #1033.

When a subagent tool returns a structured object without `llmContent`, the executor now serializes that object as the model-visible tool result instead of passing `undefined` into the next model request. Existing string results and explicit `llmContent` values retain their precedence.

## Tests

- `pnpm exec ava source/subagents/subagent-executor.spec.ts` (31 passed)
- `pnpm run test:types`
- `pnpm run test:format`
- `pnpm run test:lint`
- `git diff --check`

The repository's aggregate `test:all` script is Bash-only and cannot run directly on Windows.

### DIFF
--- a/.changeset/curvy-tools-smile.md
+++ b/.changeset/curvy-tools-smile.md
@@ -1,0 +1,5 @@
+---
+"@nanocollective/nanocoder": patch
+---
+
+Fixed subagent tool results that return structured data without `llmContent` so the complete output is preserved for the model instead of being passed as `undefined`. Closes #1033.
--- a/source/subagents/subagent-executor.spec.ts
+++ b/source/subagents/subagent-executor.spec.ts
@@ -23,7 +23,7 @@ function createMockToolManager(
 			handler: (
 				args: unknown,
 				options?: ToolExecutionContext,
-			) => Promise<string>;
+			) => Promise<unknown>;
 			readOnly: boolean;
 			needsApproval?: boolean;
 		}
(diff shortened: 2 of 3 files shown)


will-lamerton (commented):
Reviewed and verified locally. Reverting only `subagent-executor.ts` to main makes the new test fail with `Error: Cannot read properties of undefined (reading 'length')` instead of `{"someField":"value"}`, so this is a real regression test rather than one written around the new code. Spec passes at 31 tests, types/format/lint clean, CI green.

One request before merge: please add a changeset. `changeset-check` is deliberately non-blocking so it passed, but a subagent bugfix is user-facing and CONTRIBUTING asks for one. `pnpm changeset`, patch bump, something like:

> Fixed subagents dropping tool output when a tool returned a structured object without `llmContent`. Thanks to @mikemikimike. Closes #1033.

Two non-blocking notes, neither of which needs addressing in this PR:

- `result === null` still throws here, since `typeof null === 'object'`. `source/message-handler.ts:79` uses the sturdier `result && typeof result === 'object' && 'llmContent' in result` shape if we ever want the two sites to read alike.
- The same hole exists on the main, non-subagent path: `source/message-handler.ts:90-98` falls through to `result as string` and the object then reaches `truncateToolResult` in `source/ai-sdk-client/converters/message-converter.ts:166`, which throws `content.slice is not a function` outside any tool-level try/catch. Worth a separate issue, probably resolved by a shared `toolResultToText()` helper covering this file, the subagent executor, and `source/commands/update.tsx:103`.

Otherwise good to merge.

mikemikimike:
Thanks for the review. The requested patch changeset is already included in commit `24d0f29e` as `.changeset/curvy-tools-smile.md`, with the requested user-facing bugfix description and `Closes #1033`.

I also re-verified the branch locally:
- `pnpm exec ava source/subagents/subagent-executor.spec.ts` — 31 passed
- `pnpm test:types` — passed
- `pnpm test:types:vscode` — passed
- `pnpm test:format` — passed
- `pnpm test:lint` — passed
- `pnpm run build` — passed
- `git diff --check` — passed

No duplicate changeset was added.

PR #1078, opened 2026-08-30
fix: merge main and tighten memory retrieval

### LINKED ISSUE #619: [Feature]: Continuous Semantic Memory

Dhirenderchoudhary:
## Description

Brings `feat/semantic-memory` up to date with `main` and tightens recall / proposals:

- Rank memories by how much of the query they cover (not by how short the memory is), and require two term hits unless the category matched
- Treat tool-call turns with narration (and short "let me look at the file" turns) as plumbing in reversal detection
- Stop treating camelCase or the bare word `error` as technical evidence; docs match that
- `/memory propose` now requires a category for user lines, drops fragments that do not end in terminal punctuation, and strips leading list markers

Closes  #619

## Type of Change

- [x] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Documentation update

## Changeset

### Automated Tests

Covered ranking (short vs long memory), narrated tool-call reversal, camelCase/`error` evidence, uncategorized user lines, and doubled list bullets. Also ran `tsc --noEmit`. Did not run full `pnpm test:all`.

### Manual Testing

- [ ] Tested with Ollama
- [ ] Tested with OpenRouter

## Checklist

- [x] Self-review completed


will-lamerton (changes requested):
Thanks for this, the core design is solid: atomic writes with `0600`/`0700`, the in-process queue plus cross-process lockfile, fence widening so memory content can't escape the injected block, and `ProposalStore` keeping printed numbering stable are all the right calls. A few things to sort before merge.

First, a scoping note: the diff against `main` is the whole semantic-memory feature (4090 additions, 35 files), not the retrieval tightening the description covers. Worth retitling so reviewers scope correctly.

**1. `/clear` now wipes persisted tasks** (`source/app/utils/app-util.ts:314`)

Only `sharedProposalStore.clear()` belongs to this PR. `clearAllTasks()` and the new "Chat and tasks cleared." message are an unrelated behaviour change: `/clear` previously reset the live task list but left persisted tasks alone, and `/tasks clear` was how you wiped them. That's silent data loss for anyone using `/clear` to reset context mid-task, and it has no changeset, no docs update and no test. Please drop it or land it separately.

**2. Recall misses queries with a single keyword hit** (`findRelevantMemories`)

The new `categoryHit || matchedQueryTerms >= 2` rule is stricter than it reads. Against a memory of "The auth module uses Clerk and avoids middleware in the edge runtime.":

- `refactor the auth middleware` recalls it
- `update clerk auth flow` recalls it
- `fix auth` recalls nothing
- `auth` recalls nothing

Short prompts are exactly where users expect a saved memory to fire. Consider allowing a single hit above a high relevance ratio, or weighting by term rarity rather than a flat count. Either way the docs should say so.

**3. One oversized memory suppresses all recall** (`project-context.ts`, `formatProjectContextWithCount`)

`if (usedTokens + bulletTokens > tokenBudget) break;` discards every shorter memory behind the first one that doesn't fit, and the spec locks that in. `/remember` puts no length cap on content, so a single verbose memory permanently disables recall for any query it ranks first on. `continue` costs nothing and degrades gracefully.

**4. Test-only exports** (`source/memory/project-context.ts`)

`formatProjectContext`, `appendProjectContext` and `appendRelevantProjectContext` have no production consumers, only `project-context.spec.ts`. knip doesn't catch this pattern in this repo. Please delete them and test through `appendRelevantProjectContextWithCount`, or un-export.

Smaller things, non-blocking:

- ACP builds `new SemanticMemoryManager({cwd: session.cwd})` inside `prompt()` and `SubagentExecutor` builds one per `execute()`. Each instance pays a `git config --get remote.origin.url` spawn on first path resolution. `useChatHandler` already memoises this; ACP should cache on the session and the executor on the instance.
- `appendRelevantProjectContextWithCount` ends in a bare `catch {}`, so a corrupt file, a permissions problem and a lock timeout all present identically as "no memories". One `logger.warn` would cover it (the logging checkbox is unchecked in the description).
- `addMemory` calls `listMemories()` inside the lock, which rethrows on malformed JSON, so one corrupt file makes every subsequent `/remember` fail with no repair path. Reads already tolerate junk entries via `isSemanticMemory`.
- In `withExclusiveLock`, the holder's `finally` unlinks the lock path unconditionally, so after a stale reclaim it can delete the new holder's lock. Writing the owner PID in and checking before unlink closes it.
- The settings row is labelled "Semantic Memory" but opens a panel titled "Advanced Settings" implemented as `SettingsAdvancedPanel`. The next advanced setting collides with it.
- `createMemoryCommand` builds a `SemanticMemoryManager` and a `SummarizerService` that constructs its own, doubling the git spawns. Pass the manager in.
- `/memory accept` never passes `sourceSessionId` even though `acceptProposal` takes it, so provenance is lost for accepted proposals.
- ACP has a test that recall doesn't accumulate across turns; the TUI path has no equivalent. It's correct today since `useChatHandler` rebuilds `systemPrompt` each turn, but nothing guards it.

Given the `/clear` change touches shared app state, please run the full suite before the next push rather than just the memory specs.

will-lamerton:
Merging! Thanks for this! One epic feature to finish :)

PR #1079, opened 2026-08-30
fix(events): expand brace patterns in subscription globs

tisankanj:
## Description

A skill subscribing with `*.{ts,tsx}` never fired. `globToRegex` escaped `{`, `}` and `,` as regex literals, so the pattern only matched a path that literally contained the braces.

Braces now expand to an alternation, so `*.{ts,tsx}` matches `app.ts` and `app.tsx` and still refuses `app.js`.

## On the picomatch suggestion

The issue asks for a swap to `picomatch`, and the comment above the matcher says the same:

> Step 9 brings chokidar (and its picomatch dep) into the tree, at which point we can swap this for picomatch and delete the helper.

That premise no longer holds. chokidar 4 dropped glob support along with picomatch, and the installed chokidar declares only one dependency:

```
chokidar 5.0.0
deps: {"readdirp":"^5.0.0"}
```

`require.resolve('picomatch')` fails from the app. So the swap would now mean taking a new direct dependency rather than reusing one already in the tree, which felt like your call rather than mine. I fixed the matcher instead and corrected the comment so it no longer promises a swap that does not follow from chokidar.

Happy to redo this as a picomatch dependency if you would rather have the full glob surface, including negation. That would be a bigger change than this one and would delete the helper.

## Type of Change

- [x] Bug fix

## Behaviour notes

- An unbalanced brace keeps the old literal behaviour instead of building a regex that would throw. Patterns come from user-authored skill frontmatter, so a malformed one should not take the router down. `docs/{a` still matches the literal path `docs/{a`.
- A comma outside braces stays literal, so `a,b.md` is unchanged.
- Negation (`!`) is still unsupported. The issue mentions it in passing, but the title and the reproduction are brace expansion, so I left it alone rather than widen the change.

## Tests

Three cases added to `source/events/event-router.spec.ts`: the alternation itself, unbalanced braces staying literal, and a literal comma outside braces.

The alternation test fails on `main` and passes here:

```text
# against main
✘ [fail]: matchGlob: brace alternation
  1 test failed
```

The other two pass either way by design. They pin behaviour this change must not break, rather than behaviour it adds.

## Commands run

```text
$ npx ava source/events/event-router.spec.ts
  16 tests passed

$ pnpm run test:all
✅ AVA tests passed
✅ Knip check passed
✅ Audit passed
✅ Everything passes!
```

Not run: Semgrep, which is not installed here. Worth noting since `globToRegex` carries a `nosemgrep` line, and I extended that comment to cover the new fixed substrings rather than leaving it describing the old set.

---
<table><tr>
<td></td>
Developed and verified</td>
</tr></table>

### DIFF
--- a/.changeset/glob-brace-alternation.md
+++ b/.changeset/glob-brace-alternation.md
@@ -1,0 +1,5 @@
+---
+"@nanocollective/nanocoder": patch
+---
+
+Fixed skill subscriptions never firing for a brace pattern such as `*.{ts,tsx}`. The event router's glob matcher escaped `{`, `}` and `,` as literals, so the pattern only matched a path that literally contained the braces. Braces now expand to an alternation, and an unbalanced brace stays literal rather than building a regex that would throw. Negation (`!`) is still unsupported. Closes #1012.
--- a/source/events/event-router.spec.ts
+++ b/source/events/event-router.spec.ts
@@ -153,6 +153,34 @@ test('matchesFilter on kind mismatch returns false', t => {
 	t.false(matchesFilter(sub, event));
 });
 
+test('matchGlob: brace alternation', t => {
+	t.true(matchGlob('*.{ts,tsx}', 'app.tsx'));
+	t.true(matchGlob('*.{ts,tsx}', 'app.ts'));
+	t.false(matchGlob('*.{ts,tsx}', 'app.js'));
+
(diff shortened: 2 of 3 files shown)


will-lamerton:
Another great one @rascal-sl :)

PR #1080, opened 2026-08-30
fix(acp): route sub-agent tool approvals to the client

tisankanj:
## Description

A tool call made inside a dispatched sub-agent goes through `signalToolApproval` in `source/utils/tool-approval-queue.ts`. That slot is created with a safe fallback of `() => false`, and the only code that ever installs a handler is `useGlobalHandlerQueues.tsx`, the Ink TUI. Under `--acp` nothing installs one, so every sub-agent tool needing approval was denied without the client seeing a `session/request_permission`.

The visible effect is the one in the issue: a client that gates writes decides every top-level call and nothing a sub-agent does, and delegated work can only write by bypassing approval entirely.

`runAcpConversation` now installs a handler for the turn that forwards these to the same channel top-level calls already use.

## Two details worth calling out

**The call is announced before the request.** `acp-question.ts` documents why, and the top-level path does the same thing with the comment "We reuse this call's id (just announced) so the permission request targets a known tool call". A permission request naming a tool call the client has not seen is rejected as invalid params, and sub-agent calls are never announced otherwise. So the handler emits a `tool_call` update first, then requests permission against that id.

**The title carries the sub-agent name**, as the issue asks, so a client can tell a sub-agent's call apart from a top-level one.

## What this does not do

A denied or cancelled decision marks the announced call `failed`. After an approval it is marked `in_progress` and no terminal status follows, because the sub-agent layer does not report its tool results back to the ACP conversation. That gap predates this change and closing it would mean threading results out of `subagent-executor`, which felt like a separate piece of work rather than something to fold in here. Flagging it because announcing the call is what makes it visible: before this change there was no card at all.

Happy to follow up on that if you want it in the same release.

## Type of Change

- [x] Bug fix

## Tests

Three cases in `source/acp/acp-conversation.spec.ts`, all failing on `main`:

```text
✘ [fail]: runAcpConversation - a sub-agent tool call reaches the client for permission
✘ [fail]: runAcpConversation - a denied sub-agent tool call is reported as failed
✘ [fail]: runAcpConversation - a cancelled sub-agent permission denies the call
  3 tests failed
```

They run a turn, which is what installs the handler, then call `signalToolApproval` exactly as `subagent-executor.ts` does. They assert the request reaches the connection, that its `toolCallId` matches a call announced earlier in the same turn, that the title names the sub-agent, and that deny and cancel both come back false with a `failed` update.

Note on the deny and cancel cases: `t.false(approved)` alone would pass without this change, since the fallback already denies. Each of those tests also asserts the permission request happened and the call was marked failed, which is the part that needs the handler.

## Commands run

```text
$ npx ava source/acp/acp-conversation.spec.ts
  45 tests passed

$ pnpm run test:all
✅ AVA tests passed
✅ Knip check passed
✅ Audit passed
✅ Everything passes!
```

Not run: Semgrep, not installed here. I also have not driven a real ACP client, so this is verified against the connection mock rather than against Zed. The announce-before-request ordering is the part most worth a second opinion.

---
<table><tr>
<td></td>
Developed and verified</td>
</tr></table>


will-lamerton (changes requested):
Thanks for this, the diagnosis is right and the fix is in the right layer. The announce-before-request ordering you asked about is correct: `requestToolPermission` embeds a `ToolCallUpdate` keyed on `toolCall.id`, and the top-level path announces first for the same reason. I applied the branch locally onto current main: spec passes (46), `tsc --noEmit` clean, format clean.

One thing I'd like changed before merge, plus a few follow-ups.

**Blocking: the handler is installed globally and never torn down.**

`setGlobalToolApprovalHandler` is called at the top of `runAcpConversation` with no matching teardown, and the slot is a module singleton with last-writer-wins semantics. `AcpAgent` keeps a `sessions` Map and the overlap guard is `session.turnActive`, which is per session, so two sessions can have turns in flight at once. If session A dispatches a sub-agent and session B starts a prompt before A's sub-agent asks for approval, A's approval runs through B's closure: the `tool_call` is emitted with session B's id and `requestPermission` names a `toolCallId` B's client has never seen. That is exactly the invalid-params rejection your comment warns about, and it also races the wrong `abortController`, so cancelling B cancels A's pending approval. Separately, after the turn returns the closure stays live holding a finished turn's `session`, `conn`, and aborted controller.

Simplest fix: have `GlobalHandlerSlot.set` return a disposer (backwards compatible for the existing TUI caller) and wrap the loop in `try { ... } finally { restore() }`. That confines the closure to the turn; keying the slot by session id, or threading an approval channel through `SubagentExecutor`, would also close the concurrent-session case.

**Worth folding in: a transport failure kills the whole sub-agent run.**

`requestToolPermission` does not catch, and in `subagent-executor.ts:724` the `await signalToolApproval(...)` sits outside the surrounding `try`, so a rejected `requestPermission` (closed connection, or the invalid-params case above) propagates and aborts the sub-agent instead of denying one tool. A `try/catch` around the handler body returning `false` keeps the safe-fallback posture.

**Follow-ups, happy for these to be separate:**

- Sub-agent approvals ignore the ACP session's mode and config `alwaysAllow`. `needsApprovalForTool` passes only `{mode: this.currentMode()}`, and the ACP executor is constructed in `source/plain/initialize.ts:96` with the default `parentMode: 'normal'` and no mode resolver. So a `yolo` or `auto-accept` session still prompts inside a sub-agent while the same tool at top level does not. Pre-existing, but this PR is what makes it visible, so it's worth a line in the changeset.
- On the in_progress gap: note it cuts both ways. Sub-agent tools that don't need approval are never announced at all, so a client gets cards only for the approval-gated subset and each of those spins forever. If threading results out is separate work, consider emitting `completed` right after approval rather than `in_progress`, so nothing is left permanently spinning.
- The tests call `signalToolApproval` after `runAcpConversation` has returned, which only works because of the leak above. They can't catch a regression once teardown is added, and they don't cover the real in-turn case. Driving the signal from inside the turn (e.g. a mock tool handler that calls it) would test the actual ordering and let you assert the session id on the emitted update.
- Minor: sub-agent `toolCallId`s come from the sub-agent's own model and share an id space with top-level calls, so a collision would merge two cards client-side. Prefixing the announced id would rule it out.
- Needs a rebase: the auto-compact work landed in `acp-conversation.ts` and new tests landed at the end of the spec since you branched.

rascal-sl:
All taken, plus the rebase. Pushed in d0249117.

## Blocking: handler scope

You are right, and the concurrent-session case is the worse half. Went with your disposer suggestion.

`GlobalHandlerSlot.set` now returns a function restoring the previous handler, and only if nobody replaced it in the meantime, so an earlier owner's teardown cannot clobber a later one. `runAcpConversation` is now a thin wrapper that installs, runs the turn, and restores in a `finally`. The TUI caller ignores the return value and is untouched.

That also let me split the handler out as `createSubagentApprovalHandler(session, conn)`, so the closure is built per turn from that turn's session rather than captured from the enclosing body.

## Transport failure

Folded in. The handler body is wrapped and returns `false`. As you say, `signalToolApproval` sits outside the executor's own `try` at `subagent-executor.ts:724`, so a rejected `requestPermission` would have taken down the whole run.

## Follow-ups

**`in_progress` gap.** Took your suggestion and emit `completed` after approval rather than `in_progress`, so nothing spins. Your point that it cuts both ways is right and I had not seen it: tools that need no approval are still never announced, so a client gets cards only for the gated subset. I left that alone since it needs the results threaded out, but it is worth an issue.

**Mode and `alwaysAllow`.** Added to the changeset as you asked, spelled out: a `yolo` or `auto-accept` session is still prompted inside a sub-agent for a tool it would not be prompted for at top level.

**The tests.** This was the sharpest catch. They passed only because of the leak, and they would have gone green against a broken teardown. They now signal from inside the turn, through the awaited `chat()` call, which is a point where the handler is installed and the turn has not returned. Two checks that they bite:

```
# teardown removed
✘ [fail]: the approval handler does not outlive the turn
  1 test failed

# handler install removed entirely
✘ [fail]: a sub-agent tool call reaches the client for permission
✘ [fail]: a denied sub-agent tool call is reported as failed
✘ [fail]: a transport failure denies the tool rather than aborting the sub-agent
✘ [fail]: the approval handler does not outlive the turn
  4 tests failed
```

**Id collision.** Took it. Announced ids are prefixed `subagent:`, and the test asserts `subagent:call-1` rather than the raw id.

## Rebase

Merged current `main`. One conflict in the spec, both sides having appended at the end. I dropped my old block rather than resolving it, since those tests were being rewritten anyway, and kept yours. `acp-conversation.ts` auto-merged around the auto-compact work; the diff against `main` is the four files this PR owns.

```
$ npx ava source/acp/acp-conversation.spec.ts
  47 tests passed

$ pnpm run test:all
✅ Everything passes!
```

Semgrep still not installed here.

PR #1083, opened 2026-08-30
Use minimal tool profile for unknown model ids

### LINKED ISSUE #1013: [Bug] inferToolProfile defaults to 'full', breaking small cloud models
## Description

The `inferToolProfile` heuristic defaults to the `full` tool profile for any unrecognized or cloud model. This is detrimental for small cloud models (like Haiku or 8B variants) which have small context windows.

## Environment

## Steps to Reproduce

1. Connect to an unknown or small cloud API model (e.g., a generic 8B model hosted on OpenRouter that doesn't match the hardcoded strings).

## Expected Behavior

The system should ideally default to a safer, smaller profile like `minimal` or `auto`, or the user should be warned.

## Actual Behavior

`inferToolProfile` returns `full`. The model is overwhelmed with the massive system prompt containing all tools, leading to context blow-up and poor performance.

## Additional Context

## Implementation Notes

In `source/tools/tool-profiles.ts:80`, adjust the fallback logic. If a model is unknown, perhaps default to `minimal` instead of `full`, or add a config setting to let the user define the default fallback profile.

mrprohack:
## Description

Change automatic tool-profile inference so model ids with no recognizable size hint use the safer `minimal` profile instead of `full`. A completely missing model still defaults to `full`, preserving startup behavior before a model is known.

Fixes #1013.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Changeset

- [x] Added a patch changeset

## Testing

- [x] Updated regression coverage for unknown model ids and missing-model behavior
- [x] `pnpm exec ava source/tools/tool-profiles.spec.ts` — 24 passed
- [x] `pnpm test:types`
- [x] `pnpm test:format`
- [x] `pnpm test:lint`
- [x] `pnpm test:knip`
- [x] `pnpm test:audit`

## Checklist

- [x] Self-review completed
- [x] No breaking changes

### DIFF
--- a/.changeset/fix-unknown-model-tool-profile.md
+++ b/.changeset/fix-unknown-model-tool-profile.md
@@ -1,0 +1,5 @@
+---
+"@nanocollective/nanocoder": patch
+---
+
+Changed automatic tool-profile fallback for unknown model ids from `full` to the safer `minimal` profile.
--- a/source/tools/tool-profiles.spec.ts
+++ b/source/tools/tool-profiles.spec.ts
@@ -141,9 +141,12 @@ test('inferToolProfile - MoE ids size by total params, not active count', t => {
 	t.is(inferToolProfile('Qwen3-235B-A22B'), 'full');
 });
 
-test('inferToolProfile - cloud/unknown models default to full', t => {
-	t.is(inferToolProfile('claude-opus-4-8'), 'full');
-	t.is(inferToolProfile('gpt-4o'), 'full');
+test('inferToolProfile - unknown model ids default to minimal', t => {
+	t.is(inferToolProfile('claude-opus-4-8'), 'minimal');
+	t.is(inferToolProfile('gpt-4o'), 'minimal');
+});
+
+test('inferToolProfile - missing model defaults to full', t => {
 	t.is(inferToolProfile(undefined), 'full');
 	t.is(inferToolProfile(''), 'full');
 });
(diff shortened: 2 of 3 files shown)


will-lamerton (changes requested):
Thanks for picking this up. Two blockers before this can land.

**1. CI is red - this PR breaks a test it didn't update.**

`source/components/development-mode-indicator.spec.tsx:68` ("tune label resolves auto to full for cloud/unknown models") renders with `currentModel="claude-opus-4-8"` and asserts `/tune: full \(auto\)/`. The component calls `resolveToolProfile` directly (`development-mode-indicator.tsx:74`), so it now renders `tune: minimal (auto)`. It's the only failing test in the job. The checklist ran `tool-profiles.spec.ts` in isolation, which is why it slipped - please run the full `pnpm run test:ava` next time.

**2. The blast radius is much wider than #1013 asks for.**

`TUNE_DEFAULTS` is `{enabled: true, toolProfile: 'auto'}` (`source/types/config.ts:429`), so this is the default path for everyone. No frontier cloud model id carries a size token (`claude-opus-5`, `gpt-4o`, `gemini-2.5-pro`, all Copilot/Codex ids), so they all now resolve to `minimal`. By default that means:

- MCP tools disappear entirely (excluded from every profile except `full`)
- 8 tools only, so `web_fetch`, todo, LSP, custom tools and skill-registered tools all vanish
- single-tool-per-response enforcement turns on (`conversation-loop.tsx:491`) for models that parallel-call well
- the slim prompt path activates (`prompt-builder.ts:211`)

#1013 is about *small* cloud models (Haiku, generic 8B on OpenRouter) getting the full surface. Flipping the fallback wholesale fixes that by degrading every capable cloud model, which is a worse regression than the bug. The issue's own implementation note offers the better route: a config setting for the auto fallback profile (e.g. `tune.autoFallbackProfile`, defaulting to `full`). That fixes #1013 for the people hitting it without silently stripping MCP and parallel tool calls from every Claude/GPT user on stock config. A provider-aware heuristic would also work; a blanket flip doesn't.

**3. Stale docs and comments, all now factually wrong:**

- `source/types/config.ts:430` - the `TUNE_DEFAULTS` comment still says large/cloud models resolve to `full` (no change)
- `docs/configuration/mcp-configuration.md:80` - tells users to switch to a cloud model so `auto` resolves to `full` and MCP tools come back. Under this PR that advice does the opposite of what it says.
- `docs/features/tune.md:38` - same wrong instruction
- `docs/features/tune.md:109` - documents defaults as `enabled: false, toolProfile: 'full'`

For what it's worth the rest is clean: the changeset is scoped correctly, the `!model` early return is preserved and separately tested, and the tooltip was updated in lockstep with the behavior.

Happy to review again once the direction on point 2 is settled.

PR #1086, opened 2026-08-30
fix(file-ops): keep replacement text literal in string_replace and diff_edit

### LINKED ISSUE #1057: [Bug] Replacement strings containing $ corrupt every string_replace and diff_edit
## Description

`string_replace` and `diff_edit` pass the model's replacement text as the second
argument to `String.prototype.replace`. That argument is not a literal the
engine runs `GetSubstitution` over it, so four token sequences are rewritten
before the bytes reach disk:

```
$$   ->  collapses to a single "$"
$&   ->  injects the matched text
$`   ->  injects everything BEFORE the match
$'   ->  injects everything AFTER the match
```

These are ordinary characters in shell scripts, Makefiles, docker-compose files,
CI YAML, and any code that builds regexes. The file silently ends up containing
something the model never asked for and the user never approved.

The `` $` `` and `$'` forms are the damaging ones: each duplicates a whole half of the file
into the middle of the edit, so corruption scales with file size.

## Environment

- **OS**: Windows 11 (language-level; reproduces on every platform)
- **Node version**: v22.22.3
- **Provider**: any
- **Model**: any

## Steps to Reproduce

1. `read_file` a shell script so the read-tracker marks it seen.
2. Call `string_replace` on it with a `new_str` containing any of the four tokens
   (the reduced case below uses all four).
3. Compare the approved preview with the bytes on disk.

Reduced to the one line that matters:

```js
const file    = '#!/bin/sh\necho "old"\nexit 0\n';
const old_str = 'echo "old"';
const new_str = 'echo "pid=$$ match=$& pre=$` post=$\'"';

file.replace(old_str, new_str);        // what the tool does
file.replace(old_str, () => new_str);  // what it should do
```

## Expected Behavior

The file contains exactly the `new_str` the model supplied, byte for byte.

## Actual Behavior

```
requested new_str : "echo \"pid=$$ match=$& pre=$` post=$'\""
current  behaviour: "#!/bin/sh\necho \"pid=$ match=echo \"old\" pre=#!/bin/sh\n post=\nexit 0\n\"\nexit 0\n"
with fn replacer  : "#!/bin/sh\necho \"pid=$$ match=$& pre=$` post=$'\"\nexit 0\n"
```

The tool reports success. Nothing warns the model or the user.

## Logs/Screenshots

Affected call sites:

- `source/tools/file-ops/string-replace.tsx:91` `fileContent.replace(old_str, new_str)`, the write path
- `source/tools/file-ops/diff-edit.tsx:165` `newContent.replace(block.search, block.replace)`, once per block
- `source/tools/file-ops/string-replace.tsx:165` the same call in the VS Code diff formatter

## Additional Context

The approval gate is bypassed by construction rather than by attack: the terminal
confirmation renders `old_str` / `new_str` directly, so the diff the user approves
is not the diff that lands.

Suggested fix — use a function replacer at both write sites, so no substitution
parsing happens at all:

```ts
const newContent = fileContent.replace(old_str, () => new_str);
```

Splicing by index works too and avoids re-scanning the string. Line 165 of
`string-replace.tsx` should get the same treatment so the VS Code preview keeps
matching what is written. Worth a regression test whose `new_str` carries all
four tokens.

addyCooks:
## Description

Closes #1057.

`string_replace` and `diff_edit` passed the model's replacement straight to
`String.prototype.replace`. That second argument is a substitution template, not a
literal, so the engine ran `GetSubstitution` over it and rewrote four token sequences
before the bytes reached disk:

| token | became |
|---|---|
| `$$` | a single `$` |
| `$&` | the matched text |
| ``$` `` | everything **before** the match |
| `$'` | everything **after** the match |

The last two duplicate a whole half of the file into the middle of the edit, so the
damage scaled with file size.

These are ordinary characters in shell scripts, Makefiles, docker-compose files, CI
YAML, and anything that builds a regex. The tool reported success and nothing warned
the model or the user, so the approval gate was bypassed by construction: the
confirmation renders `old_str` / `new_str` directly, so the diff the user approved was
not the diff that landed.

### Fix

`replaceFirstLiteral` splices by index, which sidesteps substitution parsing entirely
and avoids re-scanning the string for a second pass. Four call sites:

- `source/tools/file-ops/string-replace.tsx` the write path
- `source/tools/file-ops/string-replace.tsx` the VS Code diff in the formatter
- `source/tools/file-ops/diff-edit.tsx` once per block in `applyBlocks`
- `source/acp/acp-tool-call.ts` the ACP whole-file diff preview

The ACP site is not named in the issue, but it synthesizes the same post-edit content
for the diff shown over ACP and would have desynced from disk the same way. Both
previews now match what is written.

`string-replace-preview.tsx` renders `new_str` directly and needed no change.

Every call site is already guarded by an existing uniqueness check, so the helper's
not-found branch is a defensive no-op rather than a silently skipped edit.

Three production lines changed, plus one 28-line helper.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Changeset

### Automated Tests

### Manual Testing

- [ ] Tested with Ollama
- [ ] Tested with OpenRouter

## Checklist

- [x] Self-review completed

### DIFF
--- a/.changeset/fix-literal-replacement-dollar-tokens.md
+++ b/.changeset/fix-literal-replacement-dollar-tokens.md
@@ -1,0 +1,5 @@
+---
+"@nanocollective/nanocoder": patch
+---
+
+Fixed `string_replace` and `diff_edit` corrupting edits whose replacement text contains `$`. Both tools passed the model's replacement straight to `String.prototype.replace`, which treats that argument as a substitution template rather than a literal: `$$` collapsed to a single `$`, `$&` expanded to the matched text, and ``$` ``/`$'` spliced a whole half of the file into the middle of the edit. Those are ordinary characters in shell scripts, Makefiles, CI YAML and anything that builds a regex, so the bytes on disk silently diverged from the diff the user approved. Replacements are now spliced by index, so the approved preview - in the terminal and over ACP - is what lands. Closes #1057.
--- a/source/acp/acp-tool-call.spec.ts
+++ b/source/acp/acp-tool-call.spec.ts
@@ -222,3 +222,24 @@ test('buildToolCallMeta - withDiff true is the default', async t => {
 		rmSync(dir, {recursive: true, force: true});
 	}
 });
+
+test('buildToolCallMeta - string_replace diff shows $ tokens literally', async t => {
+	const dir = mkdtempSync(join(tmpdir(), 'acp-tc-'));
+	const file = join(dir, 'run.sh');
+	writeFileSync(file, '#!/bin/sh\necho "old"\nexit 0\n');
+	const replacement = 'echo "pid=$$ match=$& pre=$` post=$\'"';
+	try {
+		const meta = await buildToolCallMeta(
+			makeCall('string_replace', {
+				path: file,
+				old_str: 'echo "old"',
+				new_str: replacement,
+			}),
+		);
+		// The previewed diff has to be the diff that lands on disk.
+		const diff = meta.content[0] as any;
+		t.is(diff.newText, `#!/bin/sh\n${replacement}\nexit 0\n`);
+	} finally {
+		rmSync(dir, {recursive: true, force: true});
+	}
+});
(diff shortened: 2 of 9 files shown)


will-lamerton:
Great work as always @addyCooks :)

addyCooks:
Thank you so much, @will-lamerton!
Kind words are always an encouragement : )

PR #1087, opened 2026-08-30
fix: block fetch_url from loopback aliases and cloud metadata

### LINKED ISSUE #1088: [Bug] fetch_url allows loopback aliases and cloud metadata hosts
## Description

`fetch_url` is read-only and meant for public docs. The validator is a hostname string list: `localhost`, `127.0.0.1`, `0.0.0.0`, a few IPv6 loopback spellings, plus `10.*` / `192.168.*` / `172.16–31.*`. Tests only cover that list.

Node already rewrites `http://127.1` to `127.0.0.1` (blocked). These still pass validation and are fetched:

- `http://127.0.0.2` — rest of `127.0.0.0/8` is loopback
- `http://169.254.169.254/latest/meta-data/` — AWS / GCP / Azure instance metadata
- `http://metadata.google.internal` — GCE metadata name
- `http://[::ffff:127.0.0.2]` — IPv4-mapped loopback (`[::ffff:7f00:2]`)
- `http://localhost.` / `http://metadata.google.internal.` — trailing-dot FQDNs (Node does not strip them)

Execute only does `new URL`, then `convertToMarkdown`. The validator is the only gate; skip it and the fetch still happens.

`startsWith('10.')` also blocks `10.example.com`, a public name.

Closed #734 only added `[::1]` on the initial hostname. This is the rest of the initial-URL list, not HTTP redirects.

## Environment

- **OS**: Any
- **Node version**: Any
- **Nanocoder version**: main
- **Provider**: N/A
- **Model**: N/A

## Steps to Reproduce

1. Open `source/tools/fetch-url.tsx` `fetchUrlValidator`.
2. Call `fetch_url` with `url: "http://127.0.0.2/"` (or `http://169.254.169.254/latest/meta-data/`, or `http://metadata.google.internal`).
3. Validator returns `{valid: true}` and the tool fetches.

## Expected Behavior

Reject loopback (all of `127.0.0.0/8`), link-local / metadata (`169.254.0.0/16`, `metadata.google.internal`), RFC1918, CGNAT, and matching IPv6 / IPv4-mapped forms. Validator and execute should share one check. `10.example.com` should be allowed.

## Actual Behavior

Only the listed hostname strings are blocked. `127.0.0.2` and cloud metadata URLs fetch. `10.example.com` is rejected.

## Additional Context

HTTP redirects from a public URL to loopback/metadata are a **separate** bug (follow `Location` with no second check). Do not treat this issue as covering redirects or DNS rebinding.

Dhirenderchoudhary:
## Description

`fetch_url` only blocked a hostname string list (`localhost`, `127.0.0.1`, a few IPv6 loopback spellings, `10.*` / `192.168.*` / `172.16–31.*`). `http://127.0.0.2`, `http://169.254.169.254` (instance metadata), `http://metadata.google.internal`, and IPv4-mapped loopback still fetched. Execute only parsed the URL; the validator was the only gate.

Shared `assertPublicHttpUrl`: http(s) only, then CIDR for loopback / RFC1918 / link-local / CGNAT, plus `localhost` and `metadata.google.internal`. Validator and execute both use it. Trailing-dot FQDNs (`localhost.`, `metadata.google.internal.`) are stripped so they match. `10.example.com` is allowed again (the old `10.*` prefix was wrong).

Redirects to private hosts and DNS rebinding are **not** in this PR.

Closes #1088 .

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Changeset

### Automated Tests

`pnpm exec ava source/tools/fetch-url-guard.spec.ts source/tools/fetch-url.spec.tsx`

### Manual Testing

N/A — validator/execute reject before fetch. No provider needed.

## Checklist

- [x] Self-review completed
- [ ] Appropriate logging added using structured logging (see CONTRIBUTING.md)


will-lamerton:
Hey @Dhirenderchoudhary - mark this as ready to review whenever you want me to take a proper look :)

Dhirenderchoudhary:
@will-lamerton Sorry the PR is ready

will-lamerton (changes requested):
Good change, and it closes more than the description claims. `getToolValidator` has only two call sites (`tool-confirmation.tsx:84`, `streaming-bash-tool.tsx:42`), and `fetch_url` is `readOnly: true` with no confirmation, so in auto-accept / yolo / headless / subagent runs the validator never ran at all. Moving the check into `executeFetchUrl` is the real fix. Worth saying that in the changeset.

I verified the guard against the normalization tricks and it holds up: `http://2130706433`, `0177.0.0.1`, `127.1`, `2852039166`, `[::ffff:127.0.0.2]`, userinfo and ports are all blocked; `10.example.com` and `172.32.0.1` correctly allowed. Specs, biome, tsc and knip all pass locally.

Two things I'd like before merge:

**1. `metadata.goog` and bare `metadata` are not blocked.** Both are live aliases for the GCP metadata server alongside `metadata.google.internal` (`metadata` resolves via the VM search domain). Since blocking cloud metadata is the stated goal of the PR, this leaves the main target reachable:

```ts
const METADATA_HOSTS = new Set([
	'localhost',
	'metadata',
	'metadata.goog',
	'metadata.google.internal',
]);
if (METADATA_HOSTS.has(host)) return true;
```

**2. Document the redirect gap in the code and soften the changeset.** `@nanocollective/get-md` fetches with `redirect: "follow"` (`dist/cli.js:406`), so `http://attacker.example/r` -> `http://169.254.169.254/` still gets fetched and returned. Treating redirects and DNS rebinding as out of scope is fine, but the changeset currently reads as if metadata is fully blocked. A comment at the top of `fetch-url-guard.ts` naming both known gaps, plus a tracking issue, will stop the next reader assuming this file is the whole story.

Non-blocking notes:

- Still allowed: `fec0::/10`, `ff00::/8` (e.g. `ff02::1`), 6to4 `2002:7f00:1::`, NAT64 `64:ff9b::7f00:1`, `198.18.0.0/15`, `192.0.0.0/24`, `240.0.0.0/4`, `255.255.255.255`. The two IPv6 prefixes are cheap to add if you want them; the rest I'd skip.
- Test additions: an execute-path case for `metadata.google.internal` (only `127.0.0.2` is covered there), and one for `metadata.goog` if you take item 1.
- Consider a follow-up config escape hatch. "Read my dev server at localhost:3000" is now hard-blocked everywhere with no override.

Style and error messages look right: guard call is correctly outside the `try` so it isn't wrapped in `Failed to fetch URL:`, and `formatError(error)` preserves the exact prior message strings for both the format and protocol cases.

will-lamerton:

PR #1090, opened 2026-08-31
fix(fetch_url): block unsafe HTTP redirects

### LINKED ISSUE #1089: [Bug] fetch_url follows HTTP redirects to private/loopback hosts
## Description

`fetchUrlValidator` only inspects the **request** URL hostname, then `executeFetchUrl` calls `@nanocollective/get-md` `convertToMarkdown(url)`. That fetcher uses `redirect: "follow"`. A public URL that 302s to `http://127.0.0.1/`, `http://169.254.169.254/`, or RFC1918 is fetched with no second check.

`fetch_url` is `readOnly` / no approval.

This is **not** #1088. #1088 / PR #1087 only check the **first** URL (`127.0.0.2`, metadata host, etc.). Even after that lands, a public URL that redirects to loopback/metadata still fetches.

Closed #734 only added `[::1]` on the *initial* hostname.

## Environment

- **OS**: Any
- **Node version**: Any
- **Nanocoder version**: main
- **Provider**: N/A
- **Model**: N/A

## Steps to Reproduce

1. Host that returns `302 Location: http://127.0.0.1:<port>/` (or metadata IP).
2. `fetch_url` on `https://that-public-host/...` in yolo / auto-accept.
3. Validator allows the public name; get-md follows; loopback/metadata body is returned to the model.

## Expected Behavior

Do not follow redirects to hosts that would fail the same public-URL check. (Or disable automatic redirects and re-validate each Location.)

## Actual Behavior

One check on the first URL; then unrestricted follow.

## Additional Context

Related to #1088 (initial URL only). Do not close this with PR #1087.

No DNS pinning either (a public name resolving to loopback). That can be the same issue or a follow-up; v1 is re-validate redirect targets.

mikemikimike:
## Description

Fixes #1089.

`fetch_url` is a no-approval tool, but its markdown conversion dependency followed HTTP redirects automatically. A public URL could therefore redirect the process to a loopback, private-network, or cloud-metadata address after the initial URL check.

This change disables automatic redirects for `fetch_url`. Redirect responses are surfaced as fetch failures instead of being followed without validation; callers can submit a destination again only after it passes the existing URL validator.

### Implementation plan

1. Keep the existing validation for the URL supplied to `fetch_url`.
2. Disable automatic redirects in the markdown-fetching call so redirect destinations cannot bypass that validation.
3. Add a deterministic regression test proving that a redirect response targeting loopback is not followed.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Changeset

### Automated Tests

- [ ] New features include passing tests in `.spec.ts/tsx` files (not applicable; this is a bug fix)
- [ ] All existing tests pass (`pnpm test:all` could not complete on this Windows host)

Executed checks:

- `pnpm run test:ava source/tools/fetch-url.spec.tsx` (23 passed)
- `pnpm run test:format` (passed)
- `pnpm run test:types` (passed)
- `pnpm run test:lint` (passed)
- `pnpm run test:changesets` (passed)
- `pnpm audit --audit-level=high --ignore-unfixable` (no new vulnerabilities ignored)

The repository's aggregate `test:all` entrypoint is Unix-only on Windows (`./scripts/test.sh`), so its direct invocation cannot run on this host. The AVA suite was also attempted after dependency installation; unrelated CLI/ACP integration tests require generated artifacts and POSIX `npx` behavior in this environment.

### Manual Testing

- [ ] Tested with Ollama
- [ ] Tested with OpenRouter

These manual provider and MCP checks are not applicable to the isolated `fetch_url` redirect behavior.

## Checklist

- [x] Self-review completed

No documentation or logging changes are needed for this behavior fix.


will-lamerton (dismissed):
Duplicate submission caused by a transient GitHub API error. See the full review below.

will-lamerton (changes requested):
The fix is correct and does close the vulnerability, but it implements only half of what #1089 asked for, and the skipped half is the one that keeps the tool usable.

**Verified working:** `followRedirects` is a real option in `@nanocollective/get-md@1.7.0` and threads through to `redirect: 'manual'` in `dist/utils/url-fetcher.js:122`. The guard isn't bypassable, since `fetchUrlValidator` runs via `withValidation` in `tool-registry.ts`, so the interactive loop, `--plain` shell, and subagents all hit it. Against a live local redirect server, a `302 Location: http://127.0.0.1:1/secret` is genuinely not followed. All 23 tests pass, CI is green, changeset name is correct.

### Blocking: every legitimate redirect now fails, and the model can't recover

Same probe, ordinary redirect:

```
/redirect  => ERR: Failed to fetch URL: HTTP 302: Found              (the attack, correctly blocked)
/www       => ERR: Failed to fetch URL: HTTP 301: Moved Permanently  (benign redirect, also blocked)
/final     => OK
```

The PR description says callers can submit the destination again, but the `Location` header is never surfaced, so the model has no way to learn where to go. http to https upgrades, bare to www, trailing-slash normalization, and link shorteners cover a large share of real URLs, so `fetch_url` now fails on many ordinary docs pages with an error nothing can act on.

#1089 asked for exactly the better option: "disable automatic redirects and re-validate each Location". This takes the disable clause and drops the re-validate clause.

Suggested shape in `executeFetchUrl`: resolve the chain manually with `fetch(url, {redirect: 'manual'})` for up to ~5 hops, run each `Location` (resolved against the current URL) through `fetchUrlValidator`, then pass the final validated URL to `convertToMarkdown(finalUrl, {followRedirects: false})`. Satisfies the issue as written and keeps normal URLs working.

### Smaller points

- Double prefix: get-md already throws `Failed to fetch URL: ...` and the catch at `source/tools/fetch-url.tsx:47` prefixes it again. Pre-existing, but this change makes it the common path.
- The new test asserts `requests[0]?.redirect === 'manual'`, which pins get-md's internals rather than our security property. The mock discards `_input`, so it never asserts no request reached `127.0.0.1`. Capturing the input URL would test the actual contract and survive get-md refactors.
- The changeset says this protects against "redirect-based access to private and loopback addresses", which is accurate for the redirect vector, but a direct fetch to `169.254.169.254`, `127.0.0.2`, `100.64.0.1`, `[fd00::1]`, `[::ffff:10.0.0.1]`, or `metadata.google.internal` still passes the validator until #1087 lands. No conflict, just worth not over-claiming.
- #1089 also flags DNS pinning as a follow-up. Reasonably out of scope here, but worth a tracked issue rather than letting the close absorb it.

### Recommendation

Per-hop re-validation instead of a blanket block. If you'd rather ship the blunt version now, it should at least carry the `Location` in the error message, plus a tracked follow-up, since as written it trades a security bug for a broad functional regression.

will-lamerton (approved):
All points from the previous review are addressed. I checked out the head and probed the real code path against a live redirect server rather than relying on the tests alone.

**Blocking point (per-hop re-validation).** `resolveSafeRedirects` does the suggested shape: manual-redirect probe, `Location` resolved against the current URL, each hop through `fetchUrlValidator`, capped at 5, then the validated final URL handed to `convertToMarkdown(..., {followRedirects: false})`. Against a live server, main vs this branch:

```
             main                                 this branch
/rel     OK  (followed)                       OK  (followed, validated)
/redirect OK (followed)                       OK  (followed, validated)
/evil    OK  "SECRET LOOPBACK DATA"  <-- SSRF ERR Cannot fetch from internal/private network address: 127.0.0.1
/loop    ERR fetch failed (~60 hits)          ERR Too many redirects while fetching ...
/noloc   ERR HTTP 302: Found                  ERR Redirect response from ... did not include a Location header
```

The loopback host is never contacted (server log shows one hit to `/evil`, none to `/secret`). Benign redirects work again, including the live `http://github.com/nano-collective` case covering both an http to https upgrade and a host change. That was the whole objection last time: the previous revision failed both.

**Double prefix.** Fixed, and it holds for genuine get-md errors: a 404 comes back as `Failed to fetch URL: HTTP 404: Not Found`, single prefix.

**Test pinned get-md internals.** The redirect test now captures request URLs, asserts `requests.length === 1`, and asserts nothing reached `127.0.0.1`. That is the real contract.

**Changeset.** Re-reading it, "redirect-based access" is explicitly scoped to the redirect vector, so it does not over-claim. Withdrawing that point.

### Merge conflict resolution

#1087 landed on main in the meantime and both sides touched this file. Resolved in the merge commit on this branch:

- `fetch-url.spec.tsx` conflict was purely additive on both sides. Kept all three tests.
- `fetch-url.tsx` auto-merged correctly. `fetchUrlValidator` is now a thin wrapper over `assertPublicHttpUrl`, so `resolveSafeRedirects` picks up #1087's stronger checks for free. Verified live: redirects to `169.254.169.254`, `127.0.0.2`, `metadata.google.internal`, and `api.localhost` are all rejected at the hop, with no request leaving for the blocked destination.
- Updated the `fetch-url-guard.ts` header, which still described the redirect gap as open and pointed at #1089.

26 tests in `fetch-url.spec.tsx` plus 3 in `fetch-url-guard.spec.ts` pass; format, lint and types clean.

### Notes for follow-up, not blocking

Every successful fetch now issues two GETs of the final URL, the probe and then get-md. Unavoidable without get-md exposing its resolved URL, but worth knowing it is doubled origin load.

The probe uses Node's default fetch, so a different User-Agent and no proxy config versus get-md's fetcher. A UA-gated site can send the two requests down different chains. This fails safe rather than open, since get-md has `followRedirects: false` and just errors on an unexpected redirect.

DNS rebinding is still open on both the initial URL and every hop, since the check is hostname-string and parsed-IP rather than a pinned lookup. I will open a separate issue for it so the #1089 close does not absorb it.

PR #1091, opened 2026-08-31
add fix(lsp): resolve verification promise on early process exit to p…

### LINKED ISSUE #1037: [Bug] verifyLSPServerWithCommunication can hang forever if the child process fails immediately
## Description

When discovering LSP servers, if the child process exits very quickly before the `spawn` event fires, the promise in `verifyLSPServerWithCommunication` never resolves or rejects. This leads to a hanging promise that blocks further LSP discovery indefinitely.

## Environment

- **OS**: Any
- **Node version**: Node 18+

## Steps to Reproduce

1. Configure an LSP server that crashes immediately on startup.
2. Nanocoder attempts to discover/verify it.
3. The promise hangs forever because the `exit` handler fires but does not resolve the promise.

## Expected Behavior

If the child process exits, the verification should resolve to `false` and allow discovery to continue with the next server.

## Actual Behavior

The `exit` event listener clears the timeout but forgets to call `resolve(false)`.

## Additional Context

## Implementation Notes

Fix in `source/lsp/server-discovery.ts` lines 326-331. Add `resolve(false)` inside the `exit` handler.

AniketR10:
## Description

Fixes a hang in `verifyLSPServerWithCommunication` (source/lsp/server-discovery.ts): the `exit` handler cleared the timeout but never resolved the promise, so LSP discovery could stall if a server's process exited before the `spawn` event fired. The handler now calls `resolve(false)`.

Fixes #1037

## Type of Change

- [x] Bug fix

## Changeset

- [x] Ran `pnpm changeset --empty` (internal fix, no changelog entry needed)

### Automated Tests

- [x] All existing tests pass (`pnpm run test:ava source/lsp/server-discovery.spec.ts`)

### Manual Testing

- [x] Verified via a standalone reproduction (mocked child process emitting only `exit`): confirmed the promise hung before the fix and resolves to `false` immediately after
- [x] Ran the real `discoverLanguageServers()` end-to-end against a fake crashing binary on `PATH` — no hang, correct output

## Checklist

- [x] Self-review completed
- [x] No breaking changes

### DIFF
--- a/source/lsp/server-discovery.ts
+++ b/source/lsp/server-discovery.ts
@@ -328,6 +328,7 @@ function verifyLSPServerWithCommunication(
 			// A clean exit can also indicate success for some servers
 			// However, for LSP servers waiting for input, an immediate exit is often a failure
 			// The 'spawn' event is a more reliable indicator for our purpose
+			resolve(false);
 		});
 	});
 }


AniketR10:
@akramcodez @will-lamerton please reivew, small change

will-lamerton:
Nice work @AniketR10 :)

PR #1100, opened 2026-08-31
feat(vscode): add token usage visibility setting

### LINKED ISSUE #1096: [Feature] Add setting to toggle Token/Cost counter visibility (default to off)
## Description

The chat panel currently displays a token usage and cost estimator (e.g., `Tokens: 7k | <$0.01`) by default. While this information is useful for some, it adds visual noise to the screen for users who don't actively track their API costs or use local models (where costs are mostly irrelevant). 

We should add an option in the settings to toggle this UI element on or off.

## Use Case

Many users prefer a clean, minimalistic interface without unnecessary statistics cluttering the screen. Since local model users (like Ollama) don't incur per-token API costs anyway, the token counter is often irrelevant. Allowing users to hide it reduces visual noise.

## Proposed Solution

1. **Add a new VS Code setting:** Create a new boolean configuration setting (e.g., `nanocoder.showTokenUsage`).
2. **Default to Off:** Set the default value to `false` so the UI remains as clean as possible out-of-the-box.
3. **Conditionally render the UI:** Update the chat panel UI to only render the token/cost badge if this setting is enabled.

## Alternatives Considered

- **Show on hover:** Keep the element in the DOM but only reveal the token/cost stats when the user hovers over a small info icon (though a setting toggle is still better for those who never want to see it).
- **Auto-hide for local models:** Automatically hide the cost/token counter when a local provider (like Ollama) is selected, but this doesn't solve the visual noise for cloud API users who still want a clean screen. A global toggle is the best approach.

## Additional Context

*(Note: See attached screenshot of the token counter that causes the visual noise).*

Image

- Add `"nanocoder.showTokenUsage": { "type": "boolean", "default": false, "description": "Show token usage and estimated cost in the chat panel" }` to the `contributes.configuration.properties` in `plugins/vscode/package.json`.
- Expose this setting through the `SettingsManager` in `plugins/vscode/src/settings-manager.ts`.
- In the React webview (`plugins/vscode/media/chat-panel.js` or equivalent), read this setting from the state and conditionally render the token counter component.

macjayz:
## Summary
- add a `nanocoder.showTokenUsage` VS Code setting, defaulting to false
- sync the setting into the chat webview when it becomes ready and when settings are refreshed
- hide the token/cost footer unless the setting is enabled

Fixes #1096

## Testing
- pnpm run format:check
- pnpm run test:types
- pnpm run test:types:vscode
- pnpm exec ava plugins/vscode/src/settings-manager.spec.ts source/vscode/chat-panel-turn-footer.spec.ts

### DIFF
--- a/.changeset/vscode-token-usage-toggle.md
+++ b/.changeset/vscode-token-usage-toggle.md
@@ -1,0 +1,5 @@
+---
+"@nanocollective/nanocoder": patch
+---
+
+Added a `nanocoder.showTokenUsage` VS Code setting that defaults off and hides token usage plus estimated cost footers unless users opt in. Closes #1096.
--- a/.changeset/vscode-token-usage-toggle.md
+++ b/.changeset/vscode-token-usage-toggle.md
@@ -1,5 +1,5 @@
 ---
-"@nanocollective/nanocoder": patch
+"@nanocollective/nanocoder": minor
 ---
 
 Added a `nanocoder.showTokenUsage` VS Code setting that defaults off and hides token usage plus estimated cost footers unless users opt in. Closes #1096.
--- a/plugins/vscode/media/chat-panel.html
+++ b/plugins/vscode/media/chat-panel.html
@@ -182,6 +182,19 @@
 							</label>
 						</div>
 					</div>
+
+					<div class="settings-section">
+						<div class="settings-row">
+							<div class="settings-row-info">
+								<span class="settings-row-label">Token Usage</span>
+								<span class="settings-row-desc">Show token usage and estimated cost below responses</span>
+							</div>
+							<label class="settings-toggle">
+								<input type="checkbox" id="setting-showTokenUsage">
+								<span class="settings-toggle-slider"></span>
+							</label>
+						</div>
+					</div>
 				</div>
 
 				<!-- Advanced Tab -->
--- a/plugins/vscode/media/chat-panel.js
+++ b/plugins/vscode/media/chat-panel.js
@@ -38,6 +38,7 @@
 	
 	let pendingImages = [];
 	let pendingUserMessageText = null;
+	let showTokenUsage = false;
 
 	// ── Slash command autocomplete state ────────────────────
 	const slashDropdown = document.getElementById('slash-dropdown');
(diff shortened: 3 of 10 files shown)


will-lamerton (changes requested):
Approach looks right and the webview-side gating is clean. A few things to fix before merge.

**Blocking**

1. Unhandled rejection on the update path. `_handleUpdateSetting` is now `async` but `case 'updateSetting'` calls it without `await`/`.catch()`, and `_updateShowTokenUsage` has no `try`/`catch`. If `config.update` rejects, the webview never gets a `settingsUpdated` message, no error toast fires, and the promise rejects unhandled. `SettingsManager.updateSetting` already wraps its body and returns `{success: false, error}`; mirror that.

2. Always writing `ConfigurationTarget.Global`. The setting has no declared `scope`, so it defaults to `window` and can be set in `.vscode/settings.json`. When a workspace value exists, `get()` returns it, the panel writes Global, the effective value never changes, and the toggle silently snaps back with no error. Use `config.inspect()` to pick the narrowest scope that already holds a value, falling back to Global. Also worth deciding deliberately: every other toggle in this panel persists per-project, so this one changing behaviour for all workspaces is an inconsistency.

3. No `onDidChangeConfiguration` listener. It is a contributed setting now, so users will flip it in the Settings editor, and nothing pushes that to a live webview. It stays stale until reload or until the in-panel settings view is opened. `extension.ts:130` already has this pattern for `nanocoder.codeLens`; calling `_handleRequestSettings()` on `affectsConfiguration('nanocoder.showTokenUsage')` closes it.

**Non-blocking**

4. `SettingsManager.readSettings` hardcodes `showTokenUsage: false` and the provider immediately overwrites it, so the field on `SettingsData` is wrong by construction for any other caller. Prefer leaving it off `SettingsData` and typing the webview payload as `SettingsData & {showTokenUsage: boolean}`.

5. The `settings-manager.spec.ts` assertion tests that hardcoded constant, so it passes whether or not the feature works. Drop it or move it to whatever owns the real value. The `chat-panel-turn-footer.spec.ts` test is good; the missing case is enabled-then-disabled.

6. Toggling off leaves already-rendered footers on screen (and toggling on does not back-fill). Since the issue is about visual noise, hiding existing indicators in `renderSettingsData` would match expectations.

7. Changeset is `patch` for a user-facing feature; this repo uses `minor` for those (see `anthropic-prompt-caching`, `calm-work-summaries`).

8. Minor: `case 'ready'` now does a full config-file read plus a whole `renderSettingsData` pass to learn one boolean.

will-lamerton:
Thanks for this PR @macjayz  - feel free to add yourself as a contributor to our website via a PR which I will approve :)

Issue #155, opened 2025-12-20
[Feature] Implement Enhanced PR Management and Controlled Release Pipeline


Avtrkrb:
## Description

This feature request proposes implementing a comprehensive PR management and release pipeline system for NanoCoder to improve code quality, security, and release reliability. The implementation consists of four main components:

1. **Branch Protection Rules**: Requiring 2+ maintainer approvals for PR merges
2. **Enhanced PR Checks Workflow**: Comprehensive automated testing and validation
3. **Status Badge Pipeline**: Dynamic badge generation and updates
4. **Controlled Release Pipeline**: Secure, automated multi-platform publishing

### Current Challenges
- Inconsistent code quality checks across different environments
- Manual release process is error-prone and time-consuming
- Project health status is not easily visible to contributors and users

### Proposed Solutions Address

1. **Code Quality & Security**
   - Ensures every PR undergoes proper review by multiple maintainers
   - Automated checks prevent breaking changes and maintain code standards
   - Example: A PR introducing a security vulnerability would be caught by automated checks and require multiple maintainer reviews before approval

2. **Developer Experience**
   - Clear visibility into PR status and requirements
   - Consistent feedback loop for contributors
   - Example: New contributors can immediately see what checks failed and how to fix them

3. **Release Reliability**
   - Prevents accidental releases
   - Ensures all platforms are updated consistently
   - Example: When releasing v1.20.0, NPM, Nix, and Homebrew packages are updated simultaneously

4. **Project Visibility**
   - Real-time status badges show project health
   - Automated announcements reach all communities
   - Example: Users can see at a glance if the main branch is stable and what the latest version is

### Component 1: Branch Protection Rules

**Implementation**: GitHub Branch Protection Settings
**Configuration**:
```yaml
Branch: main
- Require pull request reviews before merging
  - Required approving reviews: 2
  - Dismiss stale PR approvals when new commits are pushed
  - Require review from CODEOWNERS
  - Restrict reviews to users who dismiss stale reviews
- Require status checks to pass before merging
  - Required status checks:
    - lint (ubuntu-latest, node-20)
    - unit-tests (ubuntu-latest, node-20)
    - build-verification (ubuntu-latest, node-20)
    - dependency-analysis (ubuntu-latest, node-20)
    - security-audit (ubuntu-latest, node-20)
- Require branches to be up to date before merging
- Restrict pushes that create matching branches
- Allow force pushes: false
```

### Component 2: Enhanced PR Checks Workflow

[...]

PR #1101, opened 2026-08-31
ci: pass fail-on-coverage-drop to shared PR checks

### LINKED ISSUE #1052: [Feature] Fail PRs that drop coverage vs the base branch
## Description

`pr-checks.yml` sets `FAIL_ON_COVERAGE_DROP: 'true'` and never reads it.

The `test-coverage` job only fails if total line coverage is under 80% (`ENABLE_COVERAGE_THRESHOLD`). A PR can delete tests, drop 10 points, and still pass if it stays above 80.

#155 asked for a drop-vs-base gate. The env var is the leftover. Coverage floor can stay.

## Use Case

A PR removes assertions or untested code in a way that coverage falls relative to `main`, but still sits at 81%. Today CI is green. Maintainers only notice in review.

## Proposed Solution

In `test-coverage` (`.github/workflows/pr-checks.yml`):

1. Keep the 80% floor.
2. Use `FAIL_ON_COVERAGE_DROP`: compute coverage on `github.event.pull_request.base.sha` (or a compact summary from that ref) and on HEAD. Fail if HEAD dropped more than a small delta (e.g. 1.0).
3. Skip the drop check when there is no base summary (first run / missing file), so the job does not flake.
4. Print both percentages in the log.

No new third-party Codecov account. This is local `c8` JSON, same as today.

`pnpm changeset --empty` (CI).

## Alternatives Considered

- Codecov upload from #155. Extra token and vendor. Not needed if we compare two local summaries.
- OS/node matrix. Unrelated, expensive.

## Additional Context

- [x] I have searched existing issues (#155 closed; this is the unused env var)

Dhirenderchoudhary:
## Description

After #1073, PR checks live in the org workflow. That workflow already has `fail-on-coverage-drop` (default true): it compares this PR's c8 line % to `badges/coverage.svg` on the base branch, prints both in the job summary, and skips if the badge is missing. No Codecov.

This repo now passes `fail-on-coverage-drop: true` on the reusable workflow so the leftover from #155 is explicit here.

Closes #1052.

## Type of Change

- [x] New feature

## Changeset

Empty changeset (CI chore).


will-lamerton:

PR #1103, opened 2026-08-31
fix(export): name the reason a path was rejected, split writeUniqueFile out

will-lamerton:
## Description

Follow-up to #956 (merged as e0548372), picking up the review nits that were left open plus one thing the merged code got wrong.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Changeset

- [x] Added a changeset (amended the existing unreleased `fiery-hats-pick.md` from #956 rather than adding a second one for the same feature)

## Changes

**Rejection messages named the wrong cause.** Every failed path resolution rendered `Invalid export path: outside the project directory`, discarding the real reason. For an export filename containing a null byte that message was flatly untrue. `explainInvalidPath` now re-derives the specific cause, because `resolveFilePath` collapses several distinct failures into one generic `Invalid file path`:

```
/export ~/notes.md
  Invalid export path: '~' is not expanded; use a path relative to the
  project, or an absolute path inside it

/export ../../outside.md
  Invalid export path: '..' segments are not allowed; exports stay inside
  the project

/export /tmp/chat.md
  Invalid export path: it is outside the project directory (/path/to/project)
```

**`writeUniqueFile` moved to `source/utils/write-unique-file.ts`.** It is a generic path helper with nothing export-specific about it, so `generate-export-filename.ts` was a hard place to find it. Its tests moved with it, and its one export-specific error string is now generic.

**Containment is now documented as deliberate.** `/export ~/notes.md` and `/export /tmp/chat.md` wrote wherever you pointed them before #956 and are refused now. That narrowing is intentional and matches `read_file` / `write_file` / `string_replace`, but nothing said so. The changeset now calls it out, and the comment on `explainInvalidPath` records the reasoning.

**Smaller points**

- Recorded that the export date is deliberately UTC (`toISOString`) rather than local, with the reasoning.
- Two export specs created directories inside the working tree and cleaned up with a trailing `fs.rm`, so a mid-test failure left `tmp-session-cwd-test` / `tmp-prjroot-test` behind as untracked dirs. Now `t.teardown`, registered up front.
- Changeset bumped from `patch` to `minor` - #956 is a feature.
- Fixed a stray indented `test(` in `export.spec.tsx`. Worth knowing separately: `biome.json` excludes `**/*.spec.ts` and `**/*.spec.tsx` from `files.includes`, so no format or lint check covers spec files at all. That is why it passed CI. Not changing it here, but it is a real gap.

### Automated Tests

```
pnpm run test:types  - passes
pnpm run test:knip   - clean
npx ava source/utils/generate-export-filename.spec.ts source/utils/write-unique-file.spec.ts source/commands/export.spec.tsx
  - 50 tests passed
```

The rejection specs previously asserted only `/Invalid export path/`, which passed no matter which cause was reported. They now assert the specific message, so a regression back to the generic string fails the suite.

### Manual Testing

- [ ] Tested with Ollama
- [ ] Tested with OpenRouter

## Checklist

- [x] Self-review completed

https://claude.ai/code/session_01KjG4A5ao56WdZYjDHJjMch

### DIFF
--- a/.changeset/fiery-hats-pick.md
+++ b/.changeset/fiery-hats-pick.md
@@ -1,5 +1,7 @@
 ---
-"@nanocollective/nanocoder": patch
+"@nanocollective/nanocoder": minor
 ---
 
 Auto-generate descriptive filenames for /export instead of generic timestamps. Closes #934
+
+Exports are now contained to the project directory, matching read_file / write_file / string_replace: `~` is not expanded and absolute paths outside the project root are refused rather than written. Rejections name the specific cause (null byte, `~`, `..` segment, outside the root) instead of failing generically.
--- a/.changeset/ripgrep-file-search.md
+++ b/.changeset/ripgrep-file-search.md
@@ -5,3 +5,5 @@
 File search (path matching and content search) is now backed by `ripgrep` instead of a hand-rolled JS walker.
 
 Search also respects `.nanocoderignore` and binary files again, matching `list_directory` and file autocomplete.
+
+A failed search now reports the failure instead of returning an empty result set.
--- a/source/commands/export.spec.tsx
+++ b/source/commands/export.spec.tsx
@@ -257,11 +257,11 @@ test('exportCommand rejects path traversal in filename', async t => {
 	const {lastFrame} = render(<MockThemeProvider>{result}</MockThemeProvider>);
 	const output = lastFrame();
 	t.truthy(output);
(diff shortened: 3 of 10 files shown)


will-lamerton:
Pushed directly to main as e1af6d9e instead. This branch also unintentionally included the ripgrep commit, which landed separately as #1102.

PR #1106, opened 2026-08-31
fix(vscode): improve composer mode selector layout

### LINKED ISSUE #1095: [Feature] Move Mode selector next to Model dropdown and improve its label/icon
## Description

In the chat input area, the mode selector (currently displaying "auto..." with a gear icon) is positioned on the far right, directly next to the Send button. Visually and logically, this configuration option belongs next to the Model selector on the left side of the input footer. 

Additionally, the "auto..." text with a gear icon is slightly ambiguous and could be improved for better clarity.

## Use Case

Grouping related configuration options (Model selection and Agent Mode) together on the left creates a cleaner visual hierarchy. It reserves the right side exclusively for the primary "Submit/Send" action, reducing UI clutter near the most clicked button.

## Proposed Solution

1. **Relocate the Button:** Move the mode selector to the left side, immediately following the model dropdown.
   *Layout:* `[ + ] [ deepseek-v4-pro ⌄ ] [ auto... ⚙ ]` ............... `[ Send ➢ ]`

2. **Improve the Label & Icon:**
   Instead of the gear icon and truncated "auto..." text, we should make it look like a cohesive dropdown menu. Here are a few suggestions:
   - **Suggestion A (Match Model Dropdown):** Remove the gear icon entirely. Just show the current mode with a standard dropdown arrow (e.g., `Auto-Accept ⌄` or `Chat ⌄`). This makes it obvious that it's a selectable menu just like the model list.
   - **Suggestion B (Mode-Specific Icons):** Use different icons depending on the active mode. For example:
     - Chat mode: A chat bubble icon `$(comment)`
     - Auto-Accept mode: A magic wand `$(wand)` or a spark/bolt `$(zap)`
     - Plan mode: A checklist icon `$(checklist)`

## Alternatives Considered

- Keeping it on the right but adding a vertical divider `|` to separate it from the Send button. (However, moving it to the left groups the state/configuration elements much better).

## Additional Context

*(Note: See attached screenshot of the current input layout).*

Image

- In the webview's HTML/React layout for the composer, move the mode selector `<button>` out of the right-aligned flex container and into the left-aligned flex container next to the model selector.
- Update the icon mapping for the mode state to reflect the new design choice.

hiarun02:
## Description

Moves the VS Code chat composer mode selector next to the model dropdown and makes mode labels clearer. The mode selector now appears as a normal dropdown label, for example `Auto-Accept`, instead of a cramped/raw mode id near the send button.

Closes #1095

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

## Changeset

No changeset added because this is a small VS Code extension UI polish change.

### Automated Tests

Ran:
- `./node_modules/.bin/ava source/vscode/chat-panel-*.spec.ts`
- `./node_modules/.bin/tsc --noEmit`

### Manual Testing

Attached proof video showing:
- Mode selector appears next to the model dropdown
- Mode selector uses readable labels like `Auto-Accept`
- Dropdown opens and mode switching still works
- Send button remains separated on the right

https://github.com/user-attachments/assets/f5882098-a151-47eb-95de-25e15f5f3187

## Checklist

- [x] Self-review completed

No logging changes were needed because this is a UI-only change.


will-lamerton (changes requested):
Thanks for this - the layout is right and matches Suggestion A from #1095. I ran the specs locally (9/9 composer, 184/184 across `source/vscode/`) and `tsc --noEmit` is clean. A few things before merge.

**Blocking**

1. `source/vscode/chat-panel-harness.ts:34` still lists `composer-mode-badge` in `SHELL_IDS`, but the element is gone. The stub keeps fabricating the node, so a future test can assert against a phantom element and pass. Please drop it.

2. Accessibility regression on the mode control. The old gear carried `aria-haspopup="menu"`, a toggled `aria-expanded`, `aria-controls`, and an `aria-label` that announced the current mode. The new `#mode-trigger` (`chat-panel.html:253`) has none of them, and `syncModeBadge()` overwrites the static `title="Mode"` with the bare label, so the tooltip ends up as just `Auto-Accept` - redundant with the visible text and no longer saying what it controls. A screen reader now hears "Auto-Accept, button". Suggest `Mode: Auto-Accept` for title/aria-label plus the `aria-*` menu attributes.

3. Test gap on the thing the PR adds. The spec asserts the trigger label, but nothing asserts the dropdown *items* are formatted, so `item.textContent = formatDropdownLabel(...)` is uncovered. There is also no guard that model options still shorten at `/`, which this refactor touched. Two assertions closes both.

4. Please add a changeset. This is user-facing behaviour in a shipped extension. Note it has to name `@nanocollective/nanocoder` - a bare `nanocoder` passes PR checks and then breaks release-prepare on main.

**Non-blocking**

5. `syncModeBadge()` no longer syncs a badge, it sets a title. Worth renaming.

6. `formatDropdownLabel(opt, this.triggerId)` is computed twice for the selected option in `setOptions`. Hoist it.

7. Naming: everything else in nanocoder calls this mode "normal" (`DEVELOPMENT_MODE_LABELS` in `source/types/core.ts:259`, `--mode normal`). "Chat" would make the VS Code panel the only surface using that word. #1095 did suggest it, so I am fine either way, but let's make it a deliberate call and follow through on the other surfaces if we adopt it.

8. Bigger picture, for a follow-up rather than here: ACP already carries a human `name` per mode, and we throw it away twice - `source/acp/acp-agent.ts:725` sets `name: id`, then `plugins/vscode/src/acp-client.ts:197` maps to `m.id`. Giving modes real names in `buildModeState` would fix labels for every ACP client, not just this webview, and keep them in one place.

hiarun02:
@will-lamerton Thanks Will, addressed the review items in `10eefd08`.

- Removed the stale `composer-mode-badge` shell id from the webview harness.
- Restored mode trigger accessibility with `aria-haspopup`, `aria-expanded`, `aria-controls`, and dynamic `Mode: ...` title/aria-label text.
- Added test coverage for formatted mode dropdown items and model labels shortening at `/`.
- Updated the mode-change test to select `yolo` instead of re-selecting the current mode.
- Added a changeset for `@nanocollective/nanocoder`.
- Renamed `syncModeBadge()` and hoisted the duplicate selected-option label formatting.
- Kept the `normal` mode label aligned with the rest of Nanocoder by displaying it as `Normal`.

Validation:
- `./node_modules/.bin/ava source/vscode/chat-panel-composer.spec.ts --verbose --timeout=20s` → 10 passed
- Non-server `source/vscode` specs → 188 passed
- `node scripts/validate-changesets.js` → passed
- Biome on touched files and `git diff --check` → passed

Full `source/vscode/*.spec.ts` still cannot complete in my sandbox because `vscode-server.spec.ts` hits localhost `EPERM`, unrelated to this PR.

will-lamerton:
Great work @hiarun02 :)

PR #1123, opened 2026-09-01
feat(config): add `nanocoder config` to show resolved settings and th…

### LINKED ISSUE #1006: [Feature] Config Diff / Effective Config View
## Description

Add a command to view the fully resolved effective configuration and where each value originated.

## Use Case

Configuration is resolved from 4 layers (system -> user -> project -> CLI). When a setting isn't behaving as expected, it's hard to debug which layer is overriding it.

## Proposed Solution

- Implement `nanocoder config show <key>` and `nanocoder config diff`.
- Build on `source/config/index.ts` to output a tree or table showing the final value and the file/layer it was sourced from.

## Alternatives Considered

- Manually checking all 4 config locations.

## Additional Context

AniketR10:
## Description

Adds `nanocoder config` so you can see your settings and which file each one came from.

Settings come from four places: built-in defaults, your global config folder, the project folder, and `NANOCODER_*` environment variables. Before this, finding out which one set a value meant opening all four by hand.

Three commands:

- `nanocoder config list` - every setting, its value, and the file it came from
- `nanocoder config show <key>` 0 one setting in detail, with its default and any values it beat
- `nanocoder config diff` - only what your files change, plus values that are set but not used

All three take `--json`.

**Worth knowing:** settings are grouped into blocks like `autoCompact`. Nanocoder takes the *whole* block from the first file that mentions it - it does not mix fields from two files. So if global sets `threshold` and `notifyUser`, and your project file sets only `threshold`, your `notifyUser` is dropped and the built-in default is used. Nothing told you that before. `config diff` now lists it under "Ignored values".

Two smaller things: values show the way the app really uses them (write `threshold: 200`, see `95`, because it's capped), and API keys show as `<redacted>` so output is safe to paste into an issue.

Closes #1006.

https://github.com/user-attachments/assets/5dd05a99-bd84-42c7-b6aa-a7cc93def02e

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

## Changeset

### Automated Tests

14 new tests in `source/config/effective-config.spec.ts`. `source/config/` passes 270 tests. `test:types`, `test:lint`, `test:format` and `test:knip` are all clean and the build succeeds. I have not run the full `pnpm test:all` end to end.

### Manual Testing

- [ ] Tested with Ollama
- [ ] Tested with OpenRouter

Not applicable as this command only reads config files and prints them. It never contacts a provider. Tested by hand against a real global config plus a project config, checking value, layer, source file, shadowed values, and redaction.

## Checklist

- [x] Self-review completed

No logging needed — the command prints to stdout and exits.

### DIFF
--- a/.changeset/effective-config-command.md
+++ b/.changeset/effective-config-command.md
@@ -1,0 +1,19 @@
+---
+'@nanocollective/nanocoder': minor
+---
+
+Add `nanocoder config` to see your settings and which file each one came from.
+
+Settings come from four places: built-in defaults, your global config folder, the project folder you're in, and `NANOCODER_*` environment variables. When a setting didn't do what you expected, finding out which file set it meant opening all four by hand.
+
+Three commands:
+
+- `nanocoder config list` — every setting, its value, and the file it came from
+- `nanocoder config show <key>` — one setting in detail: its default, and any values it beat
+- `nanocoder config diff` — only what your files change, plus values that are set but unused
+
+All three take `--json`.
+
+**The part worth knowing.** Settings are grouped into blocks, like `autoCompact`. Nanocoder takes the *whole* block from the first file that mentions it — it does not mix fields from two files. So if your global config sets `threshold` and `notifyUser`, and your project config sets only `threshold`, your `notifyUser` is thrown away and the built-in default is used instead. Nothing told you that before. `config diff` now lists it under "Ignored values".
+
+Two smaller things. Values are shown the way the app really uses them — write `threshold: 200` and you'll see `95`, because it's capped at 95. And API keys show as `<redacted>`, so you can paste the output into a bug report.
--- a/docs/configuration/index.md
+++ b/docs/configuration/index.md
@@ -46,6 +46,45 @@ Nanocoder looks for configuration in the following order (first found wins):
 
 > **Tip:** Use `/setup-config` to list all available configuration files and open any of them in your `$EDITOR`.
 
+## Inspecting the Effective Configuration
+
+When a setting is not behaving the way you expect, the hard question is not
+what the value is but which file set it. `nanocoder config` answers both
+without booting the interactive app.
+
+```bash
+nanocoder config list                                   # every resolved key and its layer
+nanocoder config show nanocoder.autoCompact.threshold   # one key, in detail
+nanocoder config show autoCompact                       # a whole block
+nanocoder config diff                                   # only what your files change
+nanocoder config diff --json                            # machine-readable output
+```
+
+`config list` prints the layers in play — built-in defaults, the global config
+directory, the project directory, and `NANOCODER_*` environment variables —
+along with whether each file is present, missing, or unreadable. A `*` beside a
+key means a lower-precedence layer also sets it.
+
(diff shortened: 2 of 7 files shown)


AniketR10:
@will-lamerton @akramcodez please review

akramcodez (dismissed):
Thanks @AniketR10! Overall, the implementation looks good and the PR is in a good condition. These are non-blocking suggestions, but it would be nice to have them included:

1. **MCP server shadowing:** It would be good if `config diff` also reported shadowed MCP servers under ignored values, similar to how provider shadowing is handled.

2. **MCP env origin:** When the configuration comes from `NANOCODER_MCPSERVERS_FILE`, it would be good to report that as the origin instead of always showing `NANOCODER_MCPSERVERS`.

3. **Additional test coverage:** It would be nice to add coverage for the `--json` paths (`config list --json` and `config show <key> --json`) and MCP server credential redaction.

None of these are blockers - the current implementation looks solid. Just some small improvements that would make the PR even more complete.

AniketR10:
@akramcodez  have implemented the suggestions

will-lamerton (changes requested):
Thanks @AniketR10, I went through this and all five review points are addressed:

- Copilot's preferences-ordering issue: `loadPreferences()` now runs before `readRawLayers`, with a regression test asserting the layer report matches disk.
- Copilot's `DEFAULT_SESSION_CONFIG` aliasing: loaders return copies now, and you applied the same fix to the sibling `DEFAULT_*` loaders rather than just the one flagged. Good catch.
- @akramcodez's three: MCP shadowing shows under "Ignored values", `NANOCODER_MCPSERVERS_FILE` is reported as the origin when it is the one in use, and `--json` plus MCP redaction are covered. The redaction test also caught a real gap (`\btoken\b` never matched `GITHUB_TOKEN`), which you fixed.

Two things to sort before merge:

**1. The spec fails on macOS.** Five tests fail locally: `a project value wins and records the global value it shadows`, `an environment variable outranks the config file`, `NANOCODER_CONFIG_DIR marks the project preferences file as skipped`, and both MCP tests. Every failure is the same prefix mismatch:

```
- '/private/var/folders/.../project/.mcp.json'   (actual)
+ '/var/folders/.../project/.mcp.json'           (expected)
```

`join(tmpdir(), ...)` at `effective-config.spec.ts:50` returns `/var/folders/...` on macOS, but `process.chdir()` makes `process.cwd()` resolve to `/private/var/folders/...`, so the fixture's expected paths never match the resolver's output. CI is ubuntu-only so the checks are green and this is invisible there, but it breaks the suite for any Mac contributor. Wrapping the root in `realpathSync` fixes it, plus the same for `serverFile` at line 449.

**2. Rebase needed.** The PR is currently conflicting: a content conflict in `docs/configuration/index.md`. `source/config/index.ts` auto-merges fine.

Nothing wrong with the implementation itself, so once those two are done this is good to go.

AniketR10:
@will-lamerton have fixed both the changes,please review

will-lamerton:
Great work @AniketR10 :)

PR #1124, opened 2026-09-01
fix: quote parameters correctly for cmd.exe custom tools

### LINKED ISSUE #1084: [Bug] Custom tool `{{ }}` quoting is POSIX-only; cmd.exe is not covered

awhite0030:
## Description

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Changeset

### Automated Tests

### Manual Testing

- [ ] Tested with Ollama
- [ ] Tested with OpenRouter

## Checklist

- [ ] Self-review completed

---

Root cause: The custom tool template engine universally applied POSIX single-quote escaping for all arguments (`renderValue` -> `shellQuote`). On Windows `cmd.exe`, this resulted in strings wrapped in literal `'` marks, breaking file paths and failing to prevent shell injection vectors from characters like `&` and `%VAR%`.

Fix: Extracted `isWindowsCmd` from `handler.ts` and introduced a `cmdQuote` function in `template.ts` that escapes metacharacters `%`, `&`, `|`, `<`, `>`, and `^` by prefixing them with `^` and wrapping the argument in double quotes. The template renderer now takes the shell as an optional parameter and uses `cmdQuote` if it detects a Windows `cmd` shell.

Validation:
```
pnpm run build
pnpm test:format
pnpm test:lint
pnpm test:types
pnpm test:knip
pnpm test:ava source/custom-tools/*.spec.ts
```
(Note: Only custom-tools tests were run via `ava` due to container environment timeouts)

Fixes https://github.com/Nano-Collective/nanocoder/issues/1084

Fixes #1084

### DIFF
--- a/.changeset/custom-tool-windows-cmd-quoting.md
+++ b/.changeset/custom-tool-windows-cmd-quoting.md
@@ -1,0 +1,5 @@
+---
+"@nanocollective/nanocoder": patch
+---
+
+Fix custom tool templating to quote parameters correctly under cmd.exe
--- a/source/custom-tools/handler.ts
+++ b/source/custom-tools/handler.ts
@@ -19,10 +19,10 @@ export function buildHandler(
 	projectRoot: string,
 ): ToolHandler {
 	return async (args: Record<string, unknown>): Promise<string> => {
-		const rendered = renderBody(body, args ?? {});
+		const shell = pickShell(metadata.shell);
+		const rendered = renderBody(body, args ?? {}, shell);
 		const cwd = resolveCwd(metadata.cwd, projectRoot);
 		const env = mergeEnv(metadata.env);
-		const shell = pickShell(metadata.shell);
 		return runScript(rendered, {
 			cwd,
 			env,
@@ -195,7 +195,7 @@ export function shellArgs(shell: string, script: string): string[] {
 	return isWindowsCmd(shell) ? ['/d', '/s', '/c', script] : ['-c', script];
 }
 
-function isWindowsCmd(shell: string): boolean {
+export function isWindowsCmd(shell: string): boolean {
 	const name = shell.replaceAll('\\', '/').split('/').pop() ?? '';
 	return /^cmd(\.exe)?$/i.test(name);
 }
--- a/source/custom-tools/template.spec.ts
+++ b/source/custom-tools/template.spec.ts
@@ -1,5 +1,5 @@
 import test from 'ava';
-import {renderBody, renderValue, shellQuote} from './template';
+import {renderBody, renderValue, shellQuote, cmdQuote} from './template';
 
 console.log('\ncustom-tools/template.spec.ts');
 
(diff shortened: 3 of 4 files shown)


will-lamerton (changes requested):
Thanks for picking this up, the direction matches the suggested v1 in #1084. Two things need fixing before this lands.

**1. Caret escaping is inert inside double quotes (`template.ts`, `cmdQuote`)**

`cmdQuote` both caret-escapes `%&|<>^` *and* wraps the result in `"`. In cmd.exe `^` is only an escape character outside quotes; inside `"..."` it is a literal character. So every escaped metacharacter leaks a stray `^` into the value.

Using the example from the issue: tool body `type {{ file }}` with `file = "Q3 & Q4.txt"` renders as

```
type "Q3 ^& Q4.txt"
```

and cmd looks for a file literally named `Q3 ^& Q4.txt`. We have traded the old POSIX bug (`type 'notes.txt'`) for a new one on any value containing a metacharacter.

The new specs lock this in rather than catching it: `cmdQuote('%PATH%')` asserting `"^%PATH^%"` is a string that renders as the literal text `^%PATH^%`, not `%PATH%`.

Pick one mechanism, not both. Inside `"` the separators `& | < >` are already inert and need no caret.

**2. `"` is not in the escape set**

The regex is `/[%&|<>^]/g`, so a value containing a double quote terminates the quoted region early and everything after it parses outside quotes. cmd has no backslash escape for `"` under `cmd /c`, which is why the usual handling is to reject or strip `"` (and `%`) from substituted values rather than claim to escape them. At minimum this needs a spec for a value with an embedded `"`.

**3. Needs a real Windows run**

CI is ubuntu-only on every workflow and the PR checklist is unchecked, so nothing here has executed under cmd.exe. The added tests only assert the shape of the rendered string, which is exactly the part that is wrong. Could you paste one manual `cmd.exe` run showing `&` and `%PATH%` arriving literally?

**Minor**

- Circular import: `template.ts` now imports from `handler.ts`, which imports `renderBody` back from `template.ts`. It survives on function hoisting, but `isWindowsCmd` would be better in a small shared module.
- Stale module doc at `source/custom-tools/template.ts:12-20` still says "All scalar values are passed through `shellQuote()`" and "Under cmd.exe the same quotes are not quoting, so this is not an injection barrier". The second line is the thing this PR claims to fix.
- Branch is 62 commits behind `main`, worth a rebase.

PR #1125, opened 2026-09-01
bugFix(tools): refuse edits to PDF and DOCX documents

### LINKED ISSUE #1058: [Bug] Editing a PDF or DOCX overwrites the document with its markdown transcript
## Description

`getCachedFileContent` is the single read path for the edit tools. For `.pdf` and
`.docx` it does not return the file's bytes it returns a markdown transcript
produced by `convertToMarkdown`. The write path has no matching branch. So
`string_replace`, `diff_edit`, and `write_file` take that transcript, apply the
edit to it, and write it back over the original document as UTF-8.

The document is destroyed. A "fix the typo in the spec" request replaces a PDF
with a few hundred bytes of plain text, and the tool reports success.

## Environment

- **OS**: Windows 11 (path is platform-independent)
- **Node version**: v22.22.3
- **Provider**: any
- **Model**: any

## Steps to Reproduce

1. `read_file` a `.pdf` it returns markdown and marks the path seen.
2. `string_replace` on the same path, using any word from that markdown as `old_str`.
3. Inspect the file.

## Expected Behavior

The edit tools refuse a path whose cached content is a derived representation
rather than the file itself, with an error explaining why.

## Actual Behavior

```
BEFORE: bytes= 733  magic= "%PDF-"
read_file returned markdown, first 60: "---\nwordCount: 2\nreadingTime: 1\n---\n\nHello World\n"
replacing token: "wordCount"
string_replace RESULT: Successfully replaced content at line 2 (now line 2).
AFTER : bytes= 50  magic= "---\nT"
STILL A PDF: false
```

## Logs/Screenshots

- `source/utils/file-cache.ts:150` `if (ext === '.pdf' || ext === '.docx')`, returns `result.markdown` as `content`
- `source/tools/file-ops/string-replace.tsx:92` `writeFile(absPath, newContent, 'utf-8')`
- `source/tools/file-ops/write-file.tsx:40` same, reachable once `markFileSeen` has been set by the read
- `source/tools/file-ops/diff-edit.tsx` same read path, same write

`CachedFile` carries `content`, `lines`, `mtime`, `cachedAt` nothing that tells
a caller the content is derived, so no caller can check.

## Additional Context

Both undo systems are blind to this. `TimelineManager` explicitly skips binaries,
and `FileSnapshotService` stores text, so it would have mangled the snapshot
anyway. There is no restore path the original bytes are gone the moment the
write lands.

Suggested fix have `getCachedFileContent` return a provenance flag
(`derived: true` for converted formats) and have `string_replace`, `diff_edit`,
and `write_file` refuse any path whose content is derived, naming the reason so
the model stops rather than retries. An extension deny-list on the write side is
a cheap defence-in-depth backstop for formats added later.

Related: #962 covers the same class of loss in checkpoints (UTF-8 round-tripping
destroys binaries on save). This is the edit-tool path, which #962 does not touch,
and the two fixes are independent.

addyCooks:
Closes #1058

## Description

`getCachedFileContent` is the single read path for the edit tools, and for `.pdf`
and `.docx` it returns a markdown transcript instead of the file's bytes. The write
side had no matching branch, and `CachedFile` carried nothing that let a caller tell
derived content from real content so no caller could check.

Changes:

- `isDerivedContentPath()` in `file-cache.ts` is now the single source of truth for
  which formats are transcribed. The conversion branch uses it too, so the read and
  write sides cannot drift apart as formats are added. A duplicate `.pdf`/`.docx`
  check in `read-file.tsx` was folded into it.
- `CachedFile` gained a `derived` flag, closing the gap the issue named.
- `validateEditableFormat()` in `path-validators.ts` is the shared guard, wired into
  the validator **and** the executor of all three write tools. The executor copy is
  not redundant: `ToolRegistry.fromToolExports` registers the raw `tool` alongside
  the `withValidation`-wrapped handler, so `execute` is reachable without the
  validator.
- The check is on the extension rather than on cached content, so it also covers a
  fresh `write_file` to a `.pdf` that was never read writing UTF-8 text to those
  paths is never correct.
- The guard fires before `getCachedFileContent`, so a refused edit never pays for a
  document conversion.

Scope note: #962 covers the same class of loss in checkpoints (UTF-8 round-tripping
destroying binaries on save). This PR is the edit-tool path only; the two fixes are
independent.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Changeset

### Automated Tests

### Manual Testing

- [ ] Tested with Ollama
- [ ] Tested with OpenRouter

## Checklist

- [x] Self-review completed


will-lamerton (changes requested):
The fix itself is right. `isDerivedContentPath()` as the single source of truth means the read and write sides can't drift as formats are added, the executor-level check is correctly not redundant given `fromToolExports` exposes the raw `tool`, and firing the guard before `getCachedFileContent` avoids paying for a conversion on a refused edit. Two things before it merges.

**1. `CachedFile.derived` has no consumer.** It's set in `readAndCacheFile` and asserted in the spec, but nothing in production reads it - the actual protection is entirely the extension check in `path-validators.ts`. knip won't flag an unused interface field, so this will sit as dead weight. Either drop it, or say in the PR body that it's deliberately forward-looking API surface.

**2. `path-validators.ts` now imports `file-cache.ts`.** That's a new edge from a leaf validator module into the cache module. Acyclic today, but it means loading the validators drags in the cache. Moving `isDerivedContentPath` and `DERIVED_CONTENT_EXTENSIONS` into a small shared module would keep both leaves clean.

Neither is a correctness problem. The error message is well-aimed - telling the model "do not retry" and naming the alternative is the right shape for stopping a retry loop.

addyCooks:
Thanks @will-lamerton for the review....both fixed:

CachedFile.derived now has a real consumer, read-file.tsx's metadata_only encoding label reads it instead of independently re-checking the extension, so the flag and the content it describes can't disagree.
Moved isDerivedContentPath/DERIVED_CONTENT_EXTENSIONS into a new source/utils/derived-content.ts (imports only node:path). file-cache.ts and path-validators.ts both depend on it now instead of the validator reaching into the cache.
Also merged in main to pick up the schema-freshness and sandbox-jail checks the branch predated. Ready for another look.

will-lamerton:
Great job as always :) @addyCooks

addyCooks:
Thanks @will-lamerton : )

PR #1128, opened 2026-09-02
feat(vscode): add setting to toggle token/cost counter visibility

### LINKED ISSUE #1096: [Feature] Add setting to toggle Token/Cost counter visibility (default to off)

awhite0030:
## Description

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Changeset

### Automated Tests

### Manual Testing

- [ ] Tested with Ollama
- [ ] Tested with OpenRouter

## Checklist

- [ ] Self-review completed

---

Root cause: The token and cost counter is shown unconditionally, which creates visual noise for some users.
Fix: Added a VS Code setting `nanocoder.showTokenUsage` (defaulting to false) and plumbed it into the webview payload so that the UI only renders the counter when enabled. Created a changeset.
Validation: Ran `pnpm run build`, `pnpm test:format`, `pnpm test:lint`, `pnpm test:types`, `pnpm test:knip`, and targeted tests `pnpm test:ava plugins/vscode/src/`. All passed. Validated changeset logic with `node scripts/validate-changesets.js`.
Fixes https://github.com/Nano-Collective/nanocoder/issues/1096

Fixes #1096

### DIFF
--- a/.changeset/cool-snails-jump.md
+++ b/.changeset/cool-snails-jump.md
@@ -1,0 +1,5 @@
+---
+"nanocoder-vscode": minor
+---
+
+Add setting to toggle Token/Cost counter visibility
--- a/plugins/vscode/media/chat-panel.js
+++ b/plugins/vscode/media/chat-panel.js
@@ -1905,8 +1905,8 @@
 
 	// Render a small grayed-out usage line (e.g. "Tokens: 4.2k | ~$0.01")
 	// under the finished response. Cost is omitted when unknown (local models).
-	function appendUsageIndicator(usage, cost) {
-		if (!usage) return;
+	function appendUsageIndicator(usage, cost, showTokenUsage) {
+		if (!usage || !showTokenUsage) return;
 		const total = Number.isFinite(usage.totalTokens)
 			? usage.totalTokens
 			: (Number.isFinite(usage.inputTokens) ? usage.inputTokens : 0) +
@@ -2280,7 +2280,7 @@
 			handlePlanUpdate(update);
 		} else if (update.sessionUpdate === 'prompt_response' || update.sessionUpdate === 'done') {
 			// Show token usage (and estimated cost) for the finished turn
-			appendUsageIndicator(update.usage, update.cost);
+			appendUsageIndicator(update.usage, update.cost, update.showTokenUsage);
 			// Turn is complete — restore the send button
 			setProcessing(false, update.outcome || 'completed');
 		}
--- a/plugins/vscode/package.json
+++ b/plugins/vscode/package.json
@@ -164,6 +164,11 @@
 					"default": false,
 					"description": "Automatically start Nanocoder CLI if not running"
 				},
+				"nanocoder.showTokenUsage": {
+					"type": "boolean",
+					"default": false,
+					"description": "Show token usage and estimated cost in the chat panel"
+				},
 				"nanocoder.showDiffPreview": {
 					"type": "boolean",
 					"default": true,
--- a/plugins/vscode/src/chat-webview-provider.ts
+++ b/plugins/vscode/src/chat-webview-provider.ts
@@ -243,6 +243,7 @@ export class ChatWebviewProvider
 				: response
 					? 'completed'
 					: 'failed');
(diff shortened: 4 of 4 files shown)


will-lamerton (changes requested):
Thanks for picking this up. The setting itself is right (`default: false` matches what #1096 asked for, and `workspace.getConfiguration` is the correct source here rather than `SettingsManager`, which models `agents.config.json`). One blocker though.

### Blocker: this reverts #1098

The branch predates 8a46cd3a ("restore usage footers when reopening chats"), which added a second call site for `appendUsageIndicator` in the history-replay path. Only the live-turn call site is updated, so after merging main:

```
plugins/vscode/media/chat-panel.js
2065:  function appendUsageIndicator(usage, cost, showTokenUsage) {
2429:      appendUsageIndicator(replayedUsage, replayedUsage.cost);        // 2 args, never gated
2446:      appendUsageIndicator(update.usage, update.cost, update.showTokenUsage);
```

Line 2429 passes no third argument, so `showTokenUsage` is `undefined`, the new guard bails, and reopened chats lose their token/cost line permanently even with the setting on. Merging `origin/main` into this branch (clean merge) and running the existing spec confirms it:

```
✘ [fail]: replayed response usage metadata restores the token and cost line
  source/vscode/chat-panel-turn-footer.spec.ts:102
  Value is not truthy: undefined
```

CI is green only because the branch has not been rebased.

### Suggested fix: gate once, off `syncState`

Attaching the flag to the per-turn ACP update payload is why the replay path was missed, and it has two more costs: it only arrives on `prompt_response`/`done`, so toggling the setting does nothing until the next completed turn (no `onDidChangeConfiguration` listener, though the pattern exists at `extension.ts:130` and `acp-client.ts:47`), and it mixes a UI preference into the ACP message shape.

Pushing it once over `syncState` (`chat-webview-provider.ts:112`, handled at `chat-panel.js:2162`) into a module-level variable read inside `appendUsageIndicator` fixes the replay bug and the live-toggle gap together, and leaves the signature unchanged.

### Also

- No tests. `source/vscode/chat-panel-harness` and `chat-panel-turn-footer.spec.ts` already drive this renderer and assert on the usage line, so on/off across both the live and replay paths is a few lines.
- The changeset targets `nanocoder-vscode`, which is in the `ignore` list in `.changeset/config.json`, so `changeset status` bumps nothing and this ships with no changelog entry. `validate-changesets.js` passes because the name is a real workspace package. #1098, also VS Code only, used `"@nanocollective/nanocoder": patch`.

PR #1153, opened 2026-09-02
fix(utils): bump BoundedMap access order on overwrite

### LINKED ISSUE #1143: [Bug] BoundedMap.set doesn't update access order on overwrite
## Description
When `set(k, v)` is called for an existing key, it writes a fresh entry but does NOT bump the existing entry's `accessOrder`. A frequently-overwritten key can sit at the oldest position and be incorrectly evicted.

## Environment
- **OS**: Any
- **Node version**: v22.20.0
- **Provider**: N/A
- **Model**: N/A

## Steps to Reproduce
1. Instantiate a `BoundedMap` with a small max size.
2. Continually overwrite the same key while inserting new unique keys.

## Expected Behavior
The frequently overwritten key should have its access order updated and be safe from eviction.

## Actual Behavior
It retains its old order and gets evicted prematurely.

## Additional Context

MsfPablo:
## Summary

Fixes #1143.

`BoundedMap.set(key, value)` on an **existing** key updated the value but did not change the key's position in the underlying `Map`'s iteration order. Since eviction picks the oldest key (`this.map.keys().next().value`), a frequently-overwritten key stayed at the oldest position and was evicted before newer entries — the opposite of the intended LRU behavior.

## Change

In `set()`, when the key already exists, delete it before re-inserting so `Map` moves it to the most-recent (last-evicted) position:

```ts
if (this.map.has(key)) {
    this.map.delete(key);
}
```

The existing "skip eviction when overwriting" guard (`!this.map.has(key)`) is unchanged, so overwriting still never evicts a *different* entry — it now also stops evicting the overwritten key itself.

## Test

Adds a regression test (`overwrite bumps access order so the key is not evicted`) that fails on the pre-fix code:

- `maxSize: 2`; insert `a`, `b`; overwrite `a`; insert `c`.
- Pre-fix: `a` is evicted (it kept the oldest position). ✘
- Post-fix: `b` is evicted, `a` survives. ✔

All 22 `bounded-map.spec.ts` tests pass; the new test was verified to fail without the fix.

## Checklist

- [x] Tests added / updated
- [x] Changeset added (`.changeset/fix-bounded-map-overwrite-order.md`)
- [x] `test:ava`, `test:types`, lint, and format clean on the touched files

### DIFF
--- a/.changeset/fix-bounded-map-overwrite-order.md
+++ b/.changeset/fix-bounded-map-overwrite-order.md
@@ -1,0 +1,5 @@
+---
+"@nanocollective/nanocoder": patch
+---
+
+Fix `BoundedMap.set` not bumping an existing key's access order on overwrite. A frequently-overwritten key kept its original (oldest) insertion position and could be evicted before newer entries; overwrite now moves the key to most-recent, matching the intended LRU eviction order. Thanks to @MsfPablo. Closes #1143.
--- a/source/utils/bounded-map.spec.ts
+++ b/source/utils/bounded-map.spec.ts
@@ -102,6 +102,26 @@ test('BoundedMap - updating existing key does not trigger eviction', t => {
 	t.is(map.get('b'), 2);
 });
 
+test('BoundedMap - overwrite bumps access order so the key is not evicted', t => {
+	// Regression for #1143: a frequently-overwritten key used to retain its
+	// original (oldest) insertion position and be evicted before newer keys.
+	const map = new BoundedMap<string, number>({maxSize: 2});
+
+	map.set('a', 1);
+	map.set('b', 2);
+
+	// Overwriting 'a' should move it to the most-recent position.
+	map.set('a', 11);
+
+	// Inserting 'c' evicts the oldest remaining key, which is now 'b' — not 'a'.
+	map.set('c', 3);
+
+	t.is(map.size, 2);
+	t.true(map.has('a'));
+	t.false(map.has('b'));
+	t.true(map.has('c'));
+});
+
 test('BoundedMap - TTL causes entries to expire', async t => {
 	const map = new BoundedMap<string, number>({
 		maxSize: 100,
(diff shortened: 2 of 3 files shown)


will-lamerton (changes requested):
The fix itself is correct. I ran the PR head against the scenario from the description (`maxSize: 2`, insert `a`/`b`, overwrite `a`, insert `c`) and confirmed `b` is evicted and `a` survives with the new value, and that an overwrite at capacity still evicts nothing else.

A few things to sort before merge:

- **Merge `main`.** The merge base is 2026-08-03 and all three red checks are branch staleness, not this diff: `Changeset Validation` fails on `Cannot find module scripts/validate-changesets.js` (main moved that into the shared workflow's `validate-changesets: true` input), `VS Code Extension` fails on `Command "test:types:vscode" not found` (that script is on main's `package.json`, not this branch), and the unit-test failure is `app › components › settings-tabs › renders all four category tabs`, unrelated to `bounded-map`.

- **"LRU" overstates what this does.** `get()` and `has()` don't bump order, so eviction is by *write* recency, not access recency. The changeset ("matching the intended LRU eviction order") and the code comment ("Bump access order") both read as if reads count. Please reword to write recency.

- **Update the class docstring.** `bounded-map.ts:6` still says "Maintains insertion order for predictable eviction", which is no longer true now that overwrite reorders `keys()`/`values()`/`entries()`/`forEach`. No consumer is affected (the `useAppState` token cache is order-independent, and `vscode-server`'s `pendingChanges` is keyed by `randomUUID()` so it never overwrites), but the contract should say what it now does.

Smaller nits:

- The new test asserts `has('a')` but not `get('a') === 11`, so a regression that kept the key and dropped the value would pass. Worth one more assertion.
- `if (this.map.has(key)) { this.map.delete(key); }` - `Map.delete` is already a no-op on a missing key, so the guard can go.
- Changeset has no trailing newline, and the inline "Thanks to @MsfPablo" doesn't match the other files in `.changeset/`, none of which credit contributors in the body.

PR #1154, opened 2026-09-02
fix(schedule): render */N minute as "every N minutes"

### LINKED ISSUE #1132: [Bug] formatCronHuman prints raw "*/N" for hourly expressions
## Description
The "every hour at minute X" branch renders the minute field verbatim: `formatCronHuman("*/5 * * * *")` returns "every hour at minute */5", resulting in a human label that still leaks cron syntax.

## Environment
- **OS**: Any
- **Node version**: v22.20.0
- **Provider**: N/A
- **Model**: N/A

## Steps to Reproduce
1. Use a cron expression like `*/5 * * * *`.
2. Check how it is rendered in the UI.

## Expected Behavior
It should be rendered nicely, e.g., "every 5 minutes".

## Actual Behavior
It renders as "every hour at minute */5".

## Additional Context

MsfPablo:
## Summary

`formatCronHuman` leaked raw cron syntax for `*/N` minute expressions. `*/5 * * * *` matched the hourly branch (any non-`*` minute with every other field wildcard) and rendered verbatim as `every hour at minute */5`.

Added an explicit every-N-minutes branch so `*/5 * * * *` renders as `every 5 minutes`, and `*/15 * * * *` as `every 15 minutes`.

Closes #1132.

## Test plan

- [x] New regression `formatCronHuman formats every-N-minutes without leaking cron syntax` (fails without the fix, passes with it)
- [x] `tsc --noEmit` clean
- [x] `biome check` clean
- [x] changeset added (`patch`)

### DIFF
--- a/source/schedule/cron.spec.ts
+++ b/source/schedule/cron.spec.ts
@@ -69,6 +69,14 @@ test('formatCronHuman formats hourly pattern', t => {
 	t.is(formatCronHuman('0 * * * *'), 'every hour at minute 0');
 });
 
+test('formatCronHuman formats every-N-minutes without leaking cron syntax', t => {
+	// Regression for #1132: "*/N" in the minute field used to render verbatim
+	// as "every hour at minute */5".
+	t.is(formatCronHuman('*/5 * * * *'), 'every 5 minutes');
+	t.is(formatCronHuman('*/15 * * * *'), 'every 15 minutes');
+	t.is(formatCronHuman('*/1 * * * *'), 'every 1 minutes');
+});
+
 test('formatCronHuman formats daily pattern', t => {
 	t.is(formatCronHuman('0 9 * * *'), 'daily at 9:00');
 	t.is(formatCronHuman('30 14 * * *'), 'daily at 14:30');
--- a/source/schedule/cron.ts
+++ b/source/schedule/cron.ts
@@ -46,6 +46,21 @@ export function formatCronHuman(expression: string): string {
 		return 'every minute';
 	}
 
+	// Every N minutes: minute is "*/N", all other fields wildcard. Rendered
+	// explicitly so the cron syntax does not leak into the human label (#1132).
+	if (
+		minute?.startsWith('*/') &&
+		hour === '*' &&
+		dayOfMonth === '*' &&
+		month === '*' &&
+		dayOfWeek === '*'
+	) {
+		const step = minute.slice(2);
+		if (/^[1-9]\d*$/.test(step)) {
+			return `every ${step} minutes`;
+		}
+	}
+
 	if (
 		minute !== '*' &&
 		hour === '*' &&
--- a/.changeset/fix-cron-every-n-minutes.md
+++ b/.changeset/fix-cron-every-n-minutes.md
@@ -1,0 +1,5 @@
+---
+"@nanocollective/nanocoder": patch
+---
+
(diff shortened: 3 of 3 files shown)


source/schedule/cron.ts
```diff
@@ -46,6 +46,21 @@ export function formatCronHuman(expression: string): string {
...
+	if (
+		minute?.startsWith('*/') &&
+		hour === '*' &&
+		dayOfMonth === '*' &&
+		month === '*' &&
+		dayOfWeek === '*'
+	) {
+		const step = minute.slice(2);
+		if (/^[1-9]\d*$/.test(step)) {
+			return `every ${step} minutes`;
```
will-lamerton, line 60: `*/1 * * * *` is valid cron and semantically identical to `* * * * *`, which line 46 already renders as `every minute`. This branch returns `every 1 minutes`, which is both ungrammatical and inconsistent with that. Suggest special-casing it:

```ts
if (/^[1-9]\d*$/.test(step)) {
	return step === '1' ? 'every minute' : `every ${step} minutes`;
}
```

source/schedule/cron.spec.ts
```diff
@@ -69,6 +69,14 @@ test('formatCronHuman formats hourly pattern', t => {
 	t.is(formatCronHuman('0 * * * *'), 'every hour at minute 0');
 });
 
+test('formatCronHuman formats every-N-minutes without leaking cron syntax', t => {
+	// Regression for #1132: "*/N" in the minute field used to render verbatim
+	// as "every hour at minute */5".
+	t.is(formatCronHuman('*/5 * * * *'), 'every 5 minutes');
+	t.is(formatCronHuman('*/15 * * * *'), 'every 15 minutes');
+	t.is(formatCronHuman('*/1 * * * *'), 'every 1 minutes');
```
will-lamerton, line 77: This assertion pins the ungrammatical output as intended behaviour. Should be `t.is(formatCronHuman('*/1 * * * *'), 'every minute');` once the source change above lands.

will-lamerton (changes requested):
Thanks for this, the core fix is correct and well scoped. `*/5 * * * *` and `*/15 * * * *` render properly, the new branch is correctly placed above the hourly branch, and the changeset names `@nanocollective/nanocoder` (the right package). Two things to fix before merge, plus a rebase.

**1. Rebase onto main.** This branch forks from 2026-08-03 and is 376 commits behind, so all three red checks are staleness rather than your change: `Changeset Validation` can't find `scripts/validate-changesets.js` (only on newer main), the VS Code type check is against August extension code, and the one failing unit test is `settings-tabs › renders all four category tabs`, unrelated to `source/schedule/`. CI isn't meaningful until it's rebased.

**2. `every 1 minutes` (see inline).** `*/1 * * * *` means the same thing as `* * * * *`, which this function already renders as `every minute`. The spec currently asserts the ungrammatical string, so it locks the behaviour in.

**Follow-up, not blocking:** the same syntax leak still exists in the hour field. After this patch, `0 */2 * * *` renders as `daily at */2:00` and `*/5 1 * * *` as `daily at 1:*/5`. That's pre-existing on main, not a regression here, so happy for it to be a separate issue, but `0 */2 * * *` is common enough to be worth a fast follow. Cheapest fix is a guard that returns the raw expression when any field still contains `*/`, `,` or `-`.

**Minor:** `formatCronHuman` gets unvalidated frontmatter (`validateCron` has no production callers), so `*/90 * * * *` renders as `every 90 minutes` even though croner rejects it as a step above max. Capping the step at 59 in the new regex is a one-liner if you want it.

PR #1155, opened 2026-09-02
fix(utils): classify validation errors strictly over network substring heuristics

### LINKED ISSUE #1136: [Bug] error-formatter.ts substring classification of errors
## Description
Error classification relies on simplistic `.includes()` checks. For instance, a validation error message containing the word "connection" is logged as a network error; an HTTP error whose body says "validation failed for required field" is tagged as both.

## Environment
- **OS**: Any
- **Node version**: v22.20.0
- **Provider**: N/A
- **Model**: N/A

## Steps to Reproduce
1. Throw an error with the message "validation failed for connection".
2. Check the structured logs.

## Expected Behavior
It should be classified strictly as a validation error.

## Actual Behavior
It is misclassified as a network error (or both).

## Additional Context

MsfPablo:
## Summary

`createErrorInfo` misclassified validation errors as network errors when the message happened to contain a network-sounding substring like `connection` or `fetch` (#1136). The `isNetworkError` / `isTimeoutError` helpers match on loose `.includes()` substrings, so a message such as `"validation failed for connection"` was flagged as **both** a validation and a network error.

Validation now takes strict precedence over the network/timeout substring heuristics: when an error is a validation error, `isNetworkError` and `isTimeoutError` are forced to `false`. Real network errors (identified by error `code` such as `ECONNREFUSED` / `ETIMEDOUT`, or messages without a validation cue) are unaffected.

## Behavior

| message | before | after |
| --- | --- | --- |
| `validation failed for connection` | validation + network | validation only |
| `failed to fetch: connection refused` | network | network |
| `request failed` + `code: ETIMEDOUT` | network + timeout | network + timeout |

## Testing

- Added two regression specs covering the overlapping-message case and a real network error without a validation cue.
- `pnpm run test:types` clean; `error-formatter.spec.ts` 28/28 in isolation; biome clean.
- Existing `createErrorInfo` network/timeout specs continue to pass.

Closes #1136.

### DIFF
--- a/.changeset/fix-error-formatter-strict-validation.md
+++ b/.changeset/fix-error-formatter-strict-validation.md
@@ -1,0 +1,5 @@
+---
+"@nanocollective/nanocoder": patch
+---
+
+Fix `createErrorInfo` misclassifying validation errors as network errors. A validation message containing the word "connection" (or "fetch") was also flagged `isNetworkError` because the network check used loose `.includes()` substring matching. Validation now takes strict precedence over the network/timeout substring heuristics. Thanks to @MsfPablo. Closes #1136.
--- a/source/utils/error-formatter.spec.ts
+++ b/source/utils/error-formatter.spec.ts
@@ -195,3 +195,22 @@ test('createErrorInfo - recognizes ETIMEDOUT as a network and timeout error', t 
 	t.true(result.isTimeoutError);
 	t.is(result.code, 'ETIMEDOUT');
 });
+
+// Regression for #1136: a validation message containing "connection" must be
+// classified strictly as a validation error, not also as a network error.
+test('createErrorInfo - classifies validation-overlapping messages strictly as validation', t => {
+	const error = new Error('validation failed for connection');
+	const result = createErrorInfo(error);
+
+	t.true(result.isValidationError);
+	t.false(result.isNetworkError);
+	t.false(result.isTimeoutError);
+});
+
+test('createErrorInfo - still classifies a real network error with no validation cue', t => {
+	const error = new Error('failed to fetch: connection refused');
+	const result = createErrorInfo(error);
+
+	t.true(result.isNetworkError);
+	t.false(result.isValidationError);
+});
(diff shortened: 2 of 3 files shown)


will-lamerton (changes requested):
Thanks for picking this up. The direction is right, but the current form inverts the bug rather than fixing it.

**Validation is just as loose as the network check.** `isValidationError` matches `.includes('validation' | 'invalid' | 'required')`, so giving it blanket precedence lets a message substring override the hard `error.code` signal:

- `Error('connect ECONNREFUSED 127.0.0.1:11434 - invalid endpoint')` with `code: 'ECONNREFUSED'` becomes `isNetworkError: false`
- `Error('request timed out, api key required')` with `code: 'ETIMEDOUT'` becomes `isNetworkError: false, isTimeoutError: false`

That contradicts the PR description's claim that errors identified by `code` are unaffected, and the first case is realistic for local provider failures, which is where #1136 came from.

Suggested shape: let validation outrank only the message-substring branch, keeping `error.name` and `error.code` authoritative.

```ts
const validationError = isValidationError(error);
// hard signals (name + errno code) always win; validation only outranks the
// loose message-substring heuristics
isNetworkError:
    hasNetworkSignal(error) || (!validationError && matchesNetworkMessage(error)),
isTimeoutError:
    hasTimeoutSignal(error) || (!validationError && matchesTimeoutMessage(error)),
```

Please add a regression test for a code-bearing network error whose message contains a validation cue, e.g. `ECONNREFUSED` plus "invalid", asserting `isNetworkError` stays true.

**Minor:** all three heuristics are case-sensitive, so `Error('Invalid response body while trying to fetch ...')` (a real undici message) skips the validation check entirely. Out of scope here, but worth a follow-up.

**CI:** the three red checks are stale-base artifacts, not caused by this change. CI last ran on 2026-09-02 against `afecf904`; the `settings-tabs` failure was fixed by the `useTitleShape` bugfix, and `scripts/validate-changesets.js` did not exist yet at this head. A rebase onto current `main` should clear all three.

PR #1158, opened 2026-09-02
fix: give stdout and stderr their own output byte budget

### LINKED ISSUE #1140: [Bug] bash-executor.ts shared outputBytes budget between stdout and stderr
## Description
`outputBytes` is incremented by both stdout and stderr handlers against a single `BASH_MAX_OUTPUT_BYTES` cap. A chatty stderr can exhaust the budget, silently dropping stdout data with no truncation marker on stdout. The LLM sees the marker on stderr but receives an unexpectedly truncated stdout.

## Environment
- **OS**: Any
- **Node version**: v22.20.0
- **Provider**: N/A
- **Model**: N/A

## Steps to Reproduce
1. Execute a bash script that floods `stderr` up to the byte limit, followed by crucial data on `stdout`.
2. Inspect the returned streams.

## Expected Behavior
Both streams should have their own limits, or the truncation marker should clearly appear on the stream that got truncated.

## Actual Behavior
Stdout is silently truncated without a marker.

## Additional Context

kishore280:
Fixes #1140.

stdout and stderr shared one byte counter and one truncation flag. A noisy stderr could exhaust the shared budget first, silently dropping stdout data after that point with no marker at all - the caller had no way to know anything was cut.

Split into two independent pairs, one per stream. Each stream's truncation now depends only on its own volume, and gets its own marker when it happens.

Ran: test:ava on bash-executor.spec.ts (two new tests, one per direction), test:types, test:lint, test:format, test:changesets.

### DIFF
--- a/.changeset/quiet-streams-diverge.md
+++ b/.changeset/quiet-streams-diverge.md
@@ -1,0 +1,5 @@
+---
+"@nanocollective/nanocoder": patch
+---
+
+Fixed the bash tool silently dropping stdout when stderr was noisy; each stream now gets its own output budget and truncation marker. Closes #1140.
--- a/.changeset/quiet-streams-diverge.md
+++ b/.changeset/quiet-streams-diverge.md
@@ -2,4 +2,4 @@
 "@nanocollective/nanocoder": patch
 ---
 
-Fixed the bash tool silently dropping stdout when stderr was noisy; each stream now gets its own output budget and truncation marker. Closes #1140.
+Fixed the bash tool silently dropping stdout when stderr was noisy; each stream now gets its own output budget and truncation marker. Worst case memory goes from 5MB to 10MB since both streams can now hit the cap at once. Closes #1140.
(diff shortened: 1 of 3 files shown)


will-lamerton (changes requested):
The fix itself is good. Per-stream collectors are the right shape, and I confirmed the new tests are genuine regression tests: reverting `bash-executor.ts` to the parent commit while keeping the new spec fails both tests consistently, and they pass reliably against the fix. Full spec file is 35/35, types and lint clean.

Two things before this can go in:

1. **Needs a rebase.** This conflicts with `main` in `source/services/bash-executor.ts`. `main` has since added the bash-sandbox imports and the handlers have moved to ~lines 171-204. Nothing was fixed in parallel, so the conflict is purely mechanical.

2. **Worst-case memory doubles**, 5 MB to 10 MB, now that each stream carries its own `BASH_MAX_OUTPUT_BYTES`. That is the direct consequence of what #1140 asked for and I think it is the right call, but the changeset does not mention the ceiling moved. Either add a line to the changeset, or halve the constant if 5 MB was intended as a total budget.

Happy to approve once it is rebased.

will-lamerton:
Great work @kishore280 :)

Issue #1105, opened 2026-08-31
ACP: thread sub-agent tool results back so every call is announced and settles with its real status


will-lamerton:
Follow-up to #1019 / #1080, split out as agreed in review.

## What works now

#1080 routed sub-agent tool calls that need approval to `session/request_permission`. Those calls are announced as a `tool_call` before the request, prefixed `subagent:`, and titled with the sub-agent name.

## What is still missing

**Sub-agent tools that need no approval are never announced at all.** `signalToolApproval` in `source/subagents/subagent-executor.ts` is only reached for calls that need approval; everything else runs without touching the ACP layer. So a client gets cards for the approval-gated subset of a sub-agent's work and no visibility into the rest. A read-heavy sub-agent can run a dozen tools and show nothing.

**An approved call is marked `completed` at approval time, not when it runs.** The handler in `source/acp/acp-conversation.ts` settles the card on approval because the sub-agent layer never reports results back. That was a deliberate trade against leaving the card spinning forever, but it means a client sees `completed` for a tool that may then fail, and never sees real output.

Both have the same root cause: `SubagentExecutor` does not thread its tool results out to the ACP conversation. Fixing that closes both at once - announce every sub-agent tool call, then update it with the real terminal status and output.

## Also unresolved: mode and `alwaysAllow` are ignored inside sub-agents

`needsApprovalForTool` is passed only `{mode: this.currentMode()}`, and the ACP executor is constructed in `source/plain/initialize.ts` with the default `parentMode: 'normal'` and no mode resolver. A `yolo` or `auto-accept` session is therefore prompted inside a sub-agent for a tool it would not be prompted for at top level. Pre-existing, but #1080 is what made it visible. Noted in that changeset.

## Also unresolved: overlapping sessions can cross-answer approvals

The approval slot in `source/utils/tool-approval-queue.ts` is a process-wide singleton with last-writer-wins semantics. `runAcpConversation` scopes its handler to the turn with a LIFO-correct disposer, which stops the handler outliving the turn, but `turnActive` is per session (`source/acp/acp-session.ts`), so two sessions can be mid-turn at once. For that overlap the later installer answers the earlier session's approvals against the wrong session id and abort controller, which the client rejects as invalid params.

Closing it needs the slot keyed by session id, or an approval channel threaded through `SubagentExecutor` - the same threading the two items above want, so it is probably one piece of work.

addyCooks:
Hey @will-lamerton
 
I'd like to take this one
[...]

PR #1159, opened 2026-09-02
fix(app): queue concurrent approval requests instead of overwriting them

### LINKED ISSUE #1156: [Bug] Parallel subagents hang the turn, a second approval request overwrites the first
## Description

All three "ask the user" queues store exactly one resolver in a `useRef`. A second
concurrent `signal()` overwrites it before the first has been answered, and that
first promise is orphaned permanently nothing can ever resolve it.

This is reachable through ordinary use: `tool-executor` starts up to
`MAX_CONCURRENT_AGENTS` (5) subagents at once, and each can call a tool that needs
approval. Because the batch is awaited with `Promise.allSettled`, one orphaned
promise means the batch never settles and the turn never ends.

## Environment

- **OS**: all (no platform-specific code involved)
- **Node version**: v22.22.3
- **Provider**: any
- **Model**: any

## Steps to Reproduce

1. In `normal` mode, have the model emit two or more `agent` tool calls in one turn.
2. Have both subagents call a tool that needs approval (`write_file`, `execute_bash`).
3. Answer the single prompt that appears.

## Expected Behavior

Each approval request is queued and presented in turn, and every subagent promise
settles approved, denied, or cancelled.

## Actual Behavior

Only the most recent request is answerable. Earlier ones never settle, and the turn
hangs forever. Escape does not help: the stranded subagent is parked inside an
`await` that is not abort-aware. Recovery requires killing the process.

```
agent-A promise settled          : false
agent-B promise settled          : true
Promise.all([A,B]) outcome       : TIMEOUT, a promise never settled
```

(Driven against the real `useGlobalHandlerQueues` hook through `ink-testing-library`,
firing two `signalToolApproval` calls before answering. AVA additionally reported
`Failed to exit` for the run the orphaned promise kept the worker alive.)

## Logs/Screenshots

- `source/app/hooks/useGlobalHandlerQueues.tsx:41,65,92`, one resolver ref per queue
- `source/app/hooks/useGlobalHandlerQueues.tsx:46,74,101`, `xResolverRef.current = resolve`,
  assigned unconditionally, so a second caller overwrites the first
- `source/app/hooks/useGlobalHandlerQueues.tsx:75`, `setPendingSubagentApproval(approval)`
  also overwrites the visible prompt, so the user never sees the request they strand
- `source/subagents/subagent-executor.ts:65`, `MAX_CONCURRENT_AGENTS = 5`
- `source/subagents/subagent-executor.ts:739`, `await signalToolApproval(...)`, no `AbortSignal`
- `source/hooks/chat-handler/conversation/tool-executor.tsx:360`, `await Promise.allSettled(...)`
- `source/utils/global-handler-slot.ts`, the slot holds one handler; the single-resolver
  problem is in the hook above it, not here

## Additional Context

The same single-slot design backs `ask_user` (`question-queue`) and the main agent's
tool confirmation (`tool-confirm-queue`), so the defect is not specific to subagent
approvals  subagents are just the path that reaches it routinely.

Suggested fix: replace the single resolver ref in each of the three slots with a FIFO
of `{id, input, resolve}`. Render the head of the queue, resolve by id, and show a
"1 of N" indicator. Make `signal()` accept an `AbortSignal` and reject on abort so
cancelling a turn drains the queue instead of stranding it.

Related but distinct: #1019 covered the *ACP* failure of the same module (no handler
installed, so the fallback denied silently) and is closed. This is the *TUI* failure 
a handler exists, but concurrent callers overwrite each other's resolvers. The
in-progress sub-agent tool bridge work also does not cover this: that channel is
registered per session on the ACP path, and the TUI still goes through
`signalToolApproval` and the single-resolver hook.

addyCooks:
Closes #1156

## Description

The three "ask the user" slots `question-queue`, `tool-approval-queue` and
`tool-confirm-queue` each stored a single resolver in a `useRef`. A second
caller arriving before the first was answered overwrote it, and that first
promise could then never settle.

`tool-executor` starts up to `MAX_CONCURRENT_AGENTS` (5) subagents in one turn
and awaits them with `Promise.allSettled`, so one stranded caller meant the batch
never resolved and the turn never ended. Escape did not help either: the subagent
was parked in an await that is not abort-aware, so recovery meant killing the
process.

Each slot now has its own FIFO. Requests queue in arrival order, the head is
presented, and answering it resolves that caller and promotes the next, so
every caller settles exactly once with its own answer.

The three blocks were near-identical, so they collapse into one `useHandlerQueue`
helper used three times, shared by all three slots — though each slot still owns
its own queue instance, so a subagent approval and a main-agent confirmation can
be pending at once without colliding. The hook goes from 122 to 132 lines
(+79/-69): a net increase, not a reduction, but each of the three call sites
shrinks to a couple of lines. The hook's public shape is unchanged, so `App.tsx`,
`chat-input.tsx` and `interactive-app.tsx` needed no edits for the queuing itself.

### Fixed during review

`chat-input.tsx` rendered `ToolConfirmation` with no `key`, so when the queue
advanced from one request to the next, React reconciled the same component
instance and carried over its internal state (`formatterPreview`,
`hasValidationError`, ...) from the previous request onto the new one — most
visibly, a validation error on request A would auto-confirm and then get stuck
showing A's error text over B's prompt. Both `ToolConfirmation` renders in
`chat-input.tsx` are now keyed off `toolCall.id` so each queued request gets a
fresh instance.

## Scope

Two items from the issue's suggested fix are deliberately not here:

- **A "1 of N" indicator.** It changes the hook's public shape and pulls in
  `chat-input.tsx` and `interactive-app.tsx`. Worth doing, but it is UX on top of
  the fix rather than part of it. Note for whoever picks this up:
  `interactive-app.tsx` sets `cancellable` to `false` whenever a confirmation is
  pending, so with several requests queued, Escape has to be pressed once per
  request with no on-screen sign that anything is queued — even a bare "1 of N"
  would carry most of the value.
- **An abort-aware `signal()`.** That changes `createGlobalHandlerSlot`'s
  signature and every call site, including `subagent-executor.ts:739` where the
  sub-agent tool bridge work for #1105 is also editing. Queuing alone removes the
  hang; abort handling is a clean follow-up once that lands. Right now, if a turn
  is cancelled with items queued, those entries stay queued and keep being
  presented, so the user can be asked to answer prompts for a turn that's already
  dead — tracked in #1230.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Changeset

### Automated Tests

Added coverage for the gaps called out in review: `pendingSubagentApproval`
(the real `useState`, not a spy) advancing from one request to the next,
answering into an empty queue and then receiving a new request (the
`length === 1` present-on-arrival gate), and the subagent-approval and
main-agent-confirmation slots both being pending at once and settling
independently.

### Manual Testing

- [ ] Tested with Ollama
- [ ] Tested with OpenRouter

## Checklist

- [x] Self-review completed


will-lamerton (changes requested):
The FIFO approach is right, and the extraction into one `useHandlerQueue` is a genuine simplification. Installing the handler once (deps are now just the module-level setter) is a nice side benefit. Verified on the PR head: 10/10 specs pass, `tsc` clean, biome clean, and no drift in the touched files versus current `main`.

One thing to fix before merge, plus a few follow-ups.

### Blocking: queued prompts reuse a stale `ToolConfirmation` instance

`chat-input.tsx:211-225` renders `ToolConfirmation` with no `key`. When the queue advances, `pendingSubagentApproval` goes straight from A to B with no intervening `null`, so React reconciles the same element type at the same position and keeps the component's state. `ToolConfirmation` holds `formatterPreview`, `hasFormatterError` and `hasValidationError` (`tool-confirmation.tsx:36-44`), none of which reset when the `toolCall` prop changes.

Concrete failure: request A trips the validation path, so `hasValidationError` becomes true and the effect at `tool-confirmation.tsx:141-151` auto-fires `onConfirm(true)`. That resolves A and promotes B into the same instance. `hasValidationError` is still true, and that effect's deps are `[hasFormatterError, hasValidationError, onConfirm]`, none of which change (`onConfirm` is a `useCallback([])` from the new helper), so it does not re-run. B renders A's error text, the approval prompt stays hidden behind `!hasFormatterError && !hasValidationError` at line 187, and there is no select input. B can then only be settled by Escape. The formatter-error path fails the same way.

Even on the happy path, A's `formatterPreview` stays on screen until B's async preview resolves, while the prompt line already reads B's tool name, so a fast Enter approves B while A's diff is on screen. That is an approval-safety issue, not just cosmetics. Both ternary branches render `ToolConfirmation` at the same child position, so a subagent approval handing off to a main-agent confirmation carries state over too.

Fix is two lines: `key={pendingSubagentApproval.toolCall.id}` at line 212 and `key={pendingToolConfirmation.toolCall.id}` at line 219.

This hazard technically pre-dates the PR, but it was unreachable because the concurrent case hung instead. Queuing makes it reachable on exactly the scenario this PR exists to fix, so it belongs here.

### Follow-ups

- **"N remaining" indicator.** Fine to defer, but note it interacts: `interactive-app.tsx:174-183` sets `cancellable` to false whenever a confirmation is pending, so Escape routes into the prompt rather than turn cancellation. With five subagents queued the user presses Escape five times with no on-screen sign that anything is queued. Even a bare "1 of 5" would carry most of the value.
- **Abort drain.** Also fine to defer, but worth naming the new behaviour: if a turn is cancelled with items queued, those entries stay queued and keep being presented, so the user answers prompts for a turn that is already dead. The old code self-healed by clobbering. Please file the follow-up issue and link it from the PR body.
- **Test gaps.** Nothing asserts that `pendingSubagentApproval` (real `useState`, not a spy) actually advances from A to B, which is the transition the issue above hangs off. Also untested: answering into an empty queue and then receiving a new request (the `length === 1` gate), and both approval slots live at once, which is the stated reason they are separate slots.

### Nits on the description

"Each slot now backs onto a FIFO shared by all three" reads as one shared queue. Each slot has its own queue and only the helper is shared, which is what the code comment correctly says. "The change is a net reduction in the file" is not quite right either: the hook goes from 122 to 132 lines (+79/-69).

addyCooks:
Thanks @will-lamerton, everything from your review is addressed, pushed in 5ef42d5:

- Keyed both `ToolConfirmation` renders in `chat-input.tsx` on `toolCall.id`, with a comment on why the keys are load-bearing, so each queued request gets a fresh instance instead of inheriting the previous request's error or preview.
- Added the three missing tests: `pendingSubagentApproval` advancing from A to B through the real `useState`, a new request arriving after the queue fully drains (the `length === 1` gate), and the subagent-approval and main-agent-confirmation slots pending at once and settling independently.
- Corrected the description: each slot owns its own queue (only the helper is shared), and the hook is +79/-69, a net increase.
- Filed #1230 for abort-aware draining and linked it from the description.

will-lamerton (approved):
Approving. Every item from my previous review is addressed, and I resolved the merge conflict myself in c952ce54.

### Blocking item: fixed

The unkeyed `ToolConfirmation` is fixed in 5ef42d54 with the two keys, plus a comment recording why they are load-bearing. That closes the validation-error path where request A's auto-`onConfirm(true)` promoted B into A's live instance with `hasValidationError` still true, leaving B with A's error text, no select input, and only Escape to settle it.

### Test gaps: covered

All three gaps I named now have tests, and they assert the right things: `pendingSubagentApproval` through the real `useState` advancing A to B, the `length === 1` present-on-arrival gate after a full drain, and both slots pending at once settling independently. 13/13 pass.

### Description nits: corrected

Each slot owning its own queue instance, and +79/-69 as a net increase rather than a reduction. Follow-up #1230 filed and linked for the abort drain.

### Conflict resolution

`main` moved under this branch in 71d79a3e, which wrapped both `ToolConfirmation` renders in `<Box paddingLeft={footerPadding}>` for the fullscreen footer layout, colliding with the lines this PR keyed. Resolved by keeping both: the `Box` wrapper from `main`, with the `key` on the `ToolConfirmation` inside it. Keying the inner child still forces the remount, since React remounts a single keyed child when its key changes.

Verified on the merge result: `tsc --noEmit` clean, `biome check` clean on the touched files, 13/13 in `useGlobalHandlerQueues.spec.tsx` and 17/17 in `chat-input.spec.tsx`. `git diff origin/main..HEAD` is exactly the four intended files at +235/-69, so nothing from `main` was dropped in the resolution.

### One thing I also checked

`question-queue` is FIFO-backed now too, so question A to B advancement with no intervening `null` became reachable there as well, and `QuestionPrompt` is unkeyed. It is already safe: it resets `answeredRef`, `selectedIndex`, `isFreeformMode` and `freeformValue` itself through the `previousQuestion` derived-state check. No change needed.

### Follow-up worth a note, not blocking

The answer handlers lost idempotency. The old code nulled the resolver, so a second call for an already-answered request was a no-op; now `queueRef.current.shift()?.resolve(result)` consumes the next queued request and settles it with the previous answer without ever presenting it. That would be a silent wrong approval rather than a hang.

It is not reachable today: one UI call site per slot, and both double-fire windows in `ToolConfirmation` (the auto-confirm effect at `tool-confirmation.tsx:141-151`, and Escape via `useInput`) are closed by the keying, because the instance unmounts as the queue advances. Worth an `answeredRef`-style guard in `useHandlerQueue`, or at least a comment noting the guard now lives in the keys, so a later refactor that drops them does not quietly reintroduce it.

### Merging with a red required check

`pr-checks / Format Checks` is failing, and it is not this PR. `biome check .` flags import ordering in `source/tools/file-ops/write-file.tsx:18-19`, a file this PR does not touch. I confirmed it reproduces on `origin/main` (4938854e) with no PR code at all; it arrived via the `fix/document-edit-overwrite` merge, and `main` runs only the release workflows, so it landed unnoticed and now reddens every PR branch that merges current `main`. Every other check, including Unit Tests & Coverage and CodeQL, is green. Merging with a bypass on that basis and fixing `main`'s formatting separately.

addyCooks:
Thanks @will-lamerton, 
and thanks for resolving the conflict with the footer Box yourself.

Good catch on the answer handlers. 
I'll open a small follow-up that makes a second answer for an already-answered request do nothing, 
so it can never approve the next request in the queue, with a test for it. 
That way the safety doesn't rely on the keys alone.

PR #1160, opened 2026-09-02
feat(hooks): scope a hook to the file it acted on with matchPaths

### LINKED ISSUE #1009: [Feature] Auto-Format on Save Hook

AniketR10:
## Description

runs a configured formatter on files after write_file or string_replace edits them. config is under nanocoder.autoFormat in agents.config.json. off by default. failures just show a chat message and don't block anything. closes #1009

## Type of Change

- [x] New feature

## Changeset

### Automated Tests

### Manual Testing

- [ ] Tested with Ollama
- [ ] Tested with OpenRouter

## Checklist

- [x] Self-review completed

### DIFF
--- a/.changeset/hook-match-paths.md
+++ b/.changeset/hook-match-paths.md
@@ -1,0 +1,7 @@
+---
+"@nanocollective/nanocoder": minor
+---
+
+Add `matchPaths` to lifecycle hooks, so a hook can be scoped to the file a tool acted on and not just the tool name. This is what lets one formatter per language be declared directly — `{"matchTools": ["write_file", "string_replace"], "matchPaths": ["**/*.{ts,tsx}"], "command": "npx prettier --write \"$NANOCODER_FILE\""}` — instead of dispatching on the extension inside the command with a `case` statement, which was the only option before and is not portable to Windows. It applies to any file-scoped hook, not just formatting: "only lint `src/**`", "audit-log writes under `infra/`". Patterns use the same glob dialect as skill subscriptions (`**`, `*`, `?`, `{a,b}`), and the file is whichever of `path`, `file_path` or `filePath` the tool was called with.
+
+Two deliberate behaviours: a hook scoped by path does not fire for a tool that touched no file (`execute_bash`), since `matchPaths` asks a question about a file and excluding is the right answer rather than waving it through — the opposite of an omitted `matchTools`, which widens to every tool; and an absolute path is also matched relative to the project root, so a root-anchored pattern like `src/**` fires whether the model wrote `src/a.ts` or the absolute form for the same edit.
--- a/docs/features/hooks.md
+++ b/docs/features/hooks.md
@@ -66,10 +66,37 @@ Every hook is one object with:
 |-------|----------|---------|
 | `command` | yes | Shell command to run. Runs through `sh -c` (`cmd.exe` on Windows). |
 | `matchTools` | no | Tool names this hook applies to. Omitted means every tool. Ignored by non-tool events. |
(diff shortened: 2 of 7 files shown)


AniketR10:
@will-lamerton @akramcodez @Avtrkrb please review

will-lamerton (changes requested):
Thanks for this, and sorry for the timing - the Copilot points are all properly addressed, but something landed on main underneath you.

**This now overlaps `post-tool-use` lifecycle hooks**, which merged today in #1047. The documented example in `docs/features/hooks.md` is this exact use case:

```json
"post-tool-use": [
  {"matchTools": ["write_file", "string_replace"], "command": "biome check --write \"$NANOCODER_FILE\""}
]
```

Not on you at all - this PR opened 2026-09-02 and #1009 predates hooks entirely. But merging as-is gives us two config surfaces for "run a formatter after an edit", and there are a few places where `auto-format.ts` diverges from the hardening in `lifecycle-hooks.ts`:

- **The SIGKILL escalation is dead code.** Node sets `killed = true` when a signal is *sent*, not when the process dies, so `if (!child.killed) child.kill('SIGKILL')` after the SIGTERM never fires. A formatter that ignores SIGTERM is never force-killed.
- **No process-group kill.** The timeout signals the `sh -c` child only. `lifecycle-hooks.ts` spawns `detached` and signals the group (`process.kill(-pid)`), with `taskkill /T` on Windows, because killing the shell alone orphans the grandchildren of a compound command.
- **Model-controlled path interpolated into the command string.** `lifecycle-hooks.ts` holds an explicit invariant that only `hook.command` reaches the shell and everything model-influenced goes via env. This takes the opposite approach and leans on `quoteForShell`. POSIX is sound; the cmd.exe branch doesn't neutralize `%`, so a path like `a%CD%.ts` expands before the formatter sees it. Semgrep missed it because it's `spawn(shell, ['-c', cmd])` rather than `shell: true`.
- **`stdio: 'ignore'`** reduces a failure to `exited with code 1` with no stderr; hooks capture and show the output.
- **TUI only.** `runAutoFormat` is reachable only via `processAssistantResponse`, whereas `post-tool-use` fires in the TUI, `nanocoder run`, `--plain`, ACP, and subagents.

Also worth noting: formatting happens after `turnResults` is built, so the model is never told the file changed under it, and a follow-up `string_replace` can miss.

I don't want you polishing code that may not land, so the decision first: my preference is to close this and #1009 as delivered by hooks, plus a docs change making the formatter recipe prominent. If we want to keep the declarative extension-to-formatter mapping (fair - it's more discoverable than knowing `$NANOCODER_FILE`), then `autoFormat` should desugar into a `post-tool-use` hook so there's one execution path and the kill/quoting/output handling comes for free.

Happy to take either route - let me know which you'd prefer and I'll help get it there.

AniketR10:
changed this to add matchPaths to hooks instead of a separate autoFormat config. same per-language formatter setup, but it uses the existing hook path instead of adding a second one

AniketR10:
@will-lamerton please review

akramcodez:
Thanks for the PR @AniketR10

PR #1164, opened 2026-09-02
fix(ui) prevent duplicate user message

### LINKED ISSUE #1111: [Bug] user prompt submits twice in new-ui redesign
## Description

When submitting a message in the TUI, the prompt is duplicated and sent twice upon the first submission. *Note: This issue only exists in our `feature/new-ui` branch because we are actively redesigning Nanocoder's UI and handling inputs more responsively in this branch. Please ensure any fixes are targeted strictly at `feature/new-ui` and not main.*

## Environment

- **OS**: Ubuntu (WSL)
- **Node version**: v22.20.0
- **Nanocoder version**: (latest `feature/new-ui` branch)
- **Provider**: Any
- **Model**: Any

## Steps to Reproduce

1. Switch to the `feature/new-ui` branch and start the application (`pnpm run start`).
2. Type a message into the chat input box (e.g., "hi").
3. Press `Enter` to submit.

## Expected Behavior

The prompt should appear once in the chat queue, and the application should send a single API request to the LLM.

## Actual Behavior

The user's input bubble is printed twice into the chat history, and the system attempts to process the prompt concurrently due to rapid duplicate keystroke events bypassing the React `isBusy` lock.

## Logs/Screenshots

Image

## Additional Context

Forzun:
## Fix

Fix the user prompt from being submitted twice in the new UI.

https://github.com/user-attachments/assets/25f2e7c0-f694-4896-ae95-0cd9ab7caf0b

## Testing

- Reproduced the duplicate submission on `feature/new-ui`
- Implemented the fix
- Ran the project's pre-commit checks successfully

Fixes #1111

### DIFF
--- a/source/ai-sdk-client/chat/chat-handler.ts
+++ b/source/ai-sdk-client/chat/chat-handler.ts
@@ -250,6 +250,7 @@ export async function handleChat(
 							e = new Error(String(error));
 						}
 					}
+					console.log('streaming response', e);
 					streamingErrors.push(e);
 					logger.warn('Streaming error received', {
 						error: e.message,
--- a/source/ai-sdk-client/chat/chat-handler.ts
+++ b/source/ai-sdk-client/chat/chat-handler.ts
@@ -250,7 +250,6 @@ export async function handleChat(
 							e = new Error(String(error));
 						}
 					}
-					console.log('streaming response', e);
 					streamingErrors.push(e);
 					logger.warn('Streaming error received', {
 						error: e.message,
--- a/source/app/App.tsx
+++ b/source/app/App.tsx
@@ -83,6 +83,7 @@ export default function App({
 
 	// Use extracted hooks
 	const appState = useAppState(initialDevelopmentMode);
+
 	const userMessageQueue = useUserMessageQueue();
 	const queuedUserSubmitRef = React.useRef<
 		| ((
--- a/source/components/chat-queue.spec.tsx
+++ b/source/components/chat-queue.spec.tsx
@@ -59,15 +59,17 @@ test('ChatQueue merges static and queued components', t => {
 });
 
 test('ChatQueue can render the last queued component outside Static', t => {
-	const {lastFrame, rerender} = render(
+	const { lastFrame, rerender } = render(
 		<ChatQueue
 			queuedComponents={[
 				<Box key="frozen">Frozen queued message</Box>,
 				<Box key="live">Recallable queued message</Box>,
 			]}
 			renderLastQueuedComponentLive
+			disableStatic
 		/>,
 	);
+
 
 	t.regex(lastFrame() ?? '', /Frozen queued message/);
 	t.regex(lastFrame() ?? '', /Recallable queued message/);
@@ -76,6 +78,7 @@ test('ChatQueue can render the last queued component outside Static', t => {
 		<ChatQueue
 			queuedComponents={[<Box key="frozen">Frozen queued message</Box>]}
 			renderLastQueuedComponentLive
+			disableStatic
 		/>,
 	);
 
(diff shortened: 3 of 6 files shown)


will-lamerton (changes requested):
Thanks for digging into this one. A few things need addressing before this can go in.

**Debug logging left in**

- `source/ai-sdk-client/chat/chat-handler.ts:253` - `console.log('streaming response', e)`
- `source/hooks/chat-handler/useChatHandler.tsx:363` - `console.log('displayMessage', displayMessage, message)`, which dumps the full user prompt on every submit

Ink patches `console` into the transcript, so both print into the chat UI. Biome doesn't flag them, which is probably why the pre-commit run stayed green.

**Existing test fails**

I ran `source/components/chat-queue.spec.tsx` against both refs in clean worktrees:

- `feature/new-ui`: 18/18 pass
- this branch: 17/18, `ChatQueue can render the last queued component outside Static` fails at `chat-queue.spec.tsx:83`

**The fix disables the recall path in the default mode**

`!renderLastQueuedComponentLive || !disableStatic` means live rendering now only happens when `disableStatic` is true, i.e. fullscreen. Fullscreen is opt-in (`--alt-screen` or `alternateScreen: true`, see `cli.tsx:588`), so in the default inline mode `renderLastQueuedComponentLive` becomes a dead prop.

Its only producer is `recallableSubmittedDraft` (`source/app/sections/interactive-app.tsx:300`). It exists so that between submit and the first streamed token the user bubble stays outside `<Static>` and Escape can pull it back into the input (`handleRecallSubmittedDraft` slices it off `chatComponents`). With this change inline mode commits the bubble to `<Static>` immediately, so Escape-recall drops it from state but the bubble is already written to the terminal and stays on screen. That looks like trading the duplicate for a stale orphaned bubble.

It also only touches the inline path, so if the duplicate reproduces under `--alt-screen` it is untouched there.

**Scope**

#1111 claims both a duplicated bubble and a genuine double submission ("attempts to process the prompt concurrently ... bypassing the React `isBusy` lock"). This change is purely rendering-layer, so nothing here would stop a second request reaching the API. Did you see two requests go out, or was it only the visual duplicate?

**Minor**

`App.tsx`, `useAppState.tsx` and one hunk in `useChatHandler.tsx` are whitespace-only blank-line additions. Worth stripping so the diff is just the fix.

Happy to take another look once the logging is out and the recall interaction is sorted.

will-lamerton:
Hi @Forzun - thank you for this, and apologies it sat open as long as it did.

Your diagnosis was correct, and you got to it first: this PR was opened on 2 September, two days before #1182. Gating the live-queue split on `disableStatic` is the right call and is functionally the same fix we're going with in #1182 - `disableStatic` is `fullscreen || isFreshInline`, so `!renderLastQueuedComponentLive || !disableStatic` keeps the last queued component out of the live region in exactly the inline case where Ink would otherwise reprint it through `<Static>`. You found the real mechanism.

I'm closing this in favour of #1182 for one reason: removing the live slot in inline mode also breaks the thing that depends on it. `renderLastQueuedComponentLive` exists to keep an in-flight prompt out of `<Static>` during the Escape-to-recall window. Once the gate is in place, the bubble is committed to scrollback and can't be un-printed - but `handleRecallSubmittedDraft` in `interactive-app.tsx` still pops it from `chatComponents` and restores the draft into the input. The result is the prompt in the transcript *and* back in the input box, which is a worse duplicate than the one being fixed, in the default mode. #1182 carries the matching `interactive-app.tsx` change plus regression tests for both halves, which is what tipped it.

That's a consequence two layers away from the line you changed, and not an obvious one - it isn't a knock on the diagnosis.

If you'd like to pick something else up, I'm happy to point you at an issue. Thanks again for digging into this one.

Issue #1162, opened 2026-09-02
[Feature] MCP resources, prompts, and sampling support


addyCooks:
## Description

The MCP client implements exactly two operations `listTools` and `callTool`. There is no `listResources`, `readResource`, `listPrompts`, `getPrompt`, `createMessage` (sampling), `elicit`, or `listRoots` anywhere in `source/mcp/`.

## Use Case

`docs/battlemap.md` claims MCP client parity with Claude Code, but Claude Code additionally surfaces MCP resources as `@`-mentions and MCP prompts as slash commands. This is parity work on a claim already being made publicly, and it slots into infrastructure that already exists in this codebase.

## Proposed Solution

Stage it in three phases:
1. **Resources** (highest value, lowest risk) slot into the existing mention system the same way file mentions work today.
2. **Prompts** slot into the lazy command registry (`source/commands/lazy-registry.ts`) the same way custom commands from `.nanocoder/commands/` already do.
3. **Sampling**, then elicitation and roots last, since those require the client to act as a server back toward the MCP host more design work, lower immediate payoff.

## Alternatives Considered

Ship all four at once rejected in favor of staging so resources (the highest-value, most self-contained piece) land without blocking on the more involved sampling/elicitation design.

## Additional Context

- [x] I have searched existing issues to ensure this is not a duplicate
- [x] This feature aligns with the project's goals (local-first AI assistance)

## Implementation Notes (optional)

Resources phase touches the mention system and `source/mcp/mcp-client.ts`; prompts phase touches the command registry; keep sampling/elicitation/roots out of scope for the initial PR to keep it reviewable.

addyCooks:
Hi @akramcodez, 
could you assign this one to me ?

PR #1166, opened 2026-09-03
feat(skills): install skill bundles from a git repository

### LINKED ISSUE #1163: [Feature] Install and share skills (nanocoder skills add)
## Description

`docs/features/skills.md` calls a skill bundle "one shareable artifact," but there is no install path anywhere in `source/skills/` no fetch, no clone, no registry. Every skill today must be hand-authored directly into `.nanocoder/skills/`.

## Use Case

Writing a skill (a markdown bundle) is far cheaper than writing TypeScript, making distribution the cheapest available lever on contributor growth for a 51-contributor project. Right now that leverage is unused there's no way to share a working skill except copy-pasting files.

## Proposed Solution

Add `source/skills/install.ts`: fetch a tarball or shallow-clone a skill bundle into `.nanocoder/skills/`, validate it through the existing `manifest-parser.ts` and `check.ts` linter, then show a trust prompt that renders exactly which tools and event subscriptions the bundle declares before writing anything to disk. Start with a plain `skills.json` index hosted in a git repo no hosted registry infrastructure needed for v1.

## Alternatives Considered

A hosted marketplace/registry service  rejected for v1 as unnecessary infrastructure; a git-hosted index is enough to bootstrap sharing.

## Additional Context

**Read this before closing as a duplicate of #92:** #92 ("Plugin System & Marketplace") is closed as completed, so this can look like a duplicate on sight. It isn't skills delivered the *plugin* half of #92, but three of its stated goals never shipped: marketplace distribution, sandboxed execution, and lifecycle hooks. This issue is the distribution piece; lifecycle hooks are #1032 (already filed, PR #1047 open); sandboxed execution is the companion issue filed alongside this one.

addyCooks:
Closes #1163.

## Description

Adds `nanocoder skills add`, the install path for skill bundles described above. Seven files: two new (`source/skills/install.ts`, `source/skills/install.spec.ts`), three small edits to existing skills/CLI code, plus docs and a changeset.

The trust prompt as it actually renders:

```
Skill "pr-reviewer" v0.1.0 by nano-collective
  Reviews pull requests and posts findings.

  from: https://github.com/Nano-Collective/nanocoder-skills
  into: /repo/.nanocoder/skills/pr-reviewer

  Commands: /pr-reviewer:review
  Agent:    reviewer
  Tools (visible only to this skill's agent):
    - gh_pr_diff · shell script · approval: never (runs WITHOUT asking) · read-only
  Event subscriptions (the daemon fires these unattended):
    - file.changed → agent:reviewer · paths src/**

Installing a skill means running its code. Only install skills you trust.
Install this skill? [y/N]
```

`--yes` still prints that summary before installing, so an unattended install leaves a record of what it agreed to.

## Fixed during review

- **Unrelated commit dropped.** The branch had picked up an unrelated MCP resources/prompts/sampling commit (already shipping separately in #1172) that was the sole cause of all four red checks. Dropped it and resynced the branch with `main` (also picks up a newer required check, `generate:schema`, that this branch predated).
- **Bundle-root symlink escape (security).** The symlink guard only ever inspected entries found while walking *inside* the bundle root, never the root path itself. `findBundleRoot` reaches a candidate via `pathExists` (follows symlinks) and its containment check is lexical (`path.resolve`, no `realpath`), so a repo whose bundle directory is itself a symlink — reachable through an index entry's implicit name lookup, no `--subdir` needed — passed straight through: the trust prompt rendered "No tools and no event subscriptions" and files from outside the clone landed in `.nanocoder/skills/`. `assertBundleRootContained` now `realpath`s the resolved bundle root and asserts containment in the clone directory before anything reads from or copies it. Also: `buildTrustSummary`'s parse-failure fallback moved from `approval: 'always'` to `'never'` (worst-case, not the registry's own default — under-reporting risk in a trust prompt is worse than over-reporting it), and `checkSkillBundle`/`applyPromotion` are now wrapped so an unexpected throw cleans up the temp clone instead of leaking it as an unhandled rejection.
- **Two documented behaviors now actually work.** A local directory that isn't a git repository is copied instead of handed to `git clone` (which only ever accepted a git repository — `nanocoder skills add ./local-checkout` failed for any bundle not yet committed to git). `--ref` now accepts a commit SHA by falling back to `init` + `fetch` + `checkout` when the cheap `clone --branch` path rejects it (branches/tags only) — verified against a real GitHub repo, not just the local fixtures.
- **Default index doesn't exist yet.** `Nano-Collective/nanocoder-skills` 404s, so the docs' `nanocoder skills add pr-reviewer` headline example failed for every user. Publishing that repo is an org-level content decision, not something to do from this PR, so instead: reworded the install examples to lead with forms that work today (`owner/repo`, a git URL, a local path) and added a note that the default index isn't published yet. Filed #1232 to track publishing it, with a note to revert the doc ordering once it's live.
- **CLI argument parsing was silently lenient** on a security-sensitive command (does this run unattended or not?). An unrecognised flag (`--forse`) was accepted as a no-op; a value flag at the end of the args, or followed by another flag (`--ref --global`), swallowed the next flag as its value instead of erroring. Both now fail loudly. `SKILLS_CLI_USAGE` now documents `--index`, matching `--help`.
- **Non-TTY without `--yes`** now exits 1 instead of 0 — a script that forgets `--yes` gets a clear failure instead of a silent, misleading no-op success.

## Type of Change

- [x] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Documentation update

## Changeset

### Automated Tests

Added coverage for every gap named in review: the bundle-root symlink escape (via a directory junction/symlink on the local-install path, so it doesn't depend on git's mode-120000 checkout support), `buildTrustSummary`'s worst-case approval fallback, installing from a local non-git directory, `--ref` rejected against a local non-git directory, `--ref` accepting a commit SHA (checked out content verified, not just "didn't error"), an unknown flag, `--ref` with a missing value (end-of-args and followed-by-another-flag), and a non-interactive run without `--yes`. 33 tests in `install.spec.ts`, up from 23.

### Manual Testing

- [ ] Tested with Ollama
- [ ] Tested with OpenRouter

Manually ran the built CLI (`dist/cli.js`): local non-git directory install, unknown-flag rejection, `--ref` value-swallow rejection, and `--ref <sha>` against a real GitHub repo (`octocat/Hello-World` at a commit that isn't a branch tip).

## Checklist

- [x] Self-review completed


will-lamerton (changes requested):
The skills installer itself is well-built: good adversarial test coverage, sensible reuse of `checkSkillBundle` and `applyPromotion`, and a genuinely informative trust prompt. A few things to fix before merge.

## 1. Unrelated commit in the branch (blocker)

`2899e648` ("feat: Add MCP resources, prompts, and sampling support", resolves #1162) doesn't belong here, that work is already in #1172. It's also the sole cause of all four red checks:

- `source/mcp/mcp-client.ts:167` sampling handler signature doesn't match the SDK's `sampling/createMessage` request type
- `source/mcp/mcp-prompt-commands.ts:96` expected 2 arguments, got 1
- `source/utils/placeholders.ts:6` new `resource` `PlaceholderType` added to `source/types/hooks.ts` without updating the `Record<PlaceholderType, string>` map
- knip: `getMCPPromptCommandsList` unused export

Dropping it turns CI green. `6fbb117f` alone type-checks clean and its 23 tests pass.

## 2. The symlink guard doesn't cover the bundle root

`scanBundleTree` walks entries *inside* `root.path` but never checks `root.path` itself. `findBundleRoot` reaches that path via `pathExists`, which follows symlinks, and `isContained` uses `path.resolve`, which is lexical and doesn't realpath. So a repo whose `skills/<name>` is a symlink to an absolute path on the victim's disk passes the check.

Reproduced on the plainest path, `nanocoder skills add <name>` resolved through an index, no flags. Repo contained `skills/probe-skill` as a git mode-120000 entry pointing outside the clone; the trust prompt rendered normally ("No tools and no event subscriptions") and files from outside the clone landed in `.nanocoder/skills/probe-skill/`. Same via `--subdir`.

Fix: realpath the resolved bundle root and assert containment in the clone dir before `scanBundleTree` runs.

Two adjacent notes:
- `buildTrustSummary`'s catch falls back to `approval: 'always'`, which under-reports risk if the declared value was `never`. A trust prompt's fallback should be worst-case.
- `checkSkillBundle` and `applyPromotion` aren't wrapped in try/catch, so an unexpected throw leaks the temp clone and surfaces as an unhandled rejection out of `cli.tsx`.

## 3. Two documented behaviors don't work

Both confirmed by running them:

- `nanocoder skills add ./local-checkout`: docs list "a path on disk", but `parseInstallTarget` hands the path to `git clone`, so a plain bundle directory fails with `fatal: repository '...' does not exist`. Either clone-or-copy, or reword the docs to "a local git repository".
- `--ref <sha>`: docs say "branch, tag, or commit", but `cloneRepo` passes it to `git clone --branch`, which rejects SHAs (`fatal: Remote branch d9e3fe4... not found in upstream origin`). This one matters, since pinning to a commit is the safe way to install code you don't fully trust. Needs a clone-then-fetch/checkout fallback.

## 4. The default index doesn't exist

`DEFAULT_INDEX_URL` points at `Nano-Collective/nanocoder-skills`. That repo is a 404, as is the raw `skills.json`. `nanocoder skills add pr-reviewer`, the headline example in both the docs and `--help`, fails for every user until it's published.

## 5. Smaller things

- Non-TTY without `--yes` returns exit code 0 with "Aborted. Nothing was written." A CI script that forgets the flag reports success and installs nothing. Should be exit 1.
- `parseAddArgs` silently swallows unknown flags (`--forse`, `--wat` are accepted without complaint). On a security-sensitive command, a typo'd `--force` quietly not taking effect is the wrong failure mode.
- `--ref` at end-of-args consumes nothing; `--ref --global` swallows `--global` as the ref value.
- `SKILLS_CLI_USAGE` omits `--index`, though `--help` documents it.

addyCooks:
Hi @will-lamerton, 

Thanks for the thorough review! 
All addressed and pushed:

1. Dropped the unrelated MCP commit and resynced with `main`, CI is green.
2. Bundle-root symlink escape fixed: `assertBundleRootContained` realpaths the resolved root and checks containment before anything reads/copies it. Also moved the `buildTrustSummary` fallback to worst-case (`never`) and wrapped `checkSkillBundle`/`applyPromotion` so a throw can't leak the temp clone.
3. Local non-git directories now get copied instead of failing on `git clone`; `--ref` now accepts a commit SHA via an init+fetch+checkout fallback (verified against a real GitHub repo).
4. Default index doesn't exist yet, 
reworded the docs to lead with working examples and filed #1232 to track publishing it (didn't want to unilaterally create a repo under the org).
5. CLI arg parsing tightened: unknown flags and missing/swallowed flag values now error instead of being silently accepted, and non-TTY without `--yes` exits 1.

Added 10 new tests (23 --> 33) covering all of the above, all green in CI.

will-lamerton:
Great work as always @addyCooks  :)

addyCooks:
Thankyou @will-lamerton !!

PR #1168, opened 2026-09-03
fix(ui): remove vertical border characters from chat and input areas (fixes #1113)

### LINKED ISSUE #1113: [Bug] terminal selection copies border characters (┃) in chat and input areas
## Description

When selecting and copying text directly from the terminal window, the vertical border box-drawing characters (e.g., `┃`) that are used to style the input area and the AI response bubbles are copied along with the actual text content. 

## Environment

- **Nanocoder version**: (latest `main` branch)
- **Provider**: Any
- **Model**: Any

## Steps to Reproduce

1. Start Nanocoder in the terminal.
2. Submit a prompt or receive a long response from the AI that spans multiple lines and includes a left border.
3. Highlight/select the text spanning multiple lines using your mouse.
4. Copy and paste the text into an external editor.

## Expected Behavior

The copied text should only contain the actual content of the user prompt or the AI response (e.g., the raw markdown or text), omitting any visual UI elements like the left border (`┃`) or padding spaces.

## Actual Behavior

The visual border characters and trailing spaces are included in the copied text, forcing the user to manually clean up the copied string. For example:
```
┃ them via PATCH /users/:id/status with the API key, then asserts recovery returns the
┃ identical 201 shape.
```
Instead of:
```
them via PATCH /users/:id/status with the API key, then asserts recovery returns the
identical 201 shape.
```

## Logs/Screenshots

Image

## Additional Context

rishu685:
### Summary

Fixes #1113

When selecting and copying text across multiple lines directly from the terminal, the vertical border box-drawing characters (`┃`) and padding were captured into the clipboard along with the content.

This PR removes the `borderStyle="bold"` / `borderLeft={true}` box wrapper and padding from `AssistantMessageBox`, `StreamingMessage`, `UserMessage`, and `UserInput`, matching the existing borderless code block styling introduced in `842bea05`.

### Changes

- **Assistant & Streaming Messages**:
  - Removed `borderStyle="bold"`, `borderLeft={true}`, and `padding={1}` from `AssistantMessageBox`.
  - Updated `textWidth` from `boxWidth - 3` to `boxWidth` so text takes full advantage of terminal width without premature wrapping.
  - Streaming tail messages now render through the clean borderless box.
- **User Message**:
  - Removed `borderLeft` and padding box styling, updating `textWidth` to `boxWidth`.
- **User Input**:
  - Removed the `borderLeft` box container around the prompt row, updating `inputWrapWidth` to `boxWidth`.
- **Changeset**:
  - Added patch changeset (`.changeset/fix-terminal-selection-border-characters.md`).
- **Tests**:
  - Updated `assistant-message.spec.tsx`, `streaming-message.spec.tsx`, and `user-message.spec.tsx` to assert clean terminal output without any `┃` border characters.

### Verification

- `pnpm run test:ava source/components/assistant-message.spec.tsx` (90/90 passed)
- `pnpm run test:ava source/components/streaming-message.spec.tsx` (15/15 passed)
- `pnpm run test:ava source/components/user-message.spec.tsx` (24/24 passed)
- `pnpm run test:ava source/components/user-input.spec.tsx` (51/51 passed)
- `pnpm run test:types` (passed)
- `pnpm run format:check && pnpm run test:lint` (passed)
- `pnpm run build` (passed)


will-lamerton (changes requested):
The fix itself is correct. I rendered the components on `main` vs this branch and both halves of #1113 are gone: the `┃` glyphs and the padding lines. The trailing-space complaint in the issue came from `backgroundColor={colors.base}` forcing Ink to paint padding out to the box width; with the background removed Ink emits no trailing whitespace at all. Width arithmetic checks out too (`boxWidth` is already `columns - 4`, so `textWidth = boxWidth` leaves slack rather than running to the terminal edge), and `inputWrapWidth` still matches TextInput's usable width, so Up/Down line-nav vs history is unaffected.

Three things before I can approve.

**1. Copilot's whitespace-trimming comments are still unaddressed** (`assistant-message.spec.tsx:1132`, `streaming-message.spec.tsx:199`)

```js
.map(l => l.trim())      // strips the leading spaces
.filter(Boolean);
t.false(
    contentLines.slice(1).some(l => l.startsWith('   ')),  // so this can never be true
```

This is a coverage regression, not just an unimproved test: the version on `main` stripped `┃` and used `trimEnd()` only, checking both leading and trailing 3-space runs. `trim()` -> `trimEnd()` plus filtering fully-blank lines is the whole fix, and both tests still pass with it applied.

The leading-newline comments were addressed properly, and the replacement is stronger than what was on `main`. Thanks for that.

**2. `user-input.tsx` has no regression test**

It's the component the issue names first ("the input area"), and the only one of the four changed here without a `┃` assertion. Please add the same three-liner the other three specs got.

Two existing names in that file are also now stale: `user-input.spec.tsx:1042` says "(inside the input box)" and the assertion message at :1063 says "inside the bordered input box". There's no bordered input box any more.

**3. `AssistantMessageBox` is vestigial**

It's now `<Box flexDirection="column" marginBottom={1}>` plus an optional `…`, which is identical to the inline code-block branch three lines below at `assistant-message.tsx:107` and character-identical to `StreamingMessage`'s non-interactive branch at `streaming-message.tsx:78-82`. An exported component named `...Box` that no longer draws a box will mislead the next reader. Collapse the three into one helper, or rename.

---

Non-blocking, for a follow-up issue rather than this PR: `plan-review-prompt.tsx:101-106` still draws the same `┃` border, and its own comment at lines 119-127 explains that the raw artifact path sits there specifically as a copy/paste fallback. Same for `question-prompt.tsx:130` and `settings-selector.tsx:109`.

One flag for a maintainer: dropping `padding={1}` removes a blank line above and below every message body, not just the horizontal indent, so the chat is meaningfully tighter now. Reads fine to me, but it's a design change riding along in a bug fix.

Housekeeping is good: changeset correctly names `@nanocollective/nanocoder`, `tsc --noEmit` and `biome check` are clean, and all 187 tests across the four component specs pass.

rishu685:
done!

akramcodez:
Hey @rishu685, thanks for the PR! We’re now using the alternate screen mode, so this is no longer needed. Closing this PR.

Issue #955, opened 2026-08-25
[Feature] Audit and Improve Keyboard Interaction


akramcodez:
## Description

Audit Nanocoder's keyboard interaction and ensure that important actions have predictable and consistent behavior.

Terminal applications should be fully usable from the keyboard without requiring mouse interaction.

This issue should first identify the actual keyboard interaction gaps before implementing changes.

## Use Case

Users should be able to perform common Nanocoder interactions using predictable keyboard controls.

Important interactions include:

* Enter.
* Shift+Enter.
* Escape.
* Ctrl+C.
* Ctrl+V.
* Ctrl+Shift+V.
* Arrow keys.
* Home.
* End.
* History navigation.
* Multiline editing.
* Command selection.
* Contextual suggestion navigation.

## Proposed Solution

First perform a keyboard interaction audit.

Document:

1. Current behavior.
2. Expected behavior.
3. Missing/inconsistent behavior.
4. Platform/terminal-specific problems.
5. Proposed fixes.

The audit should cover:

### Composer

* Single-line input.
* Multiline input.
* Cursor movement.
* Home/End.
* Arrow navigation.
* History navigation.

### Submission

* Enter behavior.
* Shift+Enter behavior.
* Multiline submission.

### Cancellation

* Escape.
* Ctrl+C.
* Processing-state cancellation where applicable.

### Clipboard

* Ctrl+V.
* Ctrl+Shift+V.
* Multiline paste.
* Large paste.

### Contextual UI

* Navigating `/`, `!`, and `@` suggestions.
* Selecting suggestions.
* Dismissing suggestions.

The implementation should follow the audit rather than assuming that every listed key currently needs a change.

## Alternatives Considered

* Implement behavior for every listed key without auditing the existing implementation.
* Depend on mouse interaction.
* Rely entirely on terminal-native behavior.
* Document the existing behavior without making improvements.

The preferred approach is to first identify actual gaps and then fix the relevant ones.

## Additional Context

Matt specifically requested clarification around what keyboard behavior is currently missing.

Therefore, this issue is intentionally **open-ended during the investigation phase**.

Contributors should report concrete problems rather than changing keyboard behavior unnecessarily.

## Implementation Notes

This issue may overlap with the fullscreen TUI work because fullscreen terminal ownership can change how stdin and keyboard events are handled. https://github.com/Nano-Collective/nanocoder/issues/947

[...]

PR #1170, opened 2026-09-03
fix: audit & improve keyboard interaction (Home/End/Delete/Ctrl+A/E, undo/redo, showCursor guards, TitleShape cancel)

tusharui:
Description
Keyboard-interaction audit for #955 with fixes and tests:
- TextInput cursor navigation: add forward Delete, add Home/End, and guard Home/End + Ctrl+A/Ctrl+E behind showCursor so a hidden cursor can't desync the next insert. Use cursorOffsetRef to avoid setState churn on rapid keystrokes.
- Undo/redo: wire Ctrl+Z/Ctrl+Y, preserve the caret on undo/redo instead of remounting TextInput, fix stale-closure under rapid undos and same-tick input bursts via refs, and cap the redo stack (mirroring the undo cap).
- Settings: fix title-shape preview/commit so canceling no longer silently persists, and standardize footer hints.
- Tests: rendered component tests for the new key handling, exact undo/redo round-trip and cap invariants, and a title-shape cancel regression test.
Type of Change
- Bug fix
- New feature
- Breaking change
- Documentation update
Changeset
- Added a changeset (pnpm changeset) describing this change for the changelog
Docs-only or internal chores need no changeset (or run pnpm changeset --empty to note that intentionally).
Testing
Automated Tests
- New features include passing tests in .spec.ts/tsx files
- All existing tests pass (pnpm test:all completes successfully)
- Tests cover both success and error scenarios
Manual Testing
- Tested with Ollama
- Tested with OpenRouter
- Tested with OpenAI-compatible API
- Tested MCP integration (if applicable)
Checklist
- If this was for an open issue, I was assigned to it
- Code follows project style guidelines
- Self-review completed
- Documentation updated (if needed)
- No breaking changes (or clearly documented)
- Appropriate logging added using structured logging (see CONTRIBUTING.md (../CONTRIBUTING.md#logging))

### DIFF
--- a/.changeset/backspace-raw-disambiguation.md
+++ b/.changeset/backspace-raw-disambiguation.md
@@ -1,0 +1,7 @@
+---
+"@nanocollective/nanocoder": patch
+---
+
+Fixed Backspace deleting the character after the cursor on virtually every terminal. Ink parses both the physical Backspace (`\x7f`, sent by macOS Terminal, iTerm2 and essentially all Linux terminals) and the forward Delete key (`\x1b[3~`) as `key.delete`, so Backspace was routed to a forward delete (a no-op at end of line). TextInput now disambiguates at the raw-sequence level so Backspace always deletes backward.
+
+Fixed undo/redo reading stale stacks when an edit and an undo/redo arrive in the same stdin batch. `pushToUndoStack` now keeps the undo/redo stack refs in lockstep synchronously (not just after a render), and `redo()` now caps the undo stack it pushes back onto, matching `undo()`.
--- a/patches/ink@6.8.0.patch
+++ b/patches/ink@6.8.0.patch
@@ -15,3 +15,38 @@ index 30019a073d38ed05dca08f6221af187c429a536b..5254bf400cda5f4d7fdc95c0a0f6ccd8
  import ansiEscapes from 'ansi-escapes';
  import isInCi from 'is-in-ci';
  import autoBind from 'auto-bind';
+diff --git a/build/hooks/use-input.js b/build/hooks/use-input.js
+index 52b4487e..c182b5ed 100644
+--- a/build/hooks/use-input.js
++++ b/build/hooks/use-input.js
+@@ -68,6 +68,11 @@ const useInput = (inputHandler, options = {}) => {
+                 capsLock: keypress.capsLock ?? false,
+                 numLock: keypress.numLock ?? false,
+                 eventType: keypress.eventType,
++                // Patched by nanocoder: expose the raw input sequence (e.g.
++                // '\x7f' for physical Backspace vs '\x1b[3~' for forward Delete)
++                // so handlers can disambiguate keys that Ink parses to the same
++                // name (both map to `delete`).
++                raw: keypress.raw,
+             };
+             let input;
+             if (keypress.isKittyProtocol) {
+diff --git a/build/hooks/use-input.d.ts b/build/hooks/use-input.d.ts
+index e2a9f07c..2497c9ae 100644
+--- a/build/hooks/use-input.d.ts
++++ b/build/hooks/use-input.d.ts
+@@ -96,6 +96,14 @@ export type Key = {
+     Only available with kitty keyboard protocol.
+     */
+     eventType?: 'press' | 'repeat' | 'release';
++    /**
++    Raw input sequence that was parsed for this key. Useful to disambiguate
++    keys that Ink maps to the same name — e.g. physical Backspace ('\x7f') vs
++    forward Delete ('\x1b[3~'), which both set `delete: true`.
++
++    Patched by nanocoder.
++    */
++    raw?: string;
+ };
+ type Handler = (input: string, key: Key) => void;
+ type Options = {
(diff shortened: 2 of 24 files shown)


tusharui:
hey @akramcodez have look into this

tusharui:
@will-lamerton any changes required ??

will-lamerton (changes requested):
Thanks for the thorough audit - the showCursor guards, the `cursorOffsetRef` migration and the TitleShape preview/commit split all look right, and the CodeQL regex-escaping comments are addressed. Two things need fixing before this can go in.

**1. Backspace is broken on virtually every terminal (blocker)**

`text-input.tsx` splits `key.backspace || key.delete` into backwards/forwards branches, but Ink can't distinguish the two that way. In `ink@6.8.0` (`parse-keypress.js`), `\x7f` maps to `key.delete` and `key.backspace` is only set for `\b` (0x08, Ctrl+H). The physical Backspace key sends `\x7f` on macOS Terminal, iTerm2 and essentially all Linux terminals, so Backspace now takes the forward-delete path. Forward Delete (`\x1b[3~`) also maps to `key.delete`.

Verified by rendering the real component and writing `\u007f` to stdin:

- `"abcde"`, cursor at end, Backspace: value stays `"abcde"` (forward delete at EOL is a no-op)
- `"abcde"`, cursor between `d` and `e`, Backspace: gives `"abcd"` instead of `"abce"`

Both pass on the merge base, so this is a regression. `text-input-keys.spec.tsx` misses it because it only ever writes `\u001b[3~`, never `\u007f`. The fix needs to disambiguate at the raw-sequence level, plus a `\u007f` spec case.

**2. The undo/redo ref rewrite is asymmetric**

`useInputState.ts` makes refs the synchronous source of truth in `undo()`/`redo()`, but `pushToUndoStack` only writes `currentStateRef` - it never writes `undoStackRef`/`redoStackRef`, which are resynced only by `useEffect` after a render. So any edit followed by an undo/redo in the same stdin batch reads stale stacks:

- `updateInput('a')` then `undo()` with no render between: input stays `'a'` instead of reverting to `''`, and the queued push is clobbered by `setUndoStack(nextUndoStack)`
- undo back to `'a'` (redo stack holds `'ab'`), then `updateInput('aZ')` then `redo()` same tick: input becomes `'ab'`, so a stale redo entry resurrects and discards the typed character

This is the same stale-closure class the commit set out to fix, just not closed on the push side. The current specs pass only because they `rerender` between every mutation. Also minor: `undo()` caps the redo stack at `MAX_UNDO_STACK` but `redo()` applies no cap pushing back onto the undo stack.

**Smaller**

- `key.home`/`key.end` are checked before `key.ctrl`, so Ctrl+Home and Ctrl+End collapse to plain Home/End. Fine if deliberate.
- No changeset in the diff.

FYI the 6 failures in `user-input.spec.tsx` are pre-existing on `feature/new-ui`, not yours.

tusharui:
@will-lamerton  i have made changes according to requirements

will-lamerton (changes requested):
Thanks for turning this around. Both blockers from the last round are genuinely closed. I verified by reverting `text-input.tsx` and `useInputState.ts` on your branch and confirming the 5 new tests go red, so the coverage is load-bearing rather than just passing. The ink patch applies cleanly on `pnpm install --frozen-lockfile`, the lockfile hash is right, and types/format/lint are all clean. Changeset is there and correctly names `@nanocollective/nanocoder`.

One regression of the same class is still open, plus some housekeeping.

**1. Option/Alt+Backspace regressed the same way Backspace did (blocker)**

`parse-keypress.js:431` maps **both** `\x7f` and `\x1b\x7f` to `name: 'delete'`. The new guard only matches the bare `\x7f`, so Option+Backspace on macOS (Alt+Backspace on Linux) falls through to the forward-delete branch.

Verified by rendering the real component with `"abcde"` and the cursor between `d` and `e`:

```
base (feature/new-ui):  \x1b\x7f -> "abce"   (backward delete)
PR head:                \x1b\x7f -> "abcd"   (forward delete)
```

It should be one line: `key.raw === '\x7f' || key.raw === '\x1b\x7f'` (or `key.raw?.endsWith('\x7f')`), plus a `\x1b\x7f` spec case alongside the new Backspace ones.

**2. Merge conflict**

`source/components/welcome-message.spec.tsx` conflicts with `feature/new-ui`. Trivial, both sides touched the same three lines. Keep the base tagline assertions and your stronger regex escaping.

**3. The branch is stale**

`chat-input.spec.tsx` fails on the PR head but passes on base. That one is not yours, it is a missing base fix (71d79a3e). Merging `feature/new-ui` in resolves it. With base merged and the conflict fixed locally, the full components/hooks/app sweep leaves exactly 2 failures (`security-disclaimer` and the `app-container` narrow boot summary) and both fail identically on base, so nothing else regresses.

Also worth noting that no CI has run on this branch at all, so none of this has been validated by the pipeline yet.

**Non-blocking**

The `raw === '\x7f'` comparison also misses the kitty keyboard protocol, where Backspace arrives as `\x1b[127u` (`kittyCodepointNames[127] = 'delete'`). Ink's protocol is opt-in and nanocoder never sets `kittyKeyboard`, so this is dormant today. A short comment at the guard would stop a future opt-in from silently re-breaking Backspace.

Issue #862, opened 2026-08-12
[Feature] Agentic CI/CD Gate Phase 3: CI Watch Mode (`verify --watch`)


akramcodez:
## Description
When CI pipelines fail, developers must manually open the logs, trace the failure, and then prompt the agent for a fix. A persistent background watcher is needed to automate the investigation of CI failures the moment they happen.

## Expected Behavior
* Implement a `ci.job.failed` event kind in `source/events/types.ts` that is provider-agnostic.
* Create a `CiEventSource` that intelligently polls GitHub for check run failures, using exponential backoff to respect API rate limits.
* Wire this event source into the existing daemon (`source/daemon/daemon.ts`).
* Trigger a built-in `verify-ci-investigator.md` subagent to diagnose the failing logs when an event fires.
* Provide OS-level notifications (`source/utils/notifications.ts`) to alert the developer that a failure was detected and an investigation is complete.

## Actual Behavior
The daemon and event router exist for file watching, but there is no integration for CI events. CI failures require manual detection and investigation by the user.

## Goal
Transform Nanocoder into a persistent, background participant that automatically diagnoses CI failures and prepares a root-cause summary without user intervention. *(Note: All PRs for this issue should be opened against the `feature/agentic-cicd-gate` branch.)*

Piyushrathoree:
@akramcodez assign me.

akramcodez:
Thanks for your interest, @Piyushrathoree! @rohanshrma222 is already working on Phase 2. Once that’s completed, I’ll assign this to you.

rohanshrma222:
Working on phase 3

Issue #863, opened 2026-08-12
[Feature] Agentic CI/CD Gate Phase 4: Auto-Fix and Full-Commit


akramcodez:
## Description
Currently, even if the agent investigates a failure in the background, applying a fix requires user intervention. To fully automate the CI/CD loop, the agent needs the ability to push proposed fixes based on strict trust levels.

## Expected Behavior
* Wire the `--trust auto-fix` flag to authorize the `verify-ci-investigator` subagent to use write tools.
* When `auto-fix` is enabled, the agent should automatically branch off the failing PR, commit the fix, and open a new draft PR for human review.
* Wire the `--trust full-commit` flag to allow direct pushes to the branch, accompanied by an explicit, one-time YOLO-style security warning.
* Add persistent configuration options in `agents.config.json` for verify settings (trust levels, polling intervals, etc.).

## Actual Behavior
Trust levels for background jobs are not fully implemented. Any automated run lacks the capability or authorization to securely apply and push a fix.

## Goal
Complete the CI/CD loop by allowing the agent to not only diagnose but also autonomously propose and push fixes, governed by strict, user-configurable trust boundaries. *(Note: All PRs for this issue should be opened against the `feature/agentic-cicd-gate` branch.)*

rohanshrma222:
Working of phase 4 now.

PR #1175, opened 2026-09-04
feat(daemon): add CI watch mode with automated failure investigation

rohanshrma222:
Description

Adds background CI Watch Mode to the daemon (issue #862, Phase 3 of the Agentic CI/CD Gate roadmap). The daemon can now poll a branch's GitHub Actions runs, detect a failure, run a read-only verify-ci-investigator subagent to diagnose it, and automatically post the diagnosis as a PR comment  all without a human invoking anything. Opt-in via ciWatch.enabled config; disabled by default.

Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

Changeset

- [x] Added a changeset (pnpm changeset) describing this change for the changelog

Testing

Automated Tests

- [ ] All existing tests pass (pnpm test:all completes successfully)

Note on the unchecked item: pnpm run test:all doesn't invoke cleanly in my local Windows shell (a ./scripts/test.sh invocation issue, not a code issue), so I ran the constituent checks individually instead: test:format, test:lint, test:types, test:knip all pass clean, and the new/changed spec files (90 tests across backoff, github-checkrun, event-router, parse-subscribe, dispatcher, format-ci-report, daemon, trust, verify/cli) all pass. The full repo-wide test:ava run shows 62 pre-existing failures on this machine; I diffed against the pre-Phase-3 base commit on the same machine and it has 70 failing (more), with identical failure signatures (Windows path-separator assertions, a known chokidar timing flake, Windows-specific directory checks) — so these predate this branch. CI only runs on ubuntu-latest, so these are very likely Windows-local-only, but I haven't watched an actual Linux CI run of this branch complete, so I'm not checking this box outright.

Manual Testing

- [ ] Tested with Ollama
- [ ] Tested with OpenRouter

Not applicable to this change (no provider-facing code touched); manual end-to-end sanity check of CI Watch Mode itself (real gh-authenticated repo, a branch with a failing run) wasn't done — gh isn't installed in this environment.

Checklist

- [x] Self-review completed
- [ ] Appropriate logging added using structured logging (see CONTRIBUTING.md#logging)

### DIFF
--- a/.changeset/ci-watch-daemon.md
+++ b/.changeset/ci-watch-daemon.md
@@ -1,0 +1,5 @@
+---
+"@nanocollective/nanocoder": minor
+---
+
+Added background CI-watch: the per-project daemon can now poll GitHub Actions for check-run failures on the current branch (`ciWatch.enabled` in preferences) and automatically run a new read-only `verify-ci-investigator` subagent to diagnose the failure, posting the diagnosis as a PR comment when one exists and firing OS notifications on both detection and completion.
--- a/.gitignore
+++ b/.gitignore
@@ -117,6 +117,7 @@ coverage
 .nanocoder/tasks.json
 .nanocoder/schedule-runs.json
 .nanocoder/schedules.json
+.nanocoder/ci-watch-state.json
 .generated-prompts/
 .nanocoder/skills
 
--- a/docs/configuration/preferences.md
+++ b/docs/configuration/preferences.md
@@ -84,6 +84,29 @@ Desktop notification preferences are stored under the `nanocoder.notifications` 
 
 You can change these via `/settings` → **Notifications**. See [Desktop Notifications](../features/notifications.md) for full details including platform-specific setup.
 
+### CI Watch Configuration
+
+Background CI-watch preferences are stored under the `nanocoder.ciWatch` namespace. When enabled, the daemon (`nanocoder daemon start`) polls GitHub Actions for failures on the current branch and automatically investigates them:
+
+| Option | Type | Default | Description |
+|--------|------|---------|-------------|
+| `nanocoder.ciWatch.enabled` | boolean | `false` | Enable background CI-watch polling. Requires the `gh` CLI to be installed and authenticated. |
+| `nanocoder.ciWatch.pollIntervalMs` | number | `30000` | Base interval between polls, in milliseconds. |
+| `nanocoder.ciWatch.maxPollIntervalMs` | number | `300000` | Backoff ceiling applied when consecutive `gh` calls fail. |
+
+There's no `/settings` wizard for this yet — set it by hand-editing the preferences file:
+
+```json
+{
+  "nanocoder": {
+    "ciWatch": {
+      "enabled": true,
+      "pollIntervalMs": 30000
+    }
+  }
+}
+```
+
 When you restart Nanocoder, it automatically restores your last provider, model, theme, shape, paste threshold, and notification preferences.
 
 ## Manual Management
--- a/docs/features/notifications.md
+++ b/docs/features/notifications.md
@@ -56,7 +56,10 @@ Notification preferences are stored in `nanocoder-preferences.json` under the `n
       "events": {
         "toolConfirmation": true,
         "questionPrompt": true,
-        "generationComplete": false
+        "generationComplete": false,
+        "triggeredRunComplete": true,
+        "ciFailureDetected": true,
+        "ciInvestigationComplete": true
       },
       "customMessages": {
         "toolConfirmation": {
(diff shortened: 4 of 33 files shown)


will-lamerton (changes requested):
Nice work on this - the exhaustiveness switch in `buildTriggeredTask`, the pure `ExponentialBackoff`, the trust/frontmatter drift guard in `trust.spec.ts`, and the deterministic header in `formatCiReport` are all the right calls. A few things need fixing before merge.

## Blocking

**1. The first CI poll blocks daemon startup, so `daemon start` reports a false failure.**

`daemon.ts` puts `await ciSource?.start()` before `await ipcServer.start()` and the lockfile write. `start()` awaits `pollOnce()` awaits `router.emit()`, and that chain is awaited all the way down: `EventRouter.emit` awaits `dispatcher.dispatch`, `BackpressureDispatcher.dispatch` returns `dispatchImmediate` for non-`file.changed` kinds, and `SkillDispatcher.dispatchAgent` awaits `executor.execute()`. So the full LLM investigation runs before the IPC server opens. `daemon/cli.ts` gives up at `waitForLockfile(projectRoot, 5000)` and prints "Daemon process spawned but did not write a lockfile within 5s" on a daemon that is actually fine. Start the CI source after the lockfile write, and schedule the first poll rather than awaiting it.

**2. Every daemon restart re-posts the same diagnosis.**

`lastSeenRunId` is in-memory and `start()` polls immediately with no recency bound. Restart the daemon (or reboot, or let launchd/systemd restart it) while the failed run is still the latest completed one and it investigates again and posts another identical PR comment. The docstring frames non-persistence as acceptable "consistent with subscriptions not being persisted", but an unpersisted subscription is invisible whereas this writes to a public PR each time. Needs a persisted last-seen id, a "completed after daemon start" filter, or a check for an existing report on the PR.

**3. Any user skill subscribed to `ci.job.failed` gets its output auto-posted to the PR.**

`defaultOnActivity` branches on `activity.event.kind === 'ci.job.failed'`, not on the subscription. Since this PR makes `ci.job.failed` a first-class `subscribe:` kind for user-authored skills, any such skill's output gets wrapped in a "Nanocoder CI Investigation" header and posted. Key on `subscription.id === 'builtin:ci-investigator'`.

**4. The poll's `gh` call has no timeout.**

`pollOnce` calls `execGh([...])` with no `timeoutMs`, and `execProcess` only installs a timer when `timeoutMs > 0`. A hung `gh run list` never resolves, so `scheduleNext` never runs and the watcher dies silently; on the startup poll it turns issue 1 into a permanent hang. `getPrNumberForBranch` in this same PR passes `TIMEOUT_GH_METADATA_MS` correctly.

## Should fix

**5. `--limit 1` spans all workflows.** The latest completed run on the branch is taken regardless of workflow, so a failure is masked as soon as any other workflow completes after it, and a scalar `lastSeenRunId` cannot distinguish "same failure still latest" from "a second workflow also failed". Also only `conclusion === 'failure'` counts, so `timed_out` and `startup_failure` are ignored.

**6. The riskiest path is untested.** `formatCiReport` has good isolated coverage, but `handleCiInvestigationComplete` (the code that posts model output to a PR) has none: not the `defaultOnActivity` routing, not `getPrNumberForBranch` to `postPrComment`, not the no-PR case, not the failure path. Making it injectable would cover this and issue 3 together.

**7. `daemon.spec.ts` third test is environment-dependent.** It asserts one thing if `gh` is installed and the opposite if not, so it passes either way and only exercises one branch. `isGhAvailable` deserves the same seam as `ciEventSourceFactory` and `fileWatcherFactory`.

**8. Skill-authored `ci.job.failed` subscriptions are silently dead when `ciWatch.enabled` is false.** The registrar registers them unconditionally but nothing emits. A boot warning when `router.listByKind('ci.job.failed')` is non-empty and the source is off would save a confusing debugging session.

## Notes

- Docs are stale: `docs/configuration/preferences.md` needs the `nanocoder.ciWatch.*` block and the two new notification events, `docs/features/notifications.md` likewise, and `docs/features/skills.md` needs `ci.job.failed` as a valid `subscribe:` kind.
- The `entry.ts` comment on why the investigator cannot post is correct (`git_pr`'s `approval` fn is mode-independent, and the daemon registers no approval handler so `signalToolApproval` defaults to denied). Worth adding one caveat: `resolveToolApproval` also short-circuits on `alwaysAllow`, unreachable today only because the executor passes just `{mode}`.
- `getCiWatchPreference` uses an inline `import('@/types/config')` annotation where the rest of `preferences.ts` uses top-level imports.

akramcodez (approved):
Crazy work, @rohanshrma222! Feel free to take on Phase 4 next: [https://github.com/Nano-Collective/nanocoder/issues/863](https://github.com/Nano-Collective/nanocoder/issues/863) 🚀

Issue #1174, opened 2026-09-04
[Bug]: prompt duplicates in scrollback during live-to-static transition in inline mode


Rudra2637:
## Description

In default inline mode (non-alt-screen), submitted user prompts duplicate in the terminal scrollback history when the AI model begins streaming its response.

Holding the submitted message in a live recallable draft causes the prompt to first render in Ink's dynamic live region. When the model starts streaming and the recall window closes, moving the message into Ink's permanent `<Static>` transcript causes Ink to reprint the prompt to standard output, leaving a duplicate copy of the user's prompt in the scrollback buffer.

## Environment

- **OS**: Windows 11 (also reproducible across macOS / Linux)
- **Node version**: v22.x
- **Nanocoder version**: 1.30.0
- **Provider**: Ollama (or any provider with latency before first stream chunk)
- **Model**: qwen2.5 (or any model)

## Steps to Reproduce

1. Start Nanocoder in default inline mode:
   ```bash
   pnpm run start
   # or
   nanocoder
   ```
2. Type a prompt (e.g., `hello can u tell me what is capital of india`) and hit **Enter**.
3. While the model is processing/generating the first token, observe the prompt rendered in the dynamic live region.
4. As soon as the first stream chunk arrives and the recall draft state transitions to static history, observe that Ink prints the prompt a second time into the terminal scrollback.

## Expected Behavior

The submitted prompt should only appear once in the terminal transcript.

## Actual Behavior

The user prompt appears twice in the terminal history:

##Screenshot

<img width="657" height="386" alt="Image" src="https://github.com/user-attachments/assets/6b44a1b1-b125-4bc8-80ac-d46bc48c3138" />

## Additional Context

- [x] I have searched existing issues to ensure this is not a duplicate
- [x] I can reproduce this issue consistently
- [x] This issue occurs with the latest version of nanocoder
- The issue does not happen in fullscreen mode (`--alt-screen`) because `<Static>` is disabled (`disableStatic: true`) and rendering is handled within a virtual viewport.
- The root cause is `renderLastQueuedComponentLive` moving the prompt component from dynamic flow to `<Static>` in inline mode. Gating `renderLastQueuedComponentLive` behind `fullscreen` mode resolves the duplicate print cleanly.
- I have prepared a fix and unit tests ready to submit in a PR.

Rudra2637:
@akramcodez would like to work on this, It can be solved with a minimal fix

akramcodez:
sure @Rudra2637

PR #1182, opened 2026-09-04
fix(ui): prevent duplicate user prompt rendering in inline mode

Rudra2637:
## Description

Fixes an issue in default inline mode (non-alt-screen) where submitted user prompts are duplicated in the terminal scrollback history when the model begins streaming its response.

Issue link: https://github.com/Nano-Collective/nanocoder/issues/1174

### Root Cause
In default inline mode, chat transcript history is backed by Ink's `<Static>` component. When a prompt was submitted, `renderLastQueuedComponentLive` placed the message in the live dynamic region (to support draft recall before streaming starts). Once the model began streaming and the recall window closed, moving the message into `<Static items={allStaticComponents}>` caused Ink to reprint the component directly into terminal stdout, duplicating the prompt.

### Fix
- Gated `renderLastQueuedComponentLive` to only activate in fullscreen (`--alt-screen`) mode where `<Static>` is disabled and rendering occurs inside a virtual scrolling viewport.
- In inline mode, submitted prompts are immediately kept in the static queue, preventing redundant stdout reprints when streaming begins.
- Added unit test in `source/app/components/chat-history.spec.tsx` verifying inline mode behavior.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Changeset

### Automated Tests

- [x] All existing tests pass (`pnpm test:ava source/app/components/chat-history.spec.tsx source/components/chat-queue.spec.tsx` - 32/32 tests passed)
- [x] TypeScript checks pass (`tsc --noEmit`)
- [x] Biome formatting/linter checks pass (`biome check .`)

### Manual Testing

- [x] Tested with Ollama in default inline mode (verified prompt appears exactly once)
- [x] Tested in fullscreen mode (`--alt-screen`) to ensure no regression in alt-screen rendering

## Checklist

- [x] Self-review completed


will-lamerton (changes requested):
Thanks for the detailed root-cause writeup - the analysis of why Ink reprints on the live-to-static transition is correct. Two blockers before this can land.

**1. This regresses Escape-to-recall in inline mode.**

`renderLastQueuedComponentLive` exists specifically to keep the in-flight prompt out of `<Static>` during the recall window. Once Ink commits it to Static it is permanently in scrollback and cannot be un-printed. Gating it off in inline mode while leaving the Escape handler ungated (`source/app/sections/interactive-app.tsx:261` still reads bare `recallableSubmittedDraft`) means Escape now pops the component from `chatComponents` and restores the draft into the input, while the already-printed bubble stays in scrollback. The user ends up with the prompt in the transcript *and* back in the input box - a worse duplicate than the one being fixed, in the default mode.

This is caught by an existing test: `Escape recalls an in-flight user message before assistant streaming starts` passes on `main` (24/24) and fails on this branch.

**2. The new test fails in CI and does not guard the regression.**

```
source/app/components/chat-history.spec.tsx:237
  t.is(matches?.length, 1)
  - undefined
  + 1
```

It passes locally and fails on CI, and that environment-dependence is the tell. The assertion reads `lastFrame()`, which returns Ink's most recent write, not accumulated scrollback - and the duplication in #1174 is a scrollback artifact, so `lastFrame()` structurally cannot observe it. What it actually measures is Ink's write batching: with the gate applied, `chatQueueProps` is identical across both renders, React.memo bails out, and no new frame is written. That is why the result flips between environments. Copilot's original concern still stands. Asserting against the accumulated frames array (or the fake stdout write log) rather than `lastFrame()` would cover the real behaviour.

**Direction**

The mechanism you identified is right, but removing the live region removes the feature that depends on it. Either gate the Escape handler too - which means recall is dropped in inline mode, and that should be a deliberate call rather than a side effect - or find a way to stop Ink reprinting without giving up the live slot.

**Minor**

- Two trailing blank lines at the end of `chat-history.spec.tsx`.
- The changeset body is a paragraph of implementation detail; changelog entries read better as what changed for the user, with the mechanism left to the commit message.

Rudra2637:
Thanks for the thorough review , both points are completely valid.

On Blocker 1 (Escape-to-recall regression): You're right. I gated renderLastQueuedComponentLive off in inline mode .
That caused a worse duplicate than the original bug. I should have caught the failing test on the branch — apologies for missing that.

On Blocker 2 (test doesn't guard the regression):  I'll rewrite the assertion against the accumulated frames array or the fake stdout write log to actually cover the scrollback artifact.

On the fix direction: I dug into Ink's render loop (ink.js:342-354) and confirmed that when <Static> receives a new item, it always does stdout.write(staticOutput) — there's no hook to suppress it. So preventing the reprint at the Ink level without forking isn't feasible.

Given that, I think the cleanest path is Option A: gate the Escape handler too — disable recall in inline mode as a deliberate trade-off rather than a side effect. Inline mode already doesn't have the alt-screen viewport, so the live slot behavior is fundamentally different there anyway.

If that trade-off is acceptable, I'll:

Gate the Escape recall handler with fullscreen && to match
Rewrite the test to assert against accumulated stdout writes instead of lastFrame()
Clean up the trailing blank lines and reword the changeset to focus on user-facing behavior
Let me know if you'd prefer a different direction before I push the update.

will-lamerton:
Thanks for digging into Ink's render loop - and no worries about the failing test, that's what CI is for.

I'd rather not take Option A. Gating Escape off in inline mode means Escape falls through to `handleCancel`, so the draft is gone and the user retypes. That trades a cosmetic duplicate for losing what someone just typed, and both land in the default mode.

There's a third option that keeps the feature. Your fix is right: in inline mode, commit the prompt to Static immediately. The issue is only what recall then does. In `handleRecallSubmittedDraft`, skip the `setChatComponents(slice(0, -1))` pop when not in fullscreen, and keep everything else - trim the trailing user message from `appState.messages`, restore `submittedDraft` into the input.

The bubble stays in scrollback, which it has to anyway since it can't be un-printed. The draft comes back. Nothing duplicates. The transcript shows a prompt that was submitted and then cancelled, which is what happened. If you want to make that explicit you could mark it as cancelled, but plain is fine to start.

That way the Escape handler stays ungated and the existing recall test should pass as-is in both modes.

Two other things when you push: you'll need a rebase (main has moved under both files), and the test assertion should read the accumulated frames array or the fake stdout write log rather than `lastFrame()`, which structurally can't observe a scrollback artifact.

Rudra2637:
Thanks , that's a much better solution
I've implemented it as suggested:
Kept the committed bubble in scrollback during inline draft recall by gating setChatComponents on altScreenActive.
Updated the test to assert against the accumulated frames buffer (stdout.frames) instead of lastFrame().
Rebased cleanly onto the latest main.
All tests and checks are passing. 
Thanks again

will-lamerton:
Thanks @Rudra2637 - the runtime fix is right, and the Escape-recall handling is exactly the direction we discussed. I've pushed three small changes on top rather than sending it back round again.

**The recall test wasn't guarding the gate it was written for.** The assertion read a regex over the accumulated `frames` log:

```
const allOutput = frames.join('\n');
t.regex(allOutput, /submitted bubble: fix the typo/);
```

`frames` keeps every past write forever, so the bubble is in there from the moment it was first printed - the regex passes whether or not the pop ran. I checked by reverting `altScreenActive &&` in `handleRecallSubmittedDraft` and re-running: all 27 tests in `interactive-app.spec.tsx` stayed green. The gate was effectively untested.

Swapped it for an assertion on the state the gate actually controls:

```
t.is(latestChatComponents.length, 1);
```

That fails (`0` vs `1`) with the gate reverted and passes with it in place. Worth internalising the general shape here, since this is the third time it's come up on this PR: when you need to assert something *stays*, assert on the state, not on rendered output - a frame log can only ever tell you something appeared at least once, never that it's still there.

The `chat-history.spec.tsx` tests, for what it's worth, do now guard their gate properly - reverting the `fullscreen &&` in `chat-history.tsx` fails both of them. That part landed well, and the stale nc-review comment above predates `01c9fe1e`.

**Dropped the stats commits.** `727897cd` and `abcdb727` changed `stats-display.tsx` and `stats.spec.tsx`, which aren't related to this bug. The functional `setRange` change is correct on its own merits and I've kept the patch - it should go in as its own PR with its own changeset, not ride along under a changeset that only describes the inline-render fix. Coverage gates are worth satisfying by covering the code you wrote, not by topping up elsewhere.

**Two minor cleanups.** Removed the redundant `fullscreen &&` at the `ChatHistory` call site, since `chat-history.tsx` already gates the same prop one level down; and rewrote a comment on the static-transition test that still described an earlier version of it (it mentioned `React.memo` bailing out, but the test now mounts two separate trees).

Also rebased onto current `main` - the branch was 49 commits behind, which is the likely source of the coverage-drift failure, since head and baseline weren't measuring the same tree.

Verified locally: 70 tests pass across `chat-history`, `interactive-app`, `chat-queue` and `stats`; `tsc --noEmit` and `biome check .` both clean. Merging once CI is green.

will-lamerton:
@akramcodez - you're driving this one, so I'm off the branch. Ignore the mechanics in my comment above: I'd pushed a rebase + fixes at `241f6309`, and the force-push to `ae329594` landed on top of it, so none of that is on the branch any more. No harm done - but one finding from it is worth carrying over, because it's the thing I'd hold the merge on.

**The inline recall test can't fail.** `interactive-app.spec.tsx:430`:

```js
const allOutput = frames.join('\n');
t.regex(allOutput, /submitted bubble: fix the typo/);
```

`frames` is the accumulated write log - it keeps every past write forever, so the bubble is in there from the moment it was first printed. The regex passes whether or not the pop ran. I checked by reverting `altScreenActive &&` in `handleRecallSubmittedDraft` and re-running: all 27 tests in the file stayed green. So the `altScreenActive` gate - one of the two production changes this PR exists to make - currently has no test holding it.

Fix is small; assert on the state the gate controls rather than on rendered output:

```diff
 	let latestIsCancelling = true;
+	let latestChatComponents: React.ReactNode[] = [];
 
 		latestMessages = messages;
+		latestChatComponents = chatComponents;
 
-	const {stdin, lastFrame, frames} = renderWithTheme(<RecallHarness />);
+	const {stdin, lastFrame} = renderWithTheme(<RecallHarness />);
 
-	const allOutput = frames.join('\n');
-	t.regex(allOutput, /submitted bubble: fix the typo/);
+	t.is(latestChatComponents.length, 1);
```

Verified both directions: fails `0` vs `1` with the gate reverted, passes with it in place. The general rule worth keeping - a frame log can tell you something *appeared*, never that it's *still there*, so anything asserting persistence has to read state.

The `chat-history.spec.tsx` tests are fine, for the record - reverting the `fullscreen &&` gate fails both of them.

**On the coverage drift.** I'd leave the `stats` and `isFreshInline` tests out of this PR. The new lines here are already covered by the two recall tests; the 0.02% is near-certainly an artifact of the base being 49 commits behind main, so head and baseline aren't measuring the same tree. A rebase is the thing to try before adding tests in unrelated files - and if it still drops after that, I'd rather take the drop and look at it separately than move the number from `stats-display`. The `setRange` change in there is a genuine fix and deserves its own PR and changeset.

Sing out if you want me to look at anything else on it.

PR #1183, opened 2026-09-04
feat(commands): add inline overrides for slash commands

### LINKED ISSUE #1151: [Feature] Allow ?config= override on slash commands for one-off tuning
## Description
Currently, all per-session overrides (e.g., `/context-max`, auto-compact thresholds) are CLI-wide once set. Allowing an inline configuration via `?config=` (e.g., `/compact ?threshold=80`) would let users test a value for one command without needing `/settings` round-trips.

## Use Case
Testing temporary configurations for single commands without affecting the global state of the session or requiring tedious settings toggles.

## Proposed Solution
Add a generic `scope='once'` flag or query parameter parsing to the slash command parser, tying into the existing `session-override.ts` which already supports scoped updates.

## Alternatives Considered
Writing full script wrappers or requiring users to toggle the setting back and forth manually.

## Additional Context

`session-override.ts` already has the plumbing to support temporary overrides. Just map the query string parameters to the command invocation context.

MayurK-cmd:
## Description

Closes #1151 

 Slash commands now accept inline `?key=value` overrides that scope a setting to the current command only. Lets users test a value for one command without round-tripping through `/settings` or otherwise mutating the session state.

  ```
  /compact ?threshold=80              # session-override key, applied via auto-compact store
  /context-max 128k ?once             # session-override key, applied via models.dev store
  /compact --mechanical ?preview      # legacy flag (forwarded as --preview)
  /some-command ?unknown=1            # ignored; preserved in args so the handler can error normally
  ```

  ### How it works

  - `parseInlineOverrides(args)` (new, pure) splits an args array into `{args, overrides}`. Only tokens starting with `?` are
  considered; bad keys (leading digit, whitespace) are kept in `args` so the command's own "unknown arg" path runs.
  - `expandOverrideArgs(overrides)` turns recognised legacy keys (`preview`, `llm`, `mechanical`, `aggressive`, `conservative`,
  `auto-on`, `auto-off`) into the corresponding `--flag value` tokens, so existing handlers see them as if the user had typed the
  long form.
  - `applyOnceOverrides(overrides)` is async and writes recognised session-override keys (`threshold`, `auto-compact`, `context-max`)
  to the existing `setAutoCompactThreshold` / `setAutoCompactEnabled` / `setSessionContextLimit` setters. It returns a `restore`
  callback that the dispatcher calls in a `finally` block, so the override only persists for the single command invocation.
  - The slash-command dispatcher in `source/app/utils/app-util.ts` is wrapped in a `try/finally` so `restore` always runs, even if a
  handler throws.

  Adding a new inline-override key is a one-line change: add a `case` in `applyOnceOverrides` and (if it isn't an existing `--flag`)
  optionally add an entry in `LEGACY_FLAG_NAMES`.

  ### Why async

  `applyOnceOverrides` defers the heavy `auto-compact` / `models-dev-client` imports via dynamic `import()` so the parser itself is import-free. Without this, the spec timed out under AVA's serial worker because importing `@/utils/auto-compact` pulls in the chat-handler / config / tokenization init graph.

  ### Files

  - `source/utils/inline-overrides.ts` (new)
  - `source/utils/inline-overrides.spec.ts` (new)
  - `source/app/utils/app-util.ts` (wrap dispatcher in try/finally, rebuild `commandParts` from `parseInlineOverrides` +
  `expandOverrideArgs`)
  - `source/app/utils/app-util.spec.ts` (4 new dispatcher tests)
  - `.changeset/inline-once-overrides.md` (minor: new feature)
  - `.gitignore` (`.pnpm-store/`)

  ### Out of scope

  The pre-existing `bash command - queues a completed BashProgress with showOutput` test in `app-util.spec.ts` still fails on this branch (it has been failing since #826 / July 2026). It is unrelated to this change and is not fixed here. Worth a separate issue.

  ## Type of Change

  - [ ] Bug fix
  - [x] New feature
  - [ ] Breaking change
  - [ ] Documentation update

  ## Changeset

  `.changeset/inline-once-overrides.md` is a `minor` changeset naming `@nanocollective/nanocoder` and explaining the new `?key=value`
  syntax. Validated locally with `node scripts/validate-changesets.js` (72 changesets checked, all resolve).

  ## Testing

  ### Automated Tests

  Verified locally:
  - `pnpm test:ava source/utils/inline-overrides.spec.ts` → **16/16 passed**
  - `pnpm test:ava source/app/utils/app-util.spec.ts` → all 4 new inline-override tests passed; one pre-existing unrelated failure
  (`bash command - queues a completed BashProgress with showOutput`, present since #826) — see "Out of scope" above
  - `pnpm test:types` → clean
  - `pnpm test:lint` → 516 files, 0 fixes needed

  The 16 parser/helper tests cover: empty input, single key=value, mixed positional + override, bare `?flag` (no `=`), `?key=value`
  with extra `=` in value, dotted/dashed/underscored keys, invalid keys preserved in args, `expandOverrideArgs` for both boolean and key=value forms and unknown keys, plus `applyOnceOverrides` smoke tests for empty input, unknown keys, unparseable `threshold` values, and unparseable `context-max` values.

  ### Manual Testing

  - [ ] Tested with Ollama
  - [ ] Tested with OpenRouter

  This change is provider-agnostic — the override is wired into the slash-command dispatcher, not into any provider's chat pipeline.
  Manual end-to-end testing against a real provider was not done as part of this PR; the change was validated via the unit/
  integration tests listed above. The session-override stores it writes to (`auto-compact`, `setSessionContextLimit`) are already
  covered by their own specs and used in production by `/compact --threshold` and `/context-max <n>`.

  ## Checklist

  - [x] Self-review completed

  No documentation update included in this PR — the override syntax is a small extension of existing `--flag` behaviour and the new docblock on `applyOnceOverrides` documents the public surface. Happy to add a docs page if reviewers want one.

  No logging added: the override is purely a session-override passthrough that uses the existing setter functions. A successful
  override is observable via the normal command result (e.g. `/compact ?threshold=80` prints the same "Auto-compact threshold set to 80%" success message as `/compact --threshold 80`). Failures (unparseable values) are silently ignored per the same behaviour the existing `--threshold` parser already follows (the value simply isn't applied).


will-lamerton (changes requested):
Nice, clean parser and the `try/finally` around the dispatcher is the right structural move. Types, lint, changeset validation and the 16 parser specs all pass locally for me. A few correctness issues need fixing before this can go in.

### Blocking

**1. `restore()` wipes a pre-existing session override instead of restoring it.**

`restorations.push(() => setAutoCompactThreshold(null))` resets to the config default, not to whatever was there before. Verified on the branch:

```
before (user ran /compact --threshold 90):  90
during (/compact ?threshold=80):            80
after restore (expected 90):                null
```

Same for `auto-compact` (`setAutoCompactEnabled(null)`) and `context-max` (`resetSessionContextLimit()`). A "temporary" override silently destroys the user's real session setting, which is the opposite of what #1151 asks for. Please read the current value first (`autoCompactSessionOverrides.threshold`, `getSessionContextLimit()`) and restore that.

**2. `?context-max=128k` sets the limit to 128 tokens.**

`applyOnceOverrides` uses `Number.parseInt(String(value), 10)`, while every other caller uses `parseContextLimit` from `source/utils/parse-context-limit.ts`, which handles the `k` suffix. Confirmed: `getSessionContextLimit()` returns `128`. Please use `parseContextLimit`.

**3. Both examples in the PR description do not do what the description says.**

`threshold` is not in `LEGACY_FLAG_NAMES`, so `expandOverrideArgs` returns `[]` (your own test asserts this). `/compact ?threshold=80` therefore runs a plain manual compaction, does not print `Auto-compact threshold set to 80%`, and the temporary threshold has no effect at all because manual `/compact` never reads it - only the auto-compact path does. And `?once` in `/context-max 128k ?once` is not a recognised key anywhere, so it is a no-op while `128k` sets the limit permanently. Either wire these up or correct the description.

**4. The override is stripped for ~11 handlers and passed through raw to every other command.**

`handleBuiltInCommand(message, options)` still receives the original `message` and re-splits it, so every lazy-registry command (`/model`, `/provider`, ...) sees the literal `?foo=1` token, and `?preview` never becomes `--preview` for them. `handleRetryCommand` also re-derives its args from raw `message`. The "handlers stay oblivious to the override feature" comment only holds for the `commandParts` handlers.

**5. No test covers the feature actually working.**

All four `applyOnceOverrides` tests are `t.notThrows(() => restore())` on empty/unknown/unparseable input. None asserts that an override is applied or correctly reverted, which is why issue 1 is invisible to the suite. The app-util additions are described as "4 new dispatcher tests" but none of them invoke `handleSlashCommand`. Please add a test that sets a prior value, applies an override, restores, and asserts the prior value is back.

### Minor

- UTF-8 BOM (`EF BB BF`) at the top of `source/utils/inline-overrides.ts` and `.changeset/inline-once-overrides.md`. Biome and `@changesets/parse` both tolerate it, but no other file in the repo has one. Please strip it.
- Validation asymmetry: `/compact --threshold 40` errors with a message, `?threshold=40` silently clamps to 50.
- `.gitignore` `.pnpm-store/` is unrelated scope creep and is one of the two merge conflicts against `main` (the other is `app-util.ts`). Please rebase.
- New imports sit at line ~1088 mid-file in `app-util.spec.ts`. Specs are biome-excluded so nothing complains, but it is against the file's convention.
- `typeof arg !== 'string'` in `parseInlineOverrides` is unreachable given the `readonly string[]` parameter type.
- The `parseInlineOverrides(['? bad'])` case cannot occur in practice, the dispatcher splits on `/\s+/` so no token can contain a space.

will-lamerton (changes requested):
Thanks for the rework. Items 1, 2, 4 (for built-ins and retry) and 5 are properly fixed, and the test coverage is genuinely solid now. I re-verified on a clean worktree at `22297efb`: `tsc` clean, `biome lint` and `biome format` clean, `ava` on `inline-overrides.spec.ts` + `app-util.spec.ts` + `auto-compact.spec.ts` gives 137 passed / 0 failed (the `bash command - queues a completed BashProgress with showOutput` failure did not reproduce for me), no conflicts against `main`. Two things still need fixing.

### Blocking

**1. `knip` fails, and it will break the release workflow.**

The split into `source/utils/auto-compact-session.ts` re-exports two symbols nothing consumes:

```
Unused exports (1)
autoCompactSession  source/utils/auto-compact.ts:26:2
Unused exported types (1)
AutoCompactSessionOverrides  type  source/utils/auto-compact.ts:24:14
```

`main` is knip-clean. `pnpm test:knip` runs in `release.yml`, not pr-checks, so this merges green and then breaks release on main. Please drop the two unused re-exports.

**2. The changeset examples are no-ops (this is my earlier item 3, still open).**

`autoCompactSessionOverrides.threshold` / `.enabled` are read in exactly two places, `maybeAutoCompact` (`source/utils/auto-compact.ts:89-112`) and `useAppHandlers.tsx:377-391`. Both are the automatic compaction path, which only runs on a chat turn, never during a slash command. Since `restoreOnce()` fires in the dispatcher's `finally`, the override is applied and reverted with nothing in between that reads it.

So both examples in `.changeset/inline-once-overrides.md` do nothing:

- `/compact ?threshold=80` - past the flag loop `handleCompactCommand` only uses `mode`, `preview` and `strategy`. It never consults the threshold.
- `/context-max 128k ?auto-compact=on` - `128k` sets the limit permanently, and `auto-compact=on` is applied then reverted unread.

The one key that genuinely works today is `?context-max` on `/usage`: `source/commands/usage.tsx:177` reads `getSessionContextLimit()` inside the awaited handler, so `/usage ?context-max=200k` really does render against 200k. That is the example the changeset should lead with. Either wire `threshold` / `auto-compact` so some command actually observes them, or narrow the changeset and the PR description to the case that works. As it stands the CHANGELOG would ship two examples that do nothing.

The PR description is also still the original text: it documents `?once` (not a recognised key anywhere) and claims `/compact ?threshold=80` prints `Auto-compact threshold set to 80%`. Please bring it in line with the code.

### Non-blocking, but worth a decision

- Custom commands still bypass the feature entirely. `handleCustomCommand(message, commandName, options)` at `source/app/utils/app-util.ts:681` runs before the override parsing with the raw message, so `.nanocoder/commands/*.md` see `?foo=1` literally. Fine as a deliberate scope call (it is also why free-text custom commands do not silently lose `?word` tokens), just say so somewhere.
- Unknown `?foo=bar` keys are now dropped entirely before the handler sees them, so `/compact ?threshhold=80` is an invisible no-op. Combined with `--threshold 40` erroring while `?threshold=40` is silently skipped, the override path has no error surface at all. Worth at least a "unknown override key" message.
- `expandOverrideArgs([{key: 'preview', value: 'yes'}])` returns `['--preview', 'yes']`, leaving a stray `yes` positional (your own test asserts this). Harmless for `/compact`, wrong shape for any handler that reads args positionally.
- `import {parseInput} from '@/command-parser'` is still out of order in the `app-util.spec.ts` import block. Cosmetic, biome excludes specs.

The BOMs, the `.gitignore` creep, the unreachable `typeof arg !== 'string'` guard and the `'? bad'` case are all resolved. Fix the two blockers above and I think this is good to go.

MayurK-cmd:
Rebased onto current main (was 245 behind) and addressed the two blockers. Conflicts in app-util.ts (kept the #1162 MCP-prompt intercept; MCP prompts and custom commands both bypass ? parsing with the raw message) and auto-compact.ts (kept the auto-compact-session.ts split, now using clampThreshold, with auto-compact.ts re-exporting so existing importers are untouched) are resolved.

1. knip: dropped the unused AutoCompactSessionOverrides type re-export — pnpm exec knip exits 0.
2. Changeset/docs: narrowed to what actually works. /usage ?context-max=200k is the documented working example; threshold/auto-compact stay as tested apply/restore plumbing but are explicitly marked reserved since no built-in reads them synchronously during dispatch. Also fixed expandOverrideArgs valued booleans (?preview=yes → --preview) and removed the .pnpm-store/.gitignore creep.

Tests: tsc/lint/format/knip clean; inline-overrides.spec 21/21, auto-compact.spec 41/41, new app-util.spec override tests 5/5. The bash … showOutput failure is the pre-existing #826 one, unrelated. Also updated the PR description — the old text still documents ?once and claims /compact ?threshold=80 prints a confirmation, which isn't what the code does.

PR #1192, opened 2026-09-04
fix(custom-tools): clear the SIGKILL escalation timer and drop a dead guard

### LINKED ISSUE #1141: [Bug] custom-tools/handler.ts double-kill timer leak
## Description
After `child.kill('SIGTERM')`, an unref'd `setTimeout(... 1000)` schedules `child.kill('SIGKILL')`. If the child exited cleanly within 1s, the timer still fires and calls kill on a dead process (the `!child.killed` guard doesn't reliably catch it).

## Environment
- **OS**: Any
- **Node version**: v22.20.0
- **Provider**: N/A
- **Model**: N/A

## Steps to Reproduce
1. Run a custom tool that terminates immediately upon receiving SIGTERM.
2. Wait 1 second and monitor process logs.

## Expected Behavior
The SIGKILL timer should be cleared when the process exits.

## Actual Behavior
The timer fires, attempting to kill a dead process, potentially causing an unhandled error or noisy warning.

## Additional Context

yashksaini-coder:
Fixes #1141

## The leak

`runScript` sends `SIGTERM` on timeout, then schedules a `SIGKILL` escalation 1s later. Nothing held a reference to that inner timer, so `clearTimeout` on `'close'`/`'error'` only ever cleared the outer one. `.unref()` kept it from holding the event loop open, but inside a long-lived process (the TUI, the daemon) the loop is alive anyway, so the callback still fired against an exited child.

## The second defect

The escalation was guarded on `!child.killed`, which never held:

```ts
child.kill('SIGTERM');
setTimeout(() => {
    if (!child.killed) child.kill('SIGKILL'); // never runs
}, 1_000).unref();
```

Node sets `subprocess.killed` as soon as a signal is *delivered*, not when the child dies — the `SIGTERM` on the line above already made it `true`. So `SIGKILL` was never sent.

That has a worse consequence than the noisy kill the issue describes: a script that traps `SIGTERM` is never force-killed, `'close'` never fires, and because the rejection lives in the `'close'` handler, **`runScript`'s promise never settles**. The custom-tool call hangs past its own timeout.

## The fix

Track the escalation timer, clear both timers together on `'error'` and `'close'`, and drop the guard. If the timer fires, the child is still alive by construction — and `kill()` on an exited process is a no-op in Node (it swallows `ESRCH`), so the guard bought nothing even in the race window.

## Test

`runScript: escalates to SIGKILL when the script ignores SIGTERM` runs `trap '' TERM; sleep 5` with a 100ms timeout and asserts it settles well inside the `sleep`. Verified failing on the unpatched handler (settles at ~5s) and passing after (~1.1s = 100ms + the 1s grace window).

```
22 tests passed   # source/custom-tools/handler.spec.ts
```

`test:format`, `test:lint`, and `test:types` are clean. The 14 unrelated failures in the full `test:ava` run (`cli-integration`, `config/paths` darwin/win32, `git-diff` oversized patch) reproduce on an unmodified `upstream/main` checkout.


yashksaini-coder:
@akramcodez the PR is ready for review

will-lamerton (changes requested):
Diagnosis is right and I reproduced it: with the old `!child.killed` guard the escalation never runs and a `trap '' TERM` script settles at ~5s; with the patch the shell dies at ~1.1s. Tracking both timers and dropping the guard are both correct.

One thing to fix before merge.

The rejection still lives in the `'close'` handler, and `'close'` waits for the stdio pipes to close, not just for the process to exit. `SIGKILL` only reaches the shell, not its descendants, so an orphaned grandchild holding the inherited stdout/stderr keeps `'close'` from firing. Same script, only difference being whether the inner command redirects its stdio:

```
trap '' TERM; sleep 5 >/dev/null 2>&1   ->  'exit' at 1103ms, settled at 1107ms
trap '' TERM; sleep 5                   ->  'exit' at 1103ms, settled at 5016ms
```

That second case is the headline bug from the description, still present after the patch, and the `>/dev/null 2>&1` in the new test is exactly what hides it. Any custom tool whose body spawns something that inherits stdout (dev server, background `&` job, wrapper script) still hangs past its timeout.

Minimum fix: reject on `'exit'` rather than `'close'` when `timedOut` is true. The captured output is discarded on the timeout path anyway, so nothing is lost by not waiting for pipe drain. Please also add a test variant without the redirect to pin it.

Optional and bigger: `spawn(..., {detached: true})` plus `process.kill(-child.pid, 'SIGKILL')` to take out the whole process group. Better in general, but needs a Windows branch.

yashksaini-coder:
Fixed in f1e7af06. You're right on all counts, and the redirect in that test was indeed hiding it.

One wrinkle worth recording, because it cost me a confused half hour: your exact repro doesn't reproduce on every shell.

```
/bin/sh -> bash
trap '' TERM; sleep 5              exit=1105ms  close=1105ms   <- no hang
trap '' TERM; sleep 5; echo done   exit=1102ms  close=5003ms   <- hangs
sleep 5 & wait                     exit= 102ms  close=5004ms   <- hangs
trap '' TERM; sleep 5 & wait       exit=1103ms  close=5004ms   <- hangs
```

Bash exec-optimises a trailing command, so `sh -c "trap '' TERM; sleep 5"` *replaces* the shell with `sleep` instead of forking it. Ignored signal dispositions survive `exec`, so `sleep` inherits the SIGTERM-ignore and there is no grandchild holding the pipes — SIGKILL lands on it directly. Appending anything (`; echo done`) defeats the optimisation and your hang appears. Presumably your `/bin/sh` is dash, which doesn't do that.

Row 3 is the case I'd flag as the real severity: `sleep 5 & wait` hangs for 5s with **no trap at all**. The shell died to plain SIGTERM at 102ms and the call still sat pending on the pipe. That needs nothing exotic — any tool body that backgrounds a job hits it.

So the new test uses `trap '' TERM; sleep 5 & wait`, which pins the behaviour regardless of whether the shell exec-optimises. The redirected test stays, now scoped to escalation alone, with a comment saying so.

Two things beyond the minimum you asked for:

- The `exit` handler destroys `child.stdout`/`child.stderr`, so our end of the pipe drops rather than staying readable for as long as the orphan lives. Same reasoning as `BashExecutor.cancel`.
- `clearTimers()` now runs on `exit` unconditionally, not just on the timeout path. That closes a narrow race the original had: a process exiting a few ms before its deadline could still have the timer fire before `close`, flipping `timedOut` and reporting a completed run as a timeout.

Left the process-group kill alone as you suggested — `detached: true` plus `process.kill(-pid)` wants a Windows branch and is a wider change than this issue. Happy to open a follow-up for it. Note the orphan does still outlive the tool call; this fix stops it blocking the caller, it doesn't reap it.

akramcodez:
Thanks for the PR @yashksaini-coder

PR #1194, opened 2026-09-04
feat: add lsp_format_document tool

### LINKED ISSUE #1171: [Feature] expose lsp_format_document as an agent-callable tool
## Description

`LSPManager` has a fully implemented and tested `formatDocument()` method but there is no agent-callable tool for it. The only LSP tool exposed to the agent today is `lsp_get_diagnostics`. The format capability is completely unreachable.

## Use Case

A common agent workflow is write a file then format it to match project style. Without this tool the agent falls back to `execute_bash` and has to guess the right formatter command per language and per project setup — it might run `prettier`, `biome`, `black`, or `gofmt` depending on context it may not have. The LSP connection is already auto-discovered and running, knows exactly which formatter applies to each file, and has the full implementation ready. There is just no way to call it.

## Proposed Solution

Add a `lsp_format_document` tool that calls `getLSPManager().formatDocument(filePath)`, applies the returned `TextEdit[]` to the file on disk, and returns a confirmation. If no language server is available for the file it should return a clear message rather than silently succeeding.

## Alternatives Considered

Relying on `execute_bash` to run formatters manually. This works but requires the agent to know the right command per language, and it bypasses the LSP connection that is already established and auto-configured.

## Additional Context

## Implementation Notes

Three files needed:

- `source/tools/lsp-format-document.tsx` — new tool, model it after `source/tools/lsp-get-diagnostics.tsx`. Call `getLSPManager().formatDocument(path)`, apply the `TextEdit[]` array to the file using `writeFile`, return a short confirmation string.
- `source/tools/index.ts` — import and register the new tool in `staticTools` next to `getDiagnosticsTool`
- `source/tools/lsp-format-document.spec.ts` — cover: no LSP available returns graceful message, LSP available applies edits and returns confirmation, invalid path returns error

The `formatDocument` method in `source/lsp/lsp-manager.ts` and the underlying `source/lsp/lsp-client.ts` implementation are already complete with their own tests. No new infrastructure needed, scope is roughly the same size as the diagnostics tool was when it was first added.

Dhirenderchoudhary:
## Description

Adds an agent-callable `lsp_format_document` tool that uses the existing `LSPManager.formatDocument()` API to format a file via the connected language server, apply the returned edits, and write the result to disk. This replaces the need to guess a formatter CLI through `execute_bash`.

Also wires approval, plan-mode exclusion, ACP tool kind, compact result display, and a short docs note.

Closes #1171

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Documentation update

## Changeset

### Automated Tests

### Manual Testing

- [ ] Tested with Ollama
- [ ] Tested with OpenRouter

## Checklist

- [x] Self-review completed


will-lamerton (changes requested):
Nice work - the integration is more thorough than most tool-addition PRs. Plan-mode exclusion is wired in code and not just docs, checkpointing works for free via `isReadOnly`, and the approval matrix is covered across all four modes. A few things to fix before merge.

**1. Hardcoded 2-space formatting will mangle tab-indented projects.** `formatFileWithLsp` calls `manager.formatDocument(absPath)` with no options, so `LSPClient.formatDocument` falls back to `tabSize: 2, insertSpaces: true` (`source/lsp/lsp-client.ts:270`). Servers that honour `FormattingOptions` (typescript-language-server does) will convert tabs to spaces. This repo is Biome-formatted with tabs, so running the tool on nanocoder's own source reformats it against project style - which contradicts the tool description's promise of "project-style formatting". Please read `.editorconfig` / the project formatter config, or at minimum expose `tabSize` / `insertSpaces` in the input schema.

**2. "No formatting changes needed" is returned when the server cannot format at all.** `LSPClient.formatDocument` returns `[]` when `documentFormattingProvider` is absent (`source/lsp/lsp-client.ts:263`), and `formatFileWithLsp` maps empty edits to `No formatting changes needed for X.` The agent then concludes the file is clean when nothing was attempted. Please distinguish "server has no formatting support" from a genuine no-op.

**3. `positionToOffset` does not clamp `character` to the end of its line.** It clamps only to `text.length`, so an end position whose character overshoots the line length swallows every following line. On `"ab\ncd\n"`, an end of `{line: 0, character: 50}` resolves to offset 6 (end of file) rather than 2 (end of line 0). Some servers emit large sentinel characters for "to end of line". Worth a test case too. Separately, the doc comment calls the result a "UTF-16-agnostic byte offset" - it is a UTF-16 code-unit offset, which is correct for LSP's default encoding, but the comment states the opposite of what the code does.

**4. Missing changeset.** New user-facing tool, so it needs one (`pnpm changeset`). Note the package name must be `@nanocollective/nanocoder`.

**5. `--plain --output-format json` will not report formatted files.** `source/plain/shell.ts:317` lists `['write_file', 'string_replace', 'diff_edit']` as the mutating tools that populate `filesChanged`. `lsp_format_document` mutates a file and takes a `path` arg but is missing, so its edits silently drop out of the report.

Minor, non-blocking: `FILE_ARG_TOOLS` in `source/acp/acp-timeline.ts:8` does not include the tool, so ACP checkpoints fall back to an `'opaque'` git-status scan instead of a targeted path.

will-lamerton:
Great work @Dhirenderchoudhary :)

PR #1196, opened 2026-09-05
feat(privacy): extend prompt scrubbing to tool results

### LINKED ISSUE #1161: [Feature] Extend prompt-scrub to tool results
## Description

Privacy scrubbing runs on the system prompt and on user/assistant messages, but tool result messages are explicitly passed through unscrubbed. The exemption is one line: `if (m.role === 'tool') return m;` in `source/ai-sdk-client/chat/chat-handler.ts` (line 177).

## Use Case

Tool results are the largest volume of text leaving the machine and the least user-reviewed a `cat .env`, a `git diff` carrying a key, a config dump, or a stack trace with a token all go to the provider verbatim. Nanocoder's headline principle is zero telemetry, and `/privacy` tells users what gets removed, while the single largest channel is exempt from it.

## Proposed Solution

Run the same detector set against tool result content, and rehydrate on the way back through the existing `rehydrate` path (already wired for tool args around chat-handler.ts:435). Make the detector set for tool results configurable `PathDetector` and `UrlDetector` are already disabled on the existing scrub paths because scrubbing paths/URLs out of tool output breaks the agent's ability to act on it, and the same tradeoff applies here.

## Alternatives Considered

Leave it as user-education only (`/privacy` documents the gap) rejected, since it contradicts the project's own zero-telemetry framing.

## Additional Context

- [x] I have searched existing issues to ensure this is not a duplicate (checked against #635, which shipped the original scrubber but explicitly skips `role === 'tool'`)

addyCooks:
## Description

Privacy scrubbing ran on the system prompt and on user/assistant message content,
but tool-role messages were passed through verbatim the largest and
least-reviewed body of text leaving the machine. A `cat .env`, a `git diff`
carrying a key, or a stack trace with a token reached the provider in the clear
even with scrubbing on.

Three channels are now covered:

- tool result `content`;
- `structuredContent`, which the converter sends as `type: json` produced today
  by `lsp_get_diagnostics` and `write_tasks`;
- the arguments of assistant `tool_calls`, which are rehydrated to their real
  values before entering history and were therefore replayed unscrubbed on every
  later turn.

Structured payloads and tool-call arguments are walked leaf by leaf rather than
scrubbed as serialised JSON, so a replacement can never span a quote boundary and
corrupt the document. That also removes the stringify / rehydrate / parse round
trip on the way back, along with its failure mode.

Tool output is truncated to `MAX_TOOL_RESULT_CHARS` before the detectors run, so
the converter cut can never land mid-placeholder and the elided middle is never
scanned at all.

Paths and URLs stay in the clear, as on the existing scrub paths the agent has
to be able to act on the files and endpoints a result names.

The scrub and rehydrate logic moves out of `handleChat` into
`ai-sdk-client/chat/privacy.ts`, next to `provider-options` and
`streaming-handler`; `chat-handler.ts` is a net 86 lines shorter.

Closes #1161.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

## Changeset

### Automated Tests

### Manual Testing

- [x] Tested with Ollama
- [x] Tested with OpenRouter
- [x] Tested MCP integration (if applicable)

## Checklist

- [x] Self-review completed
- [ ] Documentation updated (if needed) no doc changes needed; the Privacy
      settings copy makes no claim this contradicts


will-lamerton (changes requested):
Sound premise and a clean refactor. Walking string leaves instead of scrubbing serialised JSON is the right call, and catching the assistant `tool_calls` arguments (rehydrated to real values before entering history, so replayed in the clear every later turn) is a good find. I checked the branch out and confirmed the 23 specs, biome and `tsc --noEmit` are all green.

Two things need fixing before merge, plus two quick ones.

**1. The "cut can never land mid-placeholder" guarantee does not hold.**

`scrubMessage` truncates before scrubbing, but `convertToModelMessages` truncates again afterwards (`message-converter.ts:166`). Scrubbing *expands* text (`«Email_1»` is 9 chars, `a@b.co` is 6), so a result already at the 20k cap grows past it and gets re-cut by the converter, which is exactly the failure mode the description says is eliminated.

Verified on this branch: 20,000 chars of dense short emails scrub to 23,841, and the converter's second cut breaks a placeholder on both sides of the elision marker:

```
... «Email_1309» «Email_1308» «
... [Output truncated: 23841 characters total; ...] ...
04» «Email_1003» «Email_1002» ...
```

`opens=1616, whole=1615`, and the tail resumes mid-token at `04»`. Not a leak, but the model gets garbage and any echo of the broken placeholder fails rehydration. The `truncates a tool result before scrubbing it` test passes only because its filler is `'x'.repeat(30_000)`, which does not expand.

Fix by making the boundary placeholder-aware rather than relying on ordering: truncate after scrubbing with a cut that snaps outside a `«…»`, or have `truncateToolResult` refuse to split one.

**2. `structuredContent` has no pre-truncation at all.**

Same root cause, worse. The converter serialises the scrubbed payload, truncates the JSON string and downgrades to `type: 'text'` (`message-converter.ts:169-178`), so the cut always lands on scrubbed text. `lsp_get_diagnostics` on a large repo is a realistic over-cap producer.

**3. Lost error containment on tool-arg rehydration.**

The old code wrapped each tool call in try/catch, logged `Failed to rehydrate tool call` and fell back to the original. `rehydrateResponse` drops that, so a throw now reaches the outer catch at `chat-handler.ts:511` where it can be misread as an API error or trigger the `skipTools` retry. Cheap to restore per tool call.

**4. The changeset is wrong about MCP.**

It says `structuredContent` is "the channel MCP results travel on (the only channel the model sees for those)". It is not. `structuredContent` is set only when a result carries `llmContent`, and the only producers are `lsp_get_diagnostics` and `write_tasks` (`message-handler.ts:98-120`). MCP results are plain strings on `content`, covered by the content scrub. The PR description has this right; the changeset ships to the public CHANGELOG.

Non-blocking notes:

- Whole history is re-scrubbed every turn and tool output is now included, which was previously the one thing skipped and is the bulk of the text. Measured at roughly 2.3ms per 1k chars, a session holding 400k chars of tool output costs about 0.9s of synchronous CPU per request on the Ink event loop. A memo cache keyed on message content would fix it, since history is immutable. Follow-up is fine.
- On the structured path the converter ignores `content` entirely, so `scrubText(truncateToolResult(message.content))` is wasted work there.
- Per-leaf scrubbing is not a perf concern, I measured it: 2000 small-leaf scrubs take 21ms against 248ms for one equivalent 108k-char pass.
- Worth adding specs for the expansion-past-cap case and an over-cap `structuredContent`, since those are the cases the current tests miss.

addyCooks:
Hi @will-lamerton, 
Fixed all four, plus the specs. Pushed as the latest commit.

1. `truncateToolResult` now snaps cuts outside `«Category_N»` tokens instead of relying on scrub-then-truncate ordering covers both the pre-scrub truncate and the converter's post-scrub re-truncate at `message-converter.ts:166`.
2. `structuredContent` is now truncated (and downgraded to text, mirroring the converter's own fallback) *before* the leaf-scrub runs, so an over-cap payload isn't fully scrubbed just to be discarded.
3. Restored the per-tool-call try/catch on rehydration falls back to the original call and logs, same as the old code.
4. Changeset corrected: `structuredContent` comes from `lsp_get_diagnostics`/`write_tasks`, not MCP (which is plain-string `content`, already covered).

Also skipped the wasted `content` scrub on the structured path, and added both missing specs (expansion-past-cap, over-cap structuredContent). Memo cache left as a follow-up per your note.

Separately, rebased onto current `main` branch was 73 commits stale, which is why Config Schema Freshness / Linux sandbox jail were failing (unrelated to this diff). CI is 16/16 green now.

will-lamerton:
Nice work @addyCooks :)

addyCooks:
Thanks @will-lamerton !!

PR #1200, opened 2026-09-05
fix(timeline): write timeline.json atomically

### LINKED ISSUE #1130: [Bug] Non-atomic write of timeline.json
## Description
`saveIndex()` writes the timeline index non-atomically (`fs.writeFile` straight to the live path). A crash mid-write corrupts the index and breaks every timeline action card until the user manually deletes the corrupted file.

## Environment
- **OS**: Any
- **Node version**: v22.20.0
- **Provider**: N/A
- **Model**: N/A

## Steps to Reproduce
1. Take an action that updates the timeline.
2. Terminate the application during the file write.
3. Open the application again.

## Expected Behavior
Atomic writes should be used (write to a temporary file, then rename).

## Actual Behavior
The index file gets corrupted, breaking timeline action cards.

## Additional Context

puri-adityakumar:
## Description

Closes #1130.

`TimelineManager.saveIndex()` overwrote the live `timeline.json` in place, so a crash mid-save could truncate the index. `loadIndex()` then discards an unparseable index and starts empty, silently losing every checkpoint of the session.

**What changed:**
- Added async `atomicWriteFile(filePath, data, options?: {mode?: number})` to the shared `source/utils/atomic-write.ts` (with spec coverage) and switched `saveIndex()` to it: write to a temp file, rename into place. Readers only ever see the complete old or complete new index
- On a failed save, the unpersisted cache entry is dropped and the index reloaded from disk, so later reads report what was actually persisted
- Temp file is unlinked when the write or rename fails, so no `.tmp` litter

This covers process death mid-save; rename without fsync does not cover power loss (noted in a code comment).

**Not changed on purpose:** before-image files in `capture()` stay non-atomic. They land in a fresh per-entry UUID dir before the index is saved, so a torn write there only leaves an orphan that `pruneStaleSessions` cleans up later. The index is the single point that needed atomicity. Migrating the two remaining private async copies in `session-manager.ts` and `semantic-memory-manager.ts` to the shared util is left as follow-up material.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Changeset

### Automated Tests

- [x] Bug fix includes passing tests in `.spec.ts/tsx` files

Note: 4 git-tool tests (`git-diff`, `getGitStatusSummarySync`) fail on a clean `main` checkout too. They spawn temp repos and the repo's own commit-msg hook rejects their fixture commits in this environment. Unrelated to this PR.

New tests:
- `timeline-manager.spec.ts`: saves atomically over a read-only live file (the discriminating test, fails on the old implementation), keeps the previous index intact on disk and in memory when a save fails, recovers from a corrupted index
- `atomic-write.spec.ts` (new file): rename-based write, replace-existing, `mode` honored, no `.tmp` left on failure
- The chmod-based tests skip on Windows and as root, where mode bits do not fail writes

### Manual Testing

Verified failure modes via read-only file/directory injection (EACCES) and truncated-JSON corruption. Biome, tsc and knip all clean.

## Checklist

- [ ] If this was for an open issue, I was assigned to it (assignment requested in #1130)
- [x] Self-review completed

### DIFF
--- a/.changeset/fix-timeline-index-atomic-write.md
+++ b/.changeset/fix-timeline-index-atomic-write.md
@@ -1,0 +1,5 @@
+---
+'@nanocollective/nanocoder': patch
+---
+
+Timeline index saves are now atomic: `timeline.json` is written to a temporary file and renamed into place instead of being overwritten in place. A crash or kill mid-save can no longer leave a truncated index behind, which previously cost the session every checkpoint (the corrupted index was discarded on next load). Thanks to @puri-adityakumar. Closes #1130.
--- a/.changeset/fix-timeline-index-atomic-write.md
+++ b/.changeset/fix-timeline-index-atomic-write.md
@@ -2,4 +2,4 @@
 '@nanocollective/nanocoder': patch
 ---
 
-Timeline index saves are now atomic: `timeline.json` is written to a temporary file and renamed into place instead of being overwritten in place. A crash or kill mid-save can no longer leave a truncated index behind, which previously cost the session every checkpoint (the corrupted index was discarded on next load). Thanks to @puri-adityakumar. Closes #1130.
+Timeline index saves are now atomic: `timeline.json` is written to a temporary file and renamed into place instead of being overwritten in place. If the process dies mid-save, readers only ever see the complete old or complete new index, so a torn write can no longer silently discard every checkpoint of the session. Thanks to @puri-adityakumar. Closes #1130.
--- a/source/services/timeline-manager.spec.ts
+++ b/source/services/timeline-manager.spec.ts
@@ -407,6 +407,152 @@ test.serial('TimelineManager refuses index paths that escape the workspace', asy
 	}
 });
 
+test.serial('TimelineManager saves the index atomically over a read-only live file', async t => {
+	const tempDir = await createTempDir();
+	try {
+		const manager = new TimelineManager(tempDir, 'session-atomic');
+		const indexPath = path.join(
+			tempDir,
+			'.nanocoder',
+			'timeline',
+			'session-atomic',
+			'timeline.json',
+		);
+		await manager.capture({
+			toolCallId: 'c1',
+			toolName: 'write_file',
+			title: 'step 1',
+			truncateToMessageIndex: 0,
+			files: filesMap([['a.ts', 'v1']]),
+		});
+
+		// rename(2) only needs write permission on the directory, so an atomic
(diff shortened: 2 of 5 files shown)


will-lamerton (changes requested):
Thanks for this - the diagnosis is right and the fix is correctly scoped. I verified locally: the spec passes on your branch (20/20), and reverting only `timeline-manager.ts` to the parent commit fails exactly one test (`saves the index atomically over a read-only live file`) with `EACCES ... open timeline.json`, so the discriminating test really does discriminate. biome and tsc are clean.

One change needed before merge:

**Use the shared atomic-write util instead of a fourth copy.** `source/utils/atomic-write.ts` already exists (`atomicWriteFileSync` / `atomicWriteJson`), and there are two private async duplicates in `source/session/session-manager.ts:64` and `source/memory/semantic-memory-manager.ts:111`. The helper added at `source/services/timeline-manager.ts:41` is a byte-identical third async copy. Please add the async variant to `source/utils/atomic-write.ts` and import it here:

```ts
export async function atomicWriteFile(
	filePath: string,
	data: string,
	options?: {mode?: number},
): Promise<void>
```

Migrating the two existing copies can be a follow-up; a new call site just shouldn't fork the pattern again.

Two optional notes:

- `saveIndex` assigns `this.index = index` before the write, and `loadIndex` returns the cached value, so a failed save leaves the process holding an index that never reached disk. Your "keeps the previous index intact" test only asserts the on-disk bytes, so it passes while the in-memory guarantee it names does not hold. Restoring the previous `this.index` in a catch would make the claim true end to end. Pre-existing behaviour, so your call.
- Both new permission tests rely on POSIX mode bits: as root (Docker dev containers) the `t.throwsAsync` case will not throw, and on Windows `chmod(dir, 0o555)` is a no-op. CI is non-root ubuntu so this is green today, but a `process.platform === 'win32' || process.getuid?.() === 0` skip guard would save contributors some confusion.

Nit: the changeset says a crash or kill mid-save "can no longer leave a truncated index behind". True for process death, which is what #1130 is about, but rename without fsync does not cover power loss. Not worth adding fsync on every capture, just worth softening the wording.

puri-adityakumar:
Thanks for verifying the discriminating test, and for pointing at the shared util I missed.

puri-adityakumar:
Hi, @will-lamerton I have made the required changes as request could you verify once again please.
Thank you

will-lamerton:
Great work @puri-adityakumar :)

PR #1207, opened 2026-09-06
fix(custom-commands): safely handle unreadable directories during scan

### LINKED ISSUE #1142: [Bug] custom-commands/loader.ts scan is not recursive-safe for namespace chains
## Description
`scanDirectory` recurses using `readdirSync(dir)` unguarded. A permission error on a sub-directory throws and aborts the entire scan, meaning a single unreadable directory (e.g., inside `.nanocoder/commands/`) silently disables all commands in that project.

## Environment
- **OS**: Any
- **Node version**: v22.20.0
- **Provider**: N/A
- **Model**: N/A

## Steps to Reproduce
1. Create a directory inside `.nanocoder/commands/` without read permissions.
2. Start nanocoder.

## Expected Behavior
The application should log a warning for the inaccessible folder and continue loading the valid commands.

## Actual Behavior
The entire scan aborts and no custom commands are loaded.

## Additional Context

Rudra2637:
## Description

Fixes #1142

In `CustomCommandLoader.scanDirectory`, directory traversal and file inspection used `readdirSync` and `statSync` unguarded. If a subdirectory (e.g. a namespaced folder inside `.nanocoder/commands/`) or file had restricted read permissions, Node.js threw an unhandled exception (such as `EACCES`). This caused the entire scan to abort prematurely and prevented any remaining valid custom commands in the project from being loaded.

This PR:
- Wraps `fs.readdirSync` and `fs.statSync` in `try...catch` within `scanDirectory` and `loadResources`.
- Logs a warning using `logWarning` when a directory or file cannot be read/inspected and continues scanning remaining directories and commands.
- Adds unit tests in `source/custom-commands/loader.spec.ts` ensuring that unreadable subdirectories and corrupted/unreadable entries don't abort the scan and valid commands are still loaded.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Changeset

### Automated Tests

### Manual Testing

- [x] Verified unit tests pass via `pnpm run test:ava source/custom-commands/loader.spec.ts`
- [x] Verified types pass via `tsc --noEmit`
- [x] Verified linting and formatting via Biome

## Checklist

- [x] Self-review completed

### DIFF
--- a/.changeset/fix-custom-commands-unreadable-dir.md
+++ b/.changeset/fix-custom-commands-unreadable-dir.md
@@ -1,0 +1,5 @@
+---
+"@nanocollective/nanocoder": patch
+---
+
+Fixed an issue in `CustomCommandLoader` where encountering an unreadable subdirectory or inaccessible file in custom command directories aborted the entire scan. `scanDirectory` and `loadResources` now catch filesystem permission and inspection errors, log a warning, and continue scanning remaining commands.
--- a/source/custom-commands/loader.spec.ts
+++ b/source/custom-commands/loader.spec.ts
@@ -1,8 +1,8 @@
-import { writeFileSync, rmSync, existsSync, mkdirSync } from 'fs';
-import { tmpdir } from 'os';
-import { join } from 'path';
+import fs, {existsSync, mkdirSync, rmSync, writeFileSync} from 'fs';
+import {tmpdir} from 'os';
+import {join} from 'path';
 import test from 'ava';
-import { CustomCommandLoader } from './loader';
+import {CustomCommandLoader} from './loader';
 
 // Helper to create a valid custom command file
 function createCommandFile(path: string, content: string) {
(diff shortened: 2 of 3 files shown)


will-lamerton (changes requested):
Thanks for this - the behaviour fix is correct and matches the guard already used in `source/custom-tools/loader.ts` and `source/skills/bundle-loader.ts`. Verified locally on your head: 34 tests pass, `tsc --noEmit` clean, biome clean.

One thing needs changing before merge.

**The production import was changed to create a mocking seam.** `loader.ts` goes from `import {existsSync, readdirSync, statSync} from 'fs'` to `import fs, {existsSync} from 'fs'` with `fs.readdirSync(...)` call sites, purely so the spec can do `(fs as any).statSync = ...`. That diverges from the closest sibling doing the identical job (`source/custom-tools/loader.ts:1`), and it makes the tests assert import style rather than behaviour.

Evidence: I ran the three new tests against the base `loader.ts`. The third one fails with `actual 2, expected 1` even though base already had `try { st = statSync(resourcePath) } catch { continue }` in `loadResources`. The only production change on that path is the added log line, so that test fails on base solely because the mock cannot reach a named-import binding.

Requested changes:

- Revert to named imports (`import {existsSync, readdirSync, statSync} from 'node:fs'`), matching custom-tools.
- Drive the tests with real permissions instead of patching `fs`: `chmod(dir, 0o000)`, restore in teardown, and skip when `process.platform === 'win32'` or `process.getuid?.() === 0` (root ignores mode bits, which will fail in some CI containers). That exercises the actual EACCES path and stays valid regardless of import style.
- Drop the `loadResources` test or make it assert the new `logWarning`. As written it covers pre-existing behaviour.

Minor, optional:

- `let stat;` is an evolving any. The sibling writes `let stat: ReturnType<typeof statSync>;`.
- The new tests assert exact `commands.length`, but `loadCommands()` also scans the real personal dir from `getConfigPath()`. Anyone with commands in `~/.config/nanocoder/commands` gets a failure. Pre-existing in this spec, but a `process.env.NANOCODER_CONFIG_DIR = <tmpdir>` in the fixture would close it.
- `logWarning` fires per entry, so a directory with many inaccessible files bursts the message queue at startup. One aggregated warning per directory would read better.

Out of scope, but worth a follow-up issue: `scanDirectory` uses `statSync`, which follows symlinks, so a symlink cycle under `.nanocoder/commands/` still recurses until the stack blows. Not introduced here.

Rudra2637:
Thanks for the clear feedback — completely agree on keeping production code free of test-mocking seams.

Updated as requested:
1. **Reverted to named imports:** In `source/custom-commands/loader.ts`, reverted back to `import {existsSync, readdirSync, statSync} from 'node:fs'` and direct call sites.
2. **Real permission-driven tests:** In `source/custom-commands/loader.spec.ts`, dropped `fs` monkey-patching and replaced it with real `chmodSync` tests (`0o000` for unreadable directory and `0o400` for unsearchable parent), restoring permissions in teardown and skipping when `process.platform === 'win32' || process.getuid?.() === 0`.
3. **Dropped the `loadResources` test:** Removed the redundant test that covered pre-existing behavior.
4. **Minor improvements addressed:**
   - Typed `stat` as `ReturnType<typeof statSync>`.
   - Added isolated `process.env.NANOCODER_CONFIG_DIR` fixture to prevent scanning the developer's personal config directory during tests.
   - Aggregated inspection warnings to fire once per directory rather than per file to avoid bursting the message queue.

All 33 tests pass, `tsc --noEmit` is clean, and Biome lint/format checks pass.

Rudra2637:
Will look after the follow up issue in a separate pr

will-lamerton:
Thanks for this PR @Rudra2637  - feel free to add yourself as a contributor to our website via a PR which I will approve :)

Rudra2637:
Thanks will raise a pr soon for that

Issue #1134, opened 2026-09-02
[Bug] bootstrap.ts silently maps unknown command sources to "project"


akramcodez:
## Description
The priority fall-through logic collapses anything that isn't `'project'` or `'personal'` to `'project'`. A future built-in commands loader would silently be marked as project priority, defeating the priority shadowing in `BundleLoader.scanLayer`.

## Environment
- **OS**: Any
- **Node version**: v22.20.0
- **Nanocoder version**: 1.30.0
- **Provider**: N/A
- **Model**: N/A

## Steps to Reproduce
1. Inject a command with a source other than 'project' or 'personal' (e.g., 'built-in').
2. Check its applied priority.

## Expected Behavior
It should map to an appropriate fallback or handle the new source correctly (perhaps throwing an error for unknown sources, or having a distinct fallback).

## Actual Behavior
It is silently mapped to `'project'`.

## Additional Context
- [x] I have searched existing issues to ensure this is not a duplicate
- [x] I can reproduce this issue consistently
- [x] This issue occurs with the latest version of nanocoder

awhite0030:
Picking this up - a fix will follow shortly.

PR #1208, opened 2026-09-06
fix: classify commands without source as built-in

Forzun:
Fixes the issue where custom commands without a defined source were incorrectly
classified as `project` skills.

## Changes

- Changed the fallback `SkillPriority` from `project` to `built-in`.
- Verified TypeScript and formatting checks.

fix:#1134

### DIFF
--- a/source/commands/custom-commands.tsx
+++ b/source/commands/custom-commands.tsx
@@ -183,8 +183,8 @@ function CommandDetail({command}: {command: CustomCommand}) {
 	const {colors} = useTheme();
 
 	const sourceLabel = command.path
-		? `${command.source ?? 'project'} (${command.path})`
-		: (command.source ?? 'project');
+		? `${command.source ?? 'built-in'} (${command.path})`
+		: (command.source ?? 'built-in');
 
 	const fields: Array<{label: string; value: string}> = [
 		{label: 'Source', value: sourceLabel},
--- a/source/custom-commands/loader.ts
+++ b/source/custom-commands/loader.ts
@@ -56,8 +56,8 @@ export class CustomCommandLoader {
 	 */
 	private scanDirectory(
 		dir: string,
-		namespace?: string,
-		source?: 'personal' | 'project',
+		namespace: string | undefined,
+		source: 'personal' | 'project',
 	): void {
 		const entries = readdirSync(dir);
 
@@ -90,8 +90,8 @@ export class CustomCommandLoader {
 	 */
 	private loadCommand(
 		filePath: string,
-		namespace?: string,
-		source?: 'personal' | 'project',
+		namespace: string | undefined,
+		source: 'personal' | 'project',
 		commandDir?: string,
 	): void {
 		try {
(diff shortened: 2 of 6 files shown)


will-lamerton (changes requested):
Thanks for picking this up, but I don't think this closes #1134.

`CustomCommand.source` is typed `'personal' | 'project'` and `loadCommands()` always passes one of them (`source/custom-commands/loader.ts:45,50`), so the fallback branch is only reached when `source` is `undefined`. The commands that actually have no `source` are bundle members: `bundle-loader.ts:276-282` builds the `CustomCommand` without one and `registrar.ts:76-89` registers it verbatim. Their real level is the bundle layer's priority, usually `project`, so mapping them to `built-in` is less accurate than the current default, not more. Today it's invisible only because `synthesizeCommandSkills` runs before `registerSkills` in `bootSkillPipeline`.

What I'd like instead:

- Propagate the true source for bundle commands in the registrar, mirroring how the subagent branch maps `skill.source.priority` through `subagentPriorityFor`. That likely means widening `CustomCommand.source` to include `'built-in'` and making it required.
- With that in place the ternary collapses to `const priority: SkillPriority = command.source ?? 'built-in';`, which is exhaustive by construction so a future source variant becomes a type error rather than a silent default.
- `source/commands/custom-commands.tsx:186-187` still does `command.source ?? 'project'`. Whatever default wins should apply there too, otherwise `/skills` and `/commands` label the same command differently.
- Please add a test asserting that a project bundle's command shows project priority in `/skills`.

Also: the branch needs a rebase on `main` (Config Schema Freshness fails on a missing `generate:schema` script and the coverage drift check is comparing against a moved baseline, neither is caused by this diff), and a patch changeset naming `@nanocollective/nanocoder` would be good for a user-visible labelling change.

will-lamerton (changes requested):
Closer. The type widening and the `/commands` alignment are right. Two things still block:

1. `bundle-loader.ts:283` hardcodes `source: 'project'`. `loadBundle` already has the layer `priority` (line 172), so please thread it into `loadCommandMembers` and set `source: priority`. As written, a personal bundle under `~/.config/nanocoder/skills/` and a built-in bundle both report `project`, which is the same mislabelling in a new place. `registrar.ts:96` already does this correctly for the subagent branch via `subagentPriorityFor`.

2. The new test passes with the `bundle-loader.ts` change reverted, I checked locally. `foo?.source.priority` reads the bundle Skill's layer priority, not the command member's, so it is green either way. Asserting the command's priority means `synthesizeCommandSkills` has to run after `registerSkills` (`bootstrap.ts:110` vs `:138`), or be exported and tested directly.

Two smaller things:

- `source` is required now, so the `?? 'built-in'` is dead in `bootstrap.ts:217` and `custom-commands.tsx:186-187`. `const priority: SkillPriority = command.source;` is the exhaustive form, which is what makes a future source variant a type error instead of a silent default.
- Could you revert the formatting-only hunks in `bootstrap.spec.ts` (double quotes, bracket spacing, wrapped arrow params)? Spec files are biome-excluded so CI stays green, but ~100 lines of churn buries the 20 lines of real change.

Then a rebase on `main` and a patch changeset naming `@nanocollective/nanocoder` and this is good to go.

PR #1210, opened 2026-09-06
fix: guard queued prompt draining across modes

mikemikimike:
## Summary / Problem

Follow-up to PR #1210. Queued prompts could start concurrently when a plan-review action dismissed the review bar, and a failed queued dispatch could be retried in a tight loop.

## Changes

- Keep queue draining disabled until the client and tool manager are ready.
- Stop re-dispatching the same queue head after a failed or throwing dispatch while keeping the message queued for an explicit retry/edit.
- Keep the queue blocked while plan clarification or plan modification is being requested.
- Signal conversation completion when chat dependencies are unavailable so a queued prompt cannot remain blocked indefinitely.
- Strengthen regression coverage with the real user-message queue, pending-plan guards, failed dispatches, and both busy and idle queue editing.
- Clarify the changeset issue reference as `Refs #1060`.

## Tests

- Focused AVA suites for `interactive-app`, `useAppHandlers`, `useChatHandler`, and `user-input`: all passed (31, 22, 24, and 52 tests respectively).
- `pnpm run test:format`, `pnpm run test:types`, `pnpm run test:types:vscode`, `pnpm run test:lint`, and `pnpm run test:changesets`: passed.
- `pnpm run test:audit` against the official npm registry: passed with no new vulnerabilities.
- `pnpm run test:knip`: passed with existing hints about ignored `.github/**` and unused tags.
- `pnpm run test:ava`: attempted on Windows; the changed suites pass, while platform-sensitive CLI, ACP, filesystem, subprocess, and network suites report existing failures/timeouts (90 failures, 245 pending after timeout, and one uncaught exception).
- `pnpm run build`: TypeScript compilation and alias rewriting passed; the final copy/chmod step cannot run on Windows because the package script uses Unix `cp` and `chmod`.

## Compatibility / Known limitations

The queue remains intact after a failed dispatch, but the same queue head is not automatically retried until the user edits or otherwise changes that queued item. This prevents an unavailable or failing setup from hot-looping.

## Issue

Refs #1060

### DIFF
--- a/.changeset/calm-queues-compact.md
+++ b/.changeset/calm-queues-compact.md
@@ -1,0 +1,5 @@
+---
+"@nanocollective/nanocoder": patch
+---
+
+Resume queued prompts after slash commands and manual context compaction complete. Closes #1060.
--- a/.changeset/calm-queues-compact.md
+++ b/.changeset/calm-queues-compact.md
@@ -2,4 +2,4 @@
 "@nanocollective/nanocoder": patch
 ---
 
-Resume queued prompts after slash commands and manual context compaction complete. Closes #1060.
+Resume queued prompts after slash commands and manual context compaction complete. Refs #1060.
--- a/.github/workflows/pr-labeler.yml
+++ b/.github/workflows/pr-labeler.yml
@@ -32,7 +32,7 @@ jobs:
               owner,
               repo,
               path: '.github/labeler.yml',
-              ref: context.payload.pull_request.head.sha,
+              ref: context.payload.pull_request.base.sha,
             });
             if (data.type !== 'file' || typeof data.content !== 'string') {
               throw new Error('labeler.yml is not a file');
--- a/.github/workflows/pr-labeler.yml
+++ b/.github/workflows/pr-labeler.yml
@@ -32,7 +32,7 @@ jobs:
               owner,
               repo,
               path: '.github/labeler.yml',
-              ref: context.payload.pull_request.base.sha,
+              ref: context.payload.pull_request.head.sha,
             });
             if (data.type !== 'file' || typeof data.content !== 'string') {
               throw new Error('labeler.yml is not a file');
--- a/source/app/App.tsx
+++ b/source/app/App.tsx
@@ -44,7 +44,6 @@ import {useUserMessageQueue} from '@/hooks/useUserMessageQueue';
 import {useVSCodeServer} from '@/hooks/useVSCodeServer';
 import {getAllSubagentProgress} from '@/services/subagent-events';
 import {generateKey} from '@/session/key-generator';
-import type {ImageAttachment} from '@/types/core';
 import type {ThemePreset} from '@/types/ui';
 import {createPinoLogger} from '@/utils/logging/pino-logger';
 import {setGlobalMessageQueue} from '@/utils/message-queue';
@@ -84,14 +83,6 @@ export default function App({
 	// Use extracted hooks
 	const appState = useAppState(initialDevelopmentMode);
 	const userMessageQueue = useUserMessageQueue();
-	const queuedUserSubmitRef = React.useRef<
-		| ((
-				message: string,
-				displayValue: string,
-				images?: ImageAttachment[],
-		  ) => Promise<void>)
-		| null
-	>(null);
 	const {exit} = useApp();
 	const {isTrusted, handleConfirmTrust, isTrustLoading, isTrustedError} =
 		useDirectoryTrust();
@@ -249,35 +240,6 @@ export default function App({
 		}
 	}, []);
 
-	const drainQueuedUserMessage = React.useCallback(() => {
-		// Defer to a macrotask, not a microtask. `onConversationComplete` fires
-		// deep inside the finishing turn's await chain, so a microtask drain would
-		// start the next turn BEFORE that turn's `resetStreamingState()` finally
-		// runs — and the stale reset would then wipe the new turn's abortController
-		// and isGenerating, leaving the busy indicator (and Escape-to-cancel) dead.
-		// A timeout runs after those continuations, so the drained turn keeps its
-		// busy state.
-		setTimeout(() => {
-			void userMessageQueue.drainNextMessage(async message => {
-				const submitQueuedMessage = queuedUserSubmitRef.current;
-				if (!submitQueuedMessage || !appState.client || !appState.toolManager) {
-					return false;
-				}
-
-				await submitQueuedMessage(
-					message.message,
-					message.displayValue,
-					message.images,
-				);
-				return true;
-			});
-		}, 0);
-	}, [
-		appState.client,
-		appState.toolManager,
-		userMessageQueue.drainNextMessage,
-	]);
-
 	// Setup chat handler
 	const chatHandler = useChatHandler({
 		client: appState.client,
@@ -299,7 +261,6 @@ export default function App({
 			appState.setCompactToolCounts(null);
 			appState.compactToolCountsRef.current = {};
 			appState.setLiveTaskList(null);
-			drainQueuedUserMessage();
 		},
 		// A turn that started in plan mode finished uninterrupted — a plan was
 		// produced. Flag it so the interactive UI can show the plan review bar.
@@ -577,10 +538,6 @@ export default function App({
 		handleMessageSubmit: appHandlers.handleMessageSubmit,
 		activeEditor: vscodeServer.activeEditor,
 	});
-
-	React.useEffect(() => {
-		queuedUserSubmitRef.current = handleUserSubmit;
-	}, [handleUserSubmit]);
 
 	// Setup non-interactive mode
 	const {nonInteractiveLoadingMessage} = useNonInteractiveMode({
--- a/source/app/sections/interactive-app.spec.tsx
+++ b/source/app/sections/interactive-app.spec.tsx
@@ -162,10 +162,28 @@ function makeProps(o: Overrides = {}) {
 				displayValue: '',
 			}),
 			removeMessage: noop,
-			drainNextMessage: o.drainNextMessage ?? (() => false),
+			drainNextMessage: o.drainNextMessage ?? (async () => false),
 		},
 		handleIdeSelect: noop,
 	} as never;
+}
+
+function QueuedPromptHarness({overrides}: {overrides: Overrides}) {
+	const userMessageQueue = useUserMessageQueue();
+
+	React.useEffect(() => {
+		userMessageQueue.enqueueMessage({
+			message: 'queued prompt',
+			displayValue: 'queued prompt',
+		});
+	}, [userMessageQueue.enqueueMessage]);
+
+	return (
+		<InteractiveApp
+			{...makeProps(overrides)}
+			userMessageQueue={userMessageQueue}
+		/>
+	);
 }
 
 test('renders without crashing in default state', t => {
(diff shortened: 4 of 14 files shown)


will-lamerton (changes requested):
The design is right and the fix is real. `/compact` is genuinely covered end to end: `compact-handler.ts` wraps `summariseWithLLM` in `setIsToolExecuting`, so `isConversationComplete` flips `false -> true` at the correct moment and the effect drains. I verified the new `planReviewState` guard is load-bearing by deleting the clause, and the suite breaks. 173 focused tests pass locally.

One defect remains, in the same class the PR set out to close.

## Blocking: the drain still starts a concurrent turn when the plan review bar is dismissed

`queueDrainBlocked` (`interactive-app.tsx:192`) covers `planReviewState?.show` and `pendingPlanProceed`, which closes the "proceed" path. It does not cover the other two exits from that bar.

`handlePlanAskMore` (`useAppHandlers.tsx:645`) clears `planReviewState` and dispatches through `handleChatMessage`, not `handleMessageSubmit`. Only `handleMessageSubmit` resets `isConversationComplete` (`:675`), and `handleChatMessage` awaits `import('@/stats/record')` (`useChatHandler.tsx:341-352`) before flipping any generating flag. React commits a render where the bar is gone, nothing is generating, and the conversation still reads complete. The drain fires there.

Confirmed with your own `makeProps` harness rather than inferred. Control with nothing blocking drains `["plain"]`. A probe replicating the ask-more body exactly (clear the bar, then an async function awaiting a dynamic import before setting `isGenerating`) drains `["queued while planning"]`, a second turn dispatched underneath the clarifying-questions turn.

`handlePlanModify` (`:654`) has the same shape: it clears the bar, prints "Plan Mode remains active. Tell Nanocoder what to change.", and the drain immediately submits a stale queued prompt into that moment.

Not a regression, since pre-PR the drain fired even earlier while the bar was still up. But it is the same failure mode one step downstream. Cheapest fix is for `handlePlanAskMore` to call `setIsConversationComplete(false)` before clearing the bar, since it really is starting a turn. `handlePlanModify` is a judgement call.

## The `pendingPlanProceed` guard has zero coverage

Replacing `appState.pendingPlanProceed !== null` with `false` leaves all 29 tests in `interactive-app.spec.tsx` passing. The guard is correct and necessary, since `handlePlanProceed` clears `planReviewState` and sets `pendingPlanProceed` in the same batch (`useAppHandlers.tsx:634-636`), but nothing pins it.

## The three "does not drain" tests assert nothing

`t.false(submitted)` holds whether or not the guard works, because the default `drainNextMessage` mock is `() => false`. They only detect a regression by crashing the file with `TypeError: userMessageQueue.drainNextMessage(...).then is not a function`. When I removed the `planReviewState` clause, the plan-review test itself still reported as passing; only the non-zero file exit revealed the break. Make the default mock async and count dispatch attempts. This was flagged on #1069 and carried over unchanged.

## Smaller notes

`drainInProgressRef` is released only inside `.then`. Nothing throws synchronously today, but if `drainNextMessage` ever did, the ref would latch true and the queue would stop draining silently.

The second commit touching `.github/workflows/pr-labeler.yml` nets to zero against main (the file list is 10 files, no labeler entry). Worth squashing.

The `create-handler.ts` await-then-complete asymmetry raised on #1069 is still unexplained. The new `retry-handler.ts` comment justifies removing the signal there but not why the sibling keeps it.

`user-input.spec.tsx` still has a trailing-newline-only hunk at EOF.

The first finding is what I want addressed before merge. The next two are cheap in the same pass.

will-lamerton (changes requested):
Thanks for the clean rewrite of #1069. The core idea is right and it does fix #1060: keying off state rather than the `onConversationComplete` edge is what lets `onCommandComplete` (slash commands, `/compact`) resume the queue, and it lets a blocked drain retry once the blocker clears. Two things need fixing before merge.

### 1. A failed dispatch becomes an unbounded retry loop

`drainNextMessage` pops the message and, on a falsy or throwing dispatch, pushes it back at the front and returns `false` (`useUserMessageQueue.ts:57-63`). Under the old edge-triggered drain that was safe: the message waited for the next completion edge. Under the new effect the restore mutates `queuedMessages`, the effect re-runs, and it re-dispatches the same message immediately.

Measured on this branch with an instrumented render:

- `handleUserSubmit` rejecting: 65 re-dispatches in 300 ms
- `toolManager` null (the `return false` at `interactive-app.tsx:225`): 101 re-dispatches in 300 ms

Both spin for as long as the condition holds, no backoff, no attempt cap, two re-renders of the queued list per iteration. The reachable trigger is a rejection out of `handleUserSubmit`: it awaits `runLifecycleHooks('user-prompt-submit', ...)` and `handleMessageSubmission` (`useAppHandlers.tsx:711,746`), and `handleChatMessage` does real work before its own try block (`useChatHandler.tsx:347-392`). Anything throwing there turns one bad queued prompt into a hot loop that may also append a chat bubble per attempt.

Suggested fix, both parts:

- Hoist the `!appState.client || !appState.toolManager` check into the effect's early return instead of doing it inside the dispatch. Both are already deps, so the drain resumes on its own once they populate and the queue is never popped-then-restored.
- Do not re-enter for a message that just failed. A `lastFailedIdRef` that is skipped while it is still at the head of the queue is enough, since any real state change can reset it.

### 2. The three new negative tests cannot fail

`does not drain queued prompts while a turn is generating`, `... while a modal mode is active`, and `... while plan review is active` all use `makeProps`'s default `drainNextMessage: () => false`, which never invokes the dispatch callback, so `submitted` stays `false` regardless of the guards.

Confirmed by deleting `queueDrainBlocked`, `activeMode`, `isSettingsMode`, and `isConversationComplete` from the effect's early return: all 29 tests in `interactive-app.spec.tsx` still pass. The guards this PR exists to add have no coverage.

Please give those three the real `useUserMessageQueue` (the harness pattern the two positive tests already use). Also make the default mock `async () => false`: production calls `.then()` on the result (`interactive-app.tsx:233`) and the `as never` cast on props hides the mismatch from tsc.

### Smaller notes (non-blocking)

- Dropping `onCommandComplete` from `/retry` is correct for the normal path, but `handleChatMessage` early-returns at `useChatHandler.tsx:346` on `!client || !toolManager` without signalling completion. That leaves `isConversationComplete` stuck at `false` and the queue permanently undrainable. Cheapest fix is to signal completion in that early return, which also closes the same gap on the plain chat path.
- `user-input.spec.tsx:1165` gains a CRLF line ending. `biome.json` excludes `**/*.spec.tsx`, so `test:format` will not catch it despite `lineEnding: "lf"`. Same exclusion is why the new `@/constants` and `@/hooks/...` imports sitting above `strip-ansi` in `interactive-app.spec.tsx` pass unsorted.
- The `isBusy` removal in `user-input.tsx` is fine: `ChatInput` renders `UserInput` only in the else branch of the approval/confirmation/question ternary (`chat-input.tsx:211-232`) and the composer is unmounted for modals, settings, plan review, and input-capturing live components, so there is no arrow-key contention. But the spec swapped the busy case for the idle case rather than covering both, so `loadSelectedQueuedMessage` while busy is now untested.
- If you keep the `void drainAttempt;` line, say in the comment that it is there to satisfy `useExhaustiveDependencies`.
- The changeset says "Closes #1060", the PR body says "Refs #1060". Pick one.

Verified locally: `ava` on the five touched specs passes on macOS (109 tests), `biome check` on the changed non-spec files is clean.

mikemikimike:
Addressed the review feedback in `c96949b`:

- Guard queue draining until both the client and tool manager are available.
- Keep a failed queue head queued without immediately retrying the same message, including synchronous throws.
- Mark plan clarification and plan modification as incomplete before dismissing the review bar.
- Signal completion when chat dependencies are unavailable.
- Replace the no-op drain mocks with the real queue in the negative tests, cover pending plan proceed, and cover busy queue editing.
- Changed the changeset wording to `Refs #1060`.

Focused tests and all required remote checks pass. The only remaining red check is the unrelated Linux sandbox-jail job, whose AVA command filters out the requested `source/services/bash-executor-sandbox.spec.ts` file before running it.

will-lamerton (changes requested):
All three blocking findings are genuinely fixed. I verified each by deleting the fix and re-running rather than reading for intent: the `client`/`toolManager` hoist plus `lastFailedDrainIdRef` stop the hot loop (1 dispatch in 300ms with the guard, 30 without, on a probe whose dispatch yields to React before throwing), and removing the `setIsConversationComplete(false)` lines in `handlePlanAskMore`/`handlePlanModify` or the `onConversationComplete?.()` in `handleChatMessage` fails their tests. `pendingPlanProceed` now has coverage, the three negative tests use the real queue and genuinely fail when their guards are removed, and the labeler scope creep and changeset wording are resolved. The code I would sign off on as-is.

One thing to fix before merge.

## `does not immediately retry a failed queued dispatch` cannot fail

It is the test for this PR's headline fix, and it passes with `lastFailedDrainIdRef` removed. Same class as the vacuous tests flagged on #1069 and in my last review.

Its `handleUserSubmit` throws without ever awaiting, so the pop and the restore land in one batch, `queuedMessageCount`/`queuedMessageId` return to their prior values, the dep array never changes and the effect never re-runs. It passes for a reason unrelated to the guard. Production `handleUserSubmit` awaits `runLifecycleHooks` and `handleMessageSubmission`, so the intermediate `count=0` render does commit and the guard is what stops the loop.

One line, `interactive-app.spec.tsx:292`:

```
handleUserSubmit: async () => {
	dispatchAttempts++;
	await new Promise(r => setTimeout(r, 5));
	throw new Error('dispatch failed');
},
```

With the await it reports 30 attempts when the guard is removed and 1 when it is present.

## Two load-bearing guards with no coverage (cheap, same pass)

Both survive deletion with all 33 tests green, and both have real failure modes I confirmed with probes:

- `!appState.isConversationComplete`: remove it and a queued prompt dispatches mid-turn. This is the guard the whole `handlePlanModify` fix leans on. The fix is pinned at the handler level but nothing pins the effect that consumes it.
- `!appState.client || !appState.toolManager`: remove it and the queued prompt is popped and silently lost (still-queued goes 1 to 0), because `handleChatMessage` early-returns and the dispatch reports success. Quieter than the hot loop it replaced, but worse for the user.

## Smaller notes

- `user-input.spec.tsx:1243` still adds a CRLF-only line at EOF. Third flag. `biome.json` excludes `**/*.spec.tsx`, so CI will never catch it.
- The `create-handler.ts` asymmetry is moot. Both `await onHandleChatMessage(...)` paths (`:82`, `:474`) already delegate completion to the chat turn, so the retry change brings it into line rather than away from it. Worth saying so and closing the thread.
- A permanently failing head now blocks the whole queue for the session with no user-visible signal. Documented in the PR body and it follows the design I suggested, so not a blocker, but a "queued prompt failed, press up to edit" notice would be a good follow-up.

The first item is what I want before merge. The next two are cheap in the same pass. All 16 checks pass and the branch is cleanly mergeable; 206 tests green across the six touched specs.

mikemikimike:
Pushed `1ccfd7b5` with the remaining review fix. The failed-dispatch regression now holds the first dispatch open across a render boundary before releasing the rejection, so removing `lastFailedDrainIdRef` causes the test to retry and fail instead of passing vacuously. Focused AVA: `does not immediately retry a failed queued dispatch` passed; pre-commit/lint-staged passed. Please re-review the updated head.

will-lamerton:
Thanks @mikemikimike, all three items from my last review are genuinely fixed. I checked each by deleting the fix and re-running rather than reading for intent: removing `lastFailedDrainIdRef` fails `does not immediately retry a failed queued dispatch`, removing `!appState.isConversationComplete` fails the new incomplete-conversation test, and removing `!appState.client || !appState.toolManager` fails both of its new tests. The CRLF-only line at EOF in `user-input.spec.tsx` is gone too.

I pushed `2431d8d6` to your branch myself rather than send it back for another round, since it is caused by something that landed after your last push rather than by anything in your diff. Sorry for touching your branch without asking first.

Architect mode (#909) merged to main yesterday and adds a fifth mode with a turn-boundary keep / revert / revert-and-revise gate. That gate has the same shape as the plan review bar: the turn is complete and nothing is generating while it is on screen, so the drain fired straight underneath it. I confirmed it with your own `QueuedPromptHarness` against a local merge of main: with `architectReviewState.show === true` and a queued message it dispatched 1 turn, where the plan review control in the same tree dispatches 0.

It is worse than the plan case. The gate's revert button restores the checkpoint taken before the first turn's mutations, so a queued prompt that drained in behind the bar has its file edits rolled back too when the user presses revert.

There were two gaps, both mirroring fixes you already made here for plan mode:

- `queueDrainBlocked` now covers `appState.architectReviewState?.show`.
- `handleArchitectRevertAndRevise` now calls `setIsConversationComplete(false)` before dismissing the bar. It dispatches through `handleChatMessage`, which never resets the flag, so the render where the bar is gone and the revise turn has not started was drainable. Exactly the `handlePlanAskMore` shape from my first review.

Both are pinned by tests that fail when the fix is removed. The commit also merges current main and fixes one collision: `cb342d46` changed the `handleChatMessage` spy wrapper to two arguments, so the plan ask-more test's call-shape assertion needed `undefined` as the second element.

227 tests pass across the seven touched specs, `tsc --noEmit` is clean, and biome is clean on both changed files.

Merging once CI is green. Note for anyone watching: `source/commands/stats.spec.tsx` "StatsDisplay changes range with arrow keys and closes on Escape" was failing on the previous run here, but it fails on a clean `main` too, so it is not from this PR.

will-lamerton (approved):
Approving. All three findings from my last review are fixed and verified by deleting each fix and re-running, not by reading for intent.

The `pr-checks / Unit Tests & Coverage Analysis` failure on the previous head was a stale base, not this PR. It was `source/commands/stats.spec.tsx` "StatsDisplay changes range with arrow keys and closes on Escape" timing out, and this PR does not touch stats. `main` had since landed `d04357d5` and `70dd2cbc` de-flaking that spec. I merged current `main` in `87aa422e` and the job is green.

Final state: 16/16 checks pass, 234 tests green across the seven touched specs plus `stats.spec.tsx` locally, `tsc --noEmit` clean, biome clean on both changed production files.

Follow-up worth opening separately, from my last review and not a blocker: a permanently failing queue head now blocks the queue for the rest of the session with no user-visible signal. A "queued prompt failed, press up to edit" notice would close that.

PR #1216, opened 2026-09-06
fix(notifications): prevent PowerShell string injection on Windows (#1138)

### LINKED ISSUE #1138: [Bug] notifications.ts Windows PowerShell string injection
## Description
The script uses single-quote interpolation for title and message and only escapes single quotes (`'` -> `''`). Backticks ( `` ` `` ) inside the user-controlled title/message may allow arbitrary code execution by PowerShell depending on the evaluation context (or simply break the notification if PowerShell attempts to resolve variables).

## Environment
- **OS**: Windows
- **Node version**: v22.20.0
- **Provider**: N/A
- **Model**: N/A

## Steps to Reproduce
1. On Windows, trigger a notification where the title is `` `whoami` `` or similar.
2. Observe the notification.

## Expected Behavior
The notification text should safely render the exact literal characters.

## Actual Behavior
Code might be executed or formatting might break.

## Additional Context

rishu685:
## Summary
Fixes #1138 by eliminating PowerShell script string interpolation when displaying native notifications on Windows.

## Details
- `sendWindows` previously interpolated notification `title` and `message` directly into a PowerShell script string passed to `powershell -Command`.
- User-controlled strings containing backticks (`` ` ``), quotes, or PowerShell variables could trigger command evaluation or break notification delivery.
- Refactored `sendWindows` to use a completely static PowerShell script executed via `-EncodedCommand` (UTF-16LE Base64) with `-NonInteractive` and `windowsHide: true`.
- Titles and messages are safely passed as Base64-encoded UTF-8 environment variables (`NANOCODER_NOTIFICATION_TITLE` and `NANOCODER_NOTIFICATION_MESSAGE`), decoded at runtime inside PowerShell via `[System.Convert]::FromBase64String` and `[System.Text.Encoding]::UTF8.GetString`.
- Added unit tests verifying script non-interpolation and roundtrip accuracy for tricky characters and injection vectors.

## Testing
- `pnpm run test:ava source/utils/notifications.spec.ts` (15/15 passed)
- `pnpm run format:check`
- `pnpm run test:lint`
- `pnpm run test:types`

### DIFF
--- a/.changeset/fix-windows-notification-powershell-injection.md
+++ b/.changeset/fix-windows-notification-powershell-injection.md
@@ -1,0 +1,6 @@
+---
+"@nanocollective/nanocoder": patch
+---
+
+Hardened Windows native notifications. Notification title and message are no longer interpolated into the PowerShell script; they are passed out-of-band as environment variables and read by a static script. Fixes rendering and parsing edge cases with backticks, quotes and other special characters. Closes #1138.
+
--- a/source/utils/notifications.spec.ts
+++ b/source/utils/notifications.spec.ts
@@ -1,5 +1,7 @@
+import childProcess from 'child_process';
 import test from 'ava';
 import {
+	buildWindowsNotificationPayload,
 	getNotificationsConfig,
 	sendNotification,
 	setNotificationsConfig,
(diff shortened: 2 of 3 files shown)


will-lamerton:
Hey @rishu685 - please can you address the points in nc-review above? :)

will-lamerton (changes requested):
Thanks for this - the direction is right and I want to merge it. Passing the user strings out-of-band instead of splicing them into the script is the correct shape. Three things first.

**1. The changeset overstates the fix (please reword).**

PowerShell single-quoted strings are verbatim: inside `'...'`, backticks, `$(...)` and `$env:X` are all literal, and the only escape is `''`, which the old code applied to every `'`. There was no break-out, so no command execution. #1138 hedges on this ("may allow ... depending on the evaluation context") but the changeset states it as fact, and that text ships to the public changelog.

Suggested:

```
Hardened Windows native notifications. Notification title and message are no longer interpolated into the PowerShell script; they are passed out-of-band as environment variables and read by a static script. Fixes rendering and parsing edge cases with backticks, quotes and other special characters. Closes #1138.
```

**2. Stub `execFile` in the win32 test** (the nc-review point).

`sendNotification handles win32 platform gracefully` spawns a real `powershell.exe` on a Windows runner and `ENOENT`s into the swallowing callback everywhere else, so `t.notThrows` only proves nothing throws synchronously. Please spy on `execFile` and assert the command that would run is the static `-EncodedCommand` form. That also closes the real gap: nothing currently verifies `sendWindows` passes the payload through to `execFile`, so a regression that reintroduced interpolation there would pass every test in the file.

Related: `t.false(decodedScript.includes(title))` cannot fail as written. `WINDOWS_NOTIFICATION_ENCODED_COMMAND` is a module constant computed at import and returned unmodified, so those assertions hold no matter what the function does with `title`. Fine to keep as an invariant, but they aren't the safety evidence they read as.

**3. Drop the base64 on the env vars.**

Env vars are already a data channel, not a code channel - `$env:FOO` yields a string and assigning it to `.BalloonTipTitle` never evaluates it. Windows env vars are natively UTF-16, so emoji, newlines and quotes survive without encoding. Removes the decode branch and the `else { '' }` fallback on both sides:

```powershell
$notify.BalloonTipTitle = $env:NANOCODER_NOTIFICATION_TITLE
$notify.BalloonTipText = $env:NANOCODER_NOTIFICATION_MESSAGE
```

Keep `-EncodedCommand` - it sidesteps `powershell.exe`'s `-Command` quoting quirks and the script is a constant now, so nothing is lost.

Nits, take or leave:
- Spreading `process.env` is load-bearing (`powershell.exe` needs `SystemRoot`/`PATH`) since passing `env` disables implicit inheritance. Worth a comment so nobody trims it to the two keys later.
- Stray blank line at the end of `notifications.spec.ts`. Biome won't flag it, `biome.json` excludes `**/*.spec.ts`.

rishu685:
Made the requested changes,can you review it @will-lamerton

will-lamerton:
Thanks for this PR @rishu685 - feel free to add yourself as a contributor to our website via a PR which I will approve :)

PR #1225, opened 2026-09-07
ci: call the shared nc-review workflow instead of a local copy

will-lamerton:
Step 8b of the ops plan. The machinery moves to [`Nano-Collective/.github`](https://github.com/Nano-Collective/.github/blob/main/.github/workflows/nc-review.yml); the rubric and trigger wiring stay here. **729 lines of workflow become 22.**

## Why this is the unblocking step

nc-review has run on this repo alone since it was built, so `copilot_code_review` still holds Review gates on the other **eight** repos. Removing it there before nc-review arrives would leave those repos with no automated review at all, which inverts Q7. Centralising is what makes adoption a two-file change per repo instead of a 760-line copy.

## The split

Same as `pr-checks`: shared machinery, local judgement.

| | |
|---|---|
| **Shared** | trigger filtering, the maintainer check on `/re-review`, context assembly, the sandboxed Nanocoder run, verdict schema validation, comment and label |
| **Local** | `.github/nc-review/rubric.md`, and this caller |

The rubric is deliberately **not** shared. It is where a project says what it cares about — its conventions, its test layout, the mistakes it keeps seeing. A shared rubric would produce reviews that read the same everywhere and land nowhere.

## What actually changed

The workflow was already almost entirely generic — it addresses the repo through `github.repository`, and the Nanocoder clone is the review *runtime*, not the repo under review. Four things needed fixing to make it portable:

- **Provider, model, base URL and context limit are inputs**, so a repo can review with something other than MiniMax without forking the workflow.
- **The Nanocoder runtime source is an input.**
- **`workflow_dispatch`'s `pr_number` is the caller's input**, so the shared workflow takes its own `pr-number` and this file passes it through.
- **The reading list is assembled at run time** rather than naming `CLAUDE.md`, `CONTRIBUTING.md` and the rubric literally. An absent architecture document is now omitted rather than sent to the agent — naming a file that does not exist wastes its first turns on a failed read and, worse, invites it to guess at the architecture instead of reporting that it could not find one. **Only this repo has a `CLAUDE.md`**, so every other adopter would have hit that.

Behaviour here is unchanged: the defaults are exactly what this repo used, and `architecture-doc: CLAUDE.md` is passed explicitly.

## ⚠️ This PR cannot exercise the new path

`pull_request_target` resolves the workflow from the **base** branch, so the caller has to be on `main` first — the same constraint `changeset-check` has. **Expect no nc-review comment on this PR. That is correct, not a failure.**

Verification plan after merge: `workflow_dispatch` against an open PR, which does exercise the full shared path. If it misbehaves, reverting is one file.

## What the rest of the rollout needs

`MINIMAX_API_KEY`, and it should be **one organisation secret**, not eight copies. Organisation Actions secrets with visibility `all` work for public repositories on the **Free** plan — I verified that empirically by creating and deleting a probe secret, because the plan's capability table doesn't cover it. This repo already holds the key as a repo secret and keeps working either way.


will-lamerton:
Both points were worth raising. Checked both.

**The `if:` gating — verified intact, no regression.** You were right that the caller cannot show it and right to flag it rather than assume. The whole `review:` job moved across verbatim; only the `on:` and `concurrency:` blocks were replaced. The shared workflow at `Nano-Collective/.github` carries:

```yaml
  review:
    if: |
      (
        github.event_name == 'pull_request_target'
        && github.event.pull_request.draft == false
        && github.event.pull_request.head.ref != 'changeset-release/main'
      )
      || (
        github.event_name == 'issue_comment'
        && github.event.issue.pull_request != null
        && startsWith(github.event.comment.body, '/re-review')
        && contains(fromJSON('["OWNER","MEMBER","COLLABORATOR"]'), github.event.comment.author_association)
      )
      || github.event_name == 'workflow_dispatch'
```

so the draft exclusion, the `changeset-release/main` exclusion and the author-association filter all survive — as does the *"Verify commenter can merge"* step, which re-checks the real permission through the API before spending a model call, because association is not a permission check. `github.*` in a called workflow resolves against the caller's event, so these evaluate exactly as before.

The one behavioural difference is cosmetic: the caller job always starts and is skipped when the called job's `if:` is false, so a non-`/re-review` comment now shows a skipped job rather than nothing.

**The changeset — fair, and the precedent is real.** `.changeset/coverage-drop-vs-base.md` is exactly the analogous entry for the `pr-checks` caller conversion, with empty frontmatter and a one-line CI note. Added `.changeset/nc-review-shared-workflow.md` in the same shape.

**One correction to my own PR description.** I wrote that this PR could not be reviewed and to expect no comment. Wrong — `pull_request_target` resolved nc-review from `main`, which still held the old workflow, so it reviewed this PR with the pre-refactor version. What remains true is that it did not exercise the *new* shared path; that still needs a `workflow_dispatch` after merge.

PR #1226, opened 2026-09-07
ci(nc-review): reduce the rubric to Nanocoder's own half

will-lamerton:
Follow-up to #1225, and a correction to something I claimed in it.

I said there that the rubric was project-specific and therefore had to stay per repo. That was asserted, not measured. Measured: **of 321 lines, five mentioned anything about Nanocoder.** The rest — base-checkout semantics, the tool list, the severity scale, the verdict rules, the output schema, what not to do — is the reviewing *method*, identical for every project.

Copying that into nine repositories would have rebuilt exactly the drift problem centralising `pr-checks` existed to remove. Worse, the parts most likely to be corrupted by a copy-paste edit are the output schema and the base-commit semantics — the two things a wrong review depends on.

So the rubric is now two files:

| | |
|---|---|
| `nc-review/rubric-base.md` in `.github` | the method — shared, fetched at run time |
| `.github/nc-review/rubric.md` here | what **Nanocoder** cares about |

This PR is the second half. The shared base and the workflow change are already on `main` in [`Nano-Collective/.github`](https://github.com/Nano-Collective/.github).

## What is left in this file

A real overlay, not a trimmed copy:

- the `CLAUDE.md` conventions that actually come up in review — tool registry, `useAppState`, the lazy command registry, and Ink's constraints
- which surfaces count as public contracts (CLI flags, `agents.config.json`, tool interfaces, the session format and `RunRecord`, the VS Code contributions)
- test conventions, plus the two failure modes that **have shipped here**: a test that asserts nothing meaningful, and a test that names a failure path it never triggers
- the changeset rules, including the empty-changeset convention and why a bad package name is worse than a missing changeset
- that this repo takes many first-time contributions, so scope should be judged on justification rather than on whether it was pre-agreed

## Verified

The shared workflow was confirmed working end to end before this change — `workflow_dispatch` against #1207 produced a genuine review through `nc-review / review` (the reusable-workflow prefix), applied `agent:clean`, and rendered the *"Addressed since the last review"* block.

The base-rubric fetch is asserted non-empty and structurally sane before use. A 404 or truncated read would otherwise produce a confident review with no method behind it, which is worse than no review.


will-lamerton:
Good finding, and half right in the way that matters — so thank you for repeating it.

**The bug is not real.** The shared workflow resolves the PR number itself, and does not rely on the forwarded input:

```yaml
- name: Resolve PR number
  env:
    N: ${{ github.event.pull_request.number || github.event.issue.number || inputs.pr-number }}
```

`github.event` inside a reusable workflow is the *caller's* event, so `pull_request_target` and `issue_comment` are both covered by the first two terms and `pr-number` is only the manual-dispatch fallback.

**The criticism of my verification was correct**, though — I had only exercised `workflow_dispatch`, which is exactly the one trigger where the input is populated, and said so as if it covered the rest. So I have now exercised the `issue_comment` path directly, by posting `/re-review` on this PR:

```
Resolve PR number:              success
Verify commenter can merge:     success   ← the maintainer gate ran, not skipped
Fetch the shared reviewing method: success
Post review:                    success
```

It posted to **this** PR, which is the proof the resolution worked — the number came from `github.event.issue.number`, since `inputs.pr_number` was empty on that trigger.

`pull_request_target` remains unexercised through the new caller and will be covered by the next PR opened on this repo. It uses the same resolve step, with `github.event.pull_request.number` as the first term.

**I have not applied the suggested fix**, deliberately. Duplicating

```yaml
pr-number: ${{ github.event.pull_request.number || github.event.issue.number || inputs.pr_number }}
```

into every caller puts the same expression in nine files with nothing keeping them in step, which is the drift the caller pattern exists to avoid. Instead the reason now sits next to the line in `templates/nc-review.caller.yml`, so the next reader — human or otherwise — does not have to leave the repository to check it.

PR #1228, opened 2026-09-07
fix(tools): tool-execution safety - approval bypass on bad args unbounded output orphaned processes

### LINKED ISSUE #1227: Bug: custom-tool execution  approval bypass on malformed args unbounded output buffering orphaned processes on timeout
Three tool-execution safety bugs found while auditing the approval and
custom-tool runtime paths.

**1. Approval bypass on malformed args** (`tool-confirmation.tsx:140-151`)
When tool args fail schema validation, the approval prompt is skipped and the
call is auto-approved (`onConfirm(true)`). Tools can run without user consent.
Fix: never auto-resolve `true` — show the validation error with the approval
prompt and let the user decide.

**2. Unbounded output buffering** (`custom-tools/handler.ts:63-81`)
Custom-tool stdout/stderr is accumulated with no byte limit; truncation only
runs after the process exits. A large-output tool can exhaust memory. The
built-in bash tool already caps this at 5 MB (`bash-executor.ts`). Fix: mirror
that byte-budget pattern while streaming.

**3. Timeout orphans processes and hangs forever** (`custom-tools/handler.ts:57-74,88-102`)
On timeout only the direct shell is SIGTERMed; grandchildren are not killed and
can hold the pipes open, so the `close` event never fires and the call never
returns. `bash-executor.ts` fixes this with `detached` + process-group kill +
stream destroy. Fix: replicate that cleanup on timeout.

All three reproduced on current code. Regression tests can be provided.
Labels: bug, area:tools, security

tusharui:
Closes #1227

1. Schema-validation errors auto-approved the tool call
   (`tool-confirmation.tsx`). `onConfirm(true)` fired whenever
   `hasValidationError` was set, so a call with malformed arguments was
   executed without ever rendering a consent prompt. Removed the branch; the
   prompt now renders alongside the validation error and the user decides.
   Approving still re-runs the validator, so the error reaches the model for
   correction through the existing path.

2. Custom-tool output was buffered unbounded (`custom-tools/handler.ts`).
   stdout/stderr accumulated until process exit, then got truncated. A tool
   printing >5 MB of output was fully kept in memory first. Mirror the
   built-in bash executor: cap combined stdout/stderr at
   `BASH_MAX_OUTPUT_BYTES` while streaming and surface a notice at the head
   of the result.

3. Timeout killed only the shell and could never settle
   (`custom-tools/handler.ts`). A backgrounded grandchild holding an inherited
   pipe kept `close` from firing, so the tool call hung past the timeout with
   the orphan left running. On Unix the shell is now spawned detached, the
   process group is signalled on timeout (SIGKILL grace), stdio is destroyed,
   and the promise settles immediately rather than waiting for `close`.

Verification:
- `tool-confirmation.spec.tsx`: invalid args no longer auto-approve; prompt + error both render
- `handler.spec.ts`: output-cap marker fires exactly once; timeout settles and (Unix) reaps the group
- `tsc --noEmit`, biome lint/format pass
- The two `runScript` tests prefer a real Git Bash binary on Windows; the
  `/bin/sh`-keyed cases that predate this change still fail locally only
  because that path does not exist on Windows (unchanged baseline on CI).

### DIFF
--- a/.changeset/tool-execution-safety.md
+++ b/.changeset/tool-execution-safety.md
@@ -1,0 +1,16 @@
+---
+'@nanocollective/nanocoder': patch
+---
+
+Tool-execution safety fixes:
+
+- Tool approval: never auto-approve a call whose arguments fail schema
+  validation. The approval prompt now renders together with the validation
+  error, so a malformed call is never executed without explicit consent.
+- Custom tools: cap captured stdout/stderr at `BASH_MAX_OUTPUT_BYTES` while
+  streaming, with a truncation notice placed at the head of the result, so a
+  large-output tool can't exhaust memory before the final truncation runs.
+- Custom tools: on timeout, the shell is spawned detached (Unix), the whole
+  process group is signalled, and the tool call settles immediately instead of
+  waiting for a `close` event that a surviving descendant may leave un-fired —
+  killing both the infinite-hang and orphaned-process failure modes.
--- a/.changeset/tool-execution-safety.md
+++ b/.changeset/tool-execution-safety.md
@@ -7,6 +7,12 @@ Tool-execution safety fixes:
 - Tool approval: never auto-approve a call whose arguments fail schema
   validation. The approval prompt now renders together with the validation
   error, so a malformed call is never executed without explicit consent.
+  (Deliberate trade-off: a malformed approval-required call now costs the user
+  a consent prompt where it previously self-corrected without one — the price
+  of "nothing executes without consent".)
+- MCP tools: the executed handler now runs the same lenient schema type-check
+  the approval prompt shows, so a "wrong type" call is rejected locally before
+  it reaches the server instead of relying on the remote validator alone.
 - Custom tools: cap captured stdout/stderr at `BASH_MAX_OUTPUT_BYTES` while
   streaming, with a per-stream truncation notice appended to the affected
   stdout/stderr section (mirroring the bash executor), so a large-output tool
--- a/.changeset/tool-execution-safety.md
+++ b/.changeset/tool-execution-safety.md
@@ -8,8 +8,9 @@ Tool-execution safety fixes:
   validation. The approval prompt now renders together with the validation
   error, so a malformed call is never executed without explicit consent.
 - Custom tools: cap captured stdout/stderr at `BASH_MAX_OUTPUT_BYTES` while
-  streaming, with a truncation notice placed at the head of the result, so a
-  large-output tool can't exhaust memory before the final truncation runs.
+  streaming, with a per-stream truncation notice appended to the affected
+  stdout/stderr section (mirroring the bash executor), so a large-output tool
+  can't exhaust memory before the final truncation runs.
 - Custom tools: on timeout, the shell is spawned detached (Unix), the whole
   process group is signalled, and the tool call settles immediately instead of
   waiting for a `close` event that a surviving descendant may leave un-fired —
--- a/source/components/tool-confirmation.spec.tsx
+++ b/source/components/tool-confirmation.spec.tsx
@@ -1,0 +1,101 @@
+import test from 'ava';
+import {render} from 'ink-testing-library';
+import React from 'react';
+import {themes} from '../config/themes';
+import {ThemeContext} from '../hooks/useTheme';
+import {setToolManagerGetter} from '../message-handler';
+import type {ToolManager} from '../tools/tool-manager';
+import type {ToolCall} from '../types/core';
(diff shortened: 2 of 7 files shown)


will-lamerton:
Merged `main` in and reconciled the timeout path. Flagging what changed beyond a plain conflict resolution, since some of it is behavioural.

**The timeout fix collided with work `main` landed independently.** `f1e7af06` and `ca94e03a` fixed the same hang (#1141) with a different design, so this was a semantic conflict rather than a textual one. The resolution keeps the stronger half of each:

- Settling stays inside the timeout handler rather than moving to `exit`. With a trapped `SIGTERM`, neither `exit` nor `close` is prompt.
- The dead `!child.killed` guard on the `SIGKILL` escalation is gone, per `main`. Node sets that flag the moment a signal is delivered, so the escalation never ran.
- The escalation now targets the process group and deliberately outlives the shell's own exit. This is the one place the result diverges from `main`, which cleared that timer on exit: for a single process that is right, but the shell exiting does not mean its group is empty, and reaping a descendant that ignored `SIGTERM` is the point of the detached spawn. Firing at a dead group is harmless since `killProcessTree` swallows the `ESRCH`.
- Both of `main`'s timeout regression tests are kept, retargeted at `testShell` so they gate on Windows the way the rest of the suite does.

**One real bug fixed in review: a stdout flood could silently discard stderr.** The single shared byte budget meant that once stdout exhausted it, the stderr listener became a no-op and `stderrCapped` was never set, so the line explaining why a tool failed vanished with no notice that anything was missing. Verified with a script that floods stdout and then writes one error line: the line was gone and no stderr marker appeared.

The fix is per-stream budgets, which is what the built-in bash executor already did (`makeStreamCollector`: "each returned collector has its own budget, so one stream's volume can't affect the other's"). That collector is now extracted to `source/utils/stream-collector.ts` and shared by both call sites, so "mirrors the bash executor" is literally true rather than a parallel reimplementation that can drift. The `stdoutCapped`/`stderrCapped` flags and the extra `formatScriptOutput` parameter are gone, since the notice is appended inline to its own stream. New regression test covers stderr written after a stdout flood.

**Test fixes.** `tool-confirmation.spec` was building a fresh `ToolManager` mock on every `getToolManager()` call. The component feeds that object into the preview effect's dependency array, so a new reference per render re-ran the effect every render and spun React's update-depth limit, emitting eight "Maximum update depth exceeded" warnings. Production installs a stable singleton, so this was a harness artifact, but it is noise and a latent flake. Now one instance per mount. The spec also moved to `renderWithTheme` so it supplies the `TitleShapeProvider` the component started requiring on `main`, and the misindented test close in `handler.spec.ts` is fixed (biome excludes `*.spec.ts`, so CI could not catch it).

**Scope note for the changelog.** The `mcp-client.ts` change is not mentioned in the PR title or body and is not part of #1227. It is correct and closes a real gap (MCP was the last registry path bypassing `withValidation`, and `getToolJsonSchema` works because MCP native tools are built via `jsonSchema()`), so I have kept it and described it in the changeset instead of splitting it out.

Also worth noting for whoever triages: #1284 targets the same issue. This PR is the more complete of the two.

will-lamerton:
Added one commit that is unrelated to this PR's subject, flagging it explicitly so it isn't mistaken for part of the tool-execution work.

`main` is currently red on its own. The `cli-value-flags` integration suite registers a module loader that short-circuits `@/config/preferences` with a stub, and that stub exported only `loadPreferences`. Meanwhile `cli.tsx` started importing `getAlternateScreen` and `getMouseReporting` from the same module for the fullscreen/mouse-reporting work. Each change passed on its own branch; they only collide once both are on `main`, and the result is `getMouseReporting is not a function` across all 30 cases in that file.

I reproduced this on pristine `main` at `fc1ddc2b` with none of this branch's changes applied, so it is not something the merge introduced. Since #1228 cannot go green while `main` is broken, the fix rides here rather than in a separate PR. It adds the two missing exports to the stub and a comment explaining why the stub has to track every export `cli.tsx` pulls from that module.

Worth noting that this class of failure will recur: the loader short-circuits the whole specifier, so a newly imported name fails at run time as "not a function" rather than as an import error, and nothing in CI catches it until two independently-green branches meet on `main`.

will-lamerton (approved):
Reviewed, reconciled with main, and verified. Conflicts were semantic (main independently fixed the same timeout hang); resolution keeps prompt settling plus group-targeted SIGKILL escalation. Fixed a real stderr silent-drop bug found in review, the spec's React update-depth loop, the missing TitleShapeProvider, and main's pre-existing cli-value-flags breakage. All 16 checks green.

PR #1231, opened 2026-09-07
ci: call the shared release workflow

will-lamerton:
**256 lines become 32** — the last repo on a local release workflow.

The publish moves to [`Nano-Collective/.github`](https://github.com/Nano-Collective/.github/blob/main/.github/workflows/release.yml). What stays here is the two downstream jobs that are genuinely specific to this package: the **Nix flake** and the **Homebrew formula**, both local reusable workflows that need the version that just shipped.

## How the composition works

The shared workflow now exposes `should-publish`, `package-version`, `package-name` and `is-prerelease` as `workflow_call` outputs, so `update-nix` and `update-homebrew` hang off it exactly the way they hung off the local `check-version` job:

```yaml
  update-nix:
    needs: release
    if: needs.release.outputs.should-publish == 'true'
    with:
      version: ${{ needs.release.outputs.package-version }}
```

That was added for this repo — without outputs the shared workflow would have been a dead end here.

## One deliberate difference from the org default

`test-command: 'test'` rather than `test:all`. This repo's `test:all` includes the audit and semgrep passes, which are **advisory on a PR** and should not be able to block a release that has already been reviewed and merged. That was the local workflow's behaviour and it is preserved on purpose rather than quietly tightened.

## Risk

Nothing publishes unless `package.json` is ahead of the registry, so the failure mode of a mistake is a release that does not happen — loud — rather than one that ships the wrong thing. The shared workflow is already merged and running on the other five package repos.

With this, **every duplicated workflow in the collective is centralised**: `pr-checks`, `release`, `release-prepare`, `changeset-check`, `nc-review` and `update-badges`.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,5 +1,12 @@
 name: Release to NPM and GitHub
 
+# Thin caller for the shared org workflow. The publish itself lives in
+# Nano-Collective/.github; what stays here are the two downstream jobs that are
+# genuinely specific to this package — the Nix flake and the Homebrew formula.
+#
+# Publishes only when package.json's version is ahead of npm, so merging the
+# Version Packages PR is what triggers a release. Nothing else does.
+
 on:
   push:
     branches: [main]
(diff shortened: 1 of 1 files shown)


will-lamerton:
Closing this — I was wrong that nanocoder should adopt the shared release workflow, and the CI failure is what surfaced it.

knip flagged **two** unused files, not one. `scripts/extract-changelog.js` was the expected consequence of centralising (the shared workflow calls it, but knip cannot see into `Nano-Collective/.github`). **`scripts/generate-contributors.js` was not** — and chasing it showed this PR would have silently dropped three things:

| | what the local workflow does | my caller |
|---|---|---|
| `fetch-depth: 0` | full history so `generate-credits.sh` sees every contributor | shallow checkout |
| `pnpm run build:credits` | regenerates `contributors.json` **before** the build, so it ships in the package | not run |
| `generate-contributors.js` | builds a contributors section injected into the release body, crediting every PR author in the range | dropped |

That third one is the one that would have hurt. It exists precisely to credit contributors *who were not hand-thanked in a changeset* — so losing it would stop crediting people, silently, and nobody would notice for several releases.

## Why nanocoder is genuinely the exception

The other five repos adopted the shared workflow because their local copies were **drifted duplicates of the same four jobs**. nanocoder's is not a duplicate — it does strictly more, and the extra is real:

- credits generation, before build and in the release body
- a Nix flake update
- a Homebrew formula update

It also had **zero mutable action tags**, so unlike get-md and prompt-scrubber there is no Semgrep benefit to weigh against the loss. The only gain would be deduplication, and this is the one repo where the duplication is carrying its weight.

Bending the shared workflow to fit — a , a  input, some way to inject release-body content — would over-fit it to one caller and make it worse for the five repos that genuinely are the same.

**The outputs I added to the shared workflow stay.** They are generally useful, backward compatible, and cost nothing.

Five of six repos are on the shared release workflow. That is the right number, not six.

will-lamerton:
Small correction to the comment above — a couple of code spans were eaten in transit. The sentence should read:

> Bending the shared workflow to fit — a `pre-publish-command`, a `fetch-depth` input, some way to inject release-body content — would over-fit it to one caller and make it worse for the five repos that genuinely are the same.

Issue #1139, opened 2026-09-02
[Bug] notifications.ts AppleScript escape missing newlines/Unicode


akramcodez:
## Description
`escapeAppleScript` only escapes backslashes and double quotes. A newline or unescaped Unicode in the title/message terminates the string literal in `display notification "..." with title "..."`, producing a syntax error and a broken notification.

## Environment
- **OS**: macOS
- **Node version**: v22.20.0
- **Nanocoder version**: 1.30.0
- **Provider**: N/A
- **Model**: N/A

## Steps to Reproduce
1. Trigger a desktop notification containing newlines (`\n`) in the message.
2. Check if the notification appears.

## Expected Behavior
Notification appears correctly with newlines.

## Actual Behavior
AppleScript syntax error occurs, and the event is silently dropped.

## Additional Context
- [x] I have searched existing issues to ensure this is not a duplicate
- [x] I can reproduce this issue consistently
- [x] This issue occurs with the latest version of nanocoder

rishu685:
hey @akramcodez! can i take this issue?

PR #1235, opened 2026-09-09
fix(notifications): pass macOS notification arguments out-of-band to osascript (#1139)

rishu685:
## Summary
Fixes an issue where macOS native notifications were dropped or failed with AppleScript syntax errors when the notification `title` or `message` contained newlines (`\n`), carriage returns (`\r`), unescaped quotes, backslashes, or Unicode characters.

Instead of interpolating user-controlled strings into an AppleScript string literal via `escapeAppleScript`, `sendDarwin` now passes `title` and `message` out-of-band as positional process arguments (`argv`) to `osascript`:

```applescript
on run argv
  display notification (item 2 of argv) with title (item 1 of argv)
end run


will-lamerton:
Hey @rishu685 - could you address the `nc-review` points please? :)

rishu685:
done :)

will-lamerton:
Nice one @rishu685 :)

will-lamerton:
Heads up @rishu685 - your other PR #1216 (the Windows hardening) landed on main first and touched the same file, which put this branch into conflict. Since the fix itself was ready I merged main in and resolved it for you rather than making you redo it:

- `sendDarwin`: kept your out-of-band `osascript` arguments, dropped `escapeAppleScript`.
- `sendWindows`: kept main's `-EncodedCommand` + env var payload from #1216.
- `notifications.spec.ts`: kept both platform suites side by side.

One thing worth flagging: I removed the `sendNotification on win32 invokes powershell` test from this branch. It asserted the pre-#1216 invocation (`-NoProfile -Command <inline script>`), so it failed against the new `-NoProfile -NonInteractive -EncodedCommand` shape once main was merged in. #1216 already covers win32 properly, so nothing is lost. Your darwin, linux and daemon tests are all intact.

Verified locally: 22/22 in `notifications.spec.ts`, types and lint clean. Merging once checks go green. Thanks for the fix :)

will-lamerton (approved):
Verified the macOS fix end to end: `osascript -e "on run argv ... end run" <title> <message>` round-trips newlines, quotes, backslashes and emoji intact, and the `sound name "default"` clause is valid in that position. Reverting `sendDarwin` to the old `escapeAppleScript` interpolation makes the new darwin test fail, so the coverage is real rather than decorative.

All three `nc-review` points were addressed. Conflicts with #1216 resolved in the merge commit above.

PR #1243, opened 2026-09-09
feat: add Ctrl+R reverse prompt search and rebind reasoning toggle to Ctrl+G

### LINKED ISSUE #936: [Feature] Ctrl+R Reverse Prompt Search
## Description

Implement a shell-like reverse history search (`Ctrl+R`) in the input component, allowing users to quickly search and reuse previously typed prompts.

## Use Case

Users often write complex, multi-line prompts and need to run slightly modified versions of them days later. Scrolling up with the arrow keys through hundreds of prompts is incredibly tedious. `Ctrl+R` is a universal standard for terminal history search.

## Proposed Solution

- Intercept the `Ctrl+R` keypress in the Ink.js input component.
- Switch the input mode to a "bck-i-search" mode.
- As the user types, fuzzy match against the stored prompt history (from `source/prompt-history.ts`) and suggest the closest match.
- Pressing `Enter` populates the input field with the match.

## Alternatives Considered

- A `/history` slash command that prints the last 10 prompts. This is less interactive and requires copy-pasting.

## Additional Context

- This requires modifying the core input handler. Care must be taken to ensure it doesn't conflict with OS-level terminal shortcuts.

akramcodez:
## Description

Closes #936

This PR implements standard shell-like **Ctrl+R reverse prompt search** (`(reverse-i-search)`) directly within the terminal input component, allowing users to quickly search, cycle through, and reuse previously entered prompts.

### Approach & Implementation Details
Rather than introducing heavy third-party fuzzy-search dependencies or rewriting `PromptHistory`, this implementation takes a lean, dependable approach true to standard terminal shells:
- **Reuse Existing History Storage:** Works seamlessly with the existing `PromptHistory` system without modifying data structures or the 100-entry limit.
- **Deterministic Reverse Substring Matching:** Performs case-insensitive reverse search (`findReverseMatch`) from newest to oldest entries.
- **Full State Preservation:** Searches display text while preserving full `InputState` (`displayValue` and `placeholderContent`) so multi-line prompt attachments and metadata are not lost when selecting or restoring.
- **Shortcut Conflict Resolution:** Rebinds the reasoning trace expansion toggle from `Ctrl+R` to `Ctrl+G`, keeping `Ctrl+T` dedicated to the live task list.
- **Interactive Shell UX:**
  - `Ctrl+R`: Enters reverse search mode displaying `(reverse-i-search)`.
  - Type / Backspace: Live updates search query and suggests matching prompt. If no match exists, displays `(failed reverse-i-search)` in error color.
  - Repeated `Ctrl+R`: Cycles backward to older matches, cleanly stopping at the oldest match without disorienting wrap-around.
  - `Enter`: Accepts and commits the selected match into the input buffer.
  - `Esc` or `Ctrl+C`: Cancels search and restores the original draft.
- **Documentation & UI Hints Updated:** Updated shortcut hints across all docs (`keyboard-shortcuts.md`, `preferences.md`, `commands.md`, `index.md`, `chatgpt-codex.md`) and UI components (`assistant-reasoning.tsx`, `streaming-reasoning.tsx`, `settings-reasoning-traces.tsx`, `constants.ts`).

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Documentation update

## Changeset

### Automated Tests

*Added comprehensive unit tests in `source/components/user-input.spec.tsx` covering search mode activation, query filtering, cycling older matches, commit on Enter, cancel on Escape/Ctrl+C, backspace handling, and failed search display.*

### Manual Testing

- [ ] Tested with Ollama
- [ ] Tested with OpenRouter

## Checklist

- [x] Self-review completed


will-lamerton (changes requested):
Nice work on this, the core is solid. `findReverseMatch` is a clean pure helper, draft restore via `reverseSearchDraftRef` is correct, the failed-search and stop-at-oldest semantics match bash, and the Ctrl+R to Ctrl+G rebind is applied consistently across the UI and docs. Four things to fix before merge.

**1. Needs a rebase, and it is not mechanical.** `e791b838 feat(tui): add a ? keyboard shortcuts overlay` landed on main and conflicts in `user-input.tsx` and `constants.ts`. Two real issues fall out of it:

- Main's `SHORTCUTS` table still has `['Ctrl+R', 'Toggle reasoning traces']`. That needs to become Ctrl+G, plus a new Ctrl+R reverse-search row, or the in-app overlay teaches the wrong binding.
- Main's `?` handler is now the first check in `useInput` and fires whenever the input is empty. The input *is* empty during reverse search, so after rebasing, typing `?` into the query opens the overlay instead. Order the reverse-search branch above the `?` check, or bail out of the `?` check when `isReverseSearchMode`.

**2. The new tests wipe the developer's real prompt history file.** `test.beforeEach(() => promptHistory.clearHistory())` runs against the singleton, which resolves to the user's real `.nano-coder-history` in the platform config dir. I ran the spec and it truncated mine to 0 bytes. Two knock-on problems: AVA `beforeEach` applies to every test in the file regardless of declaration position, so all 62 tests now do real disk I/O, and the tests also seed junk into the live file via `saveHistory()`. Please point the spec at a temp file with `new PromptHistory(tmpPath)`, or set `NANOCODER_CONFIG_DIR` for it, instead of touching the singleton.

**3. `clearHistory()` is production API added only for tests.** It has no product caller and writes to disk. Once #2 is fixed it should be dropped.

**4. Ctrl+C quits nanocoder, it does not cancel the search.** `App.tsx` registers an unconditional `useInput` that calls `handleExit()` on Ctrl+C, going straight to `gracefulShutdown(0)` with no confirmation. Ink dispatches to every mounted handler, so the search-mode branch runs and the app exits anyway. The test only passes because it renders `UserInput` without `App`, so it is giving false confidence. Main's own overlay lists `Ctrl+C` as `Exit`. Either drop the Ctrl+C branch, its test, and the claim in the description, or gate App's handler on search mode.

---

Not blocking:

- The red CI check is not yours. `pr-checks / Unit Tests` failed on `source/commands/stats.spec.tsx` "StatsDisplay changes range with arrow keys and closes on Escape" timing out at 2000ms. That file is untouched here and passes locally on your branch. All the reverse-search tests passed in CI. A rerun after rebase should clear it.
- The two earlier bot nits are still open: the redundant `inputChar.length >= 1`, and the docs row not mentioning Enter/Esc.
- Two minor divergences from bash, both fine and both encoded in your tests, just flagging: backspace at an empty query is a no-op where bash exits search, and each keystroke restarts the scan from the newest entry rather than continuing from the current position.

PR #1252, opened 2026-09-10
fix(path): allow containment when workspace root is a filesystem root

### LINKED ISSUE #1240: [Bug] isPathInside() fails path containment when project root is filesystem or drive root (/ or C:\)
## Description

In `source/utils/path-validation.ts`, `isPathInside(target, root)` checks whether `target` is lexically within `root`:
```typescript
export function isPathInside(target: string, root: string): boolean {
    const normalizedRoot = path.resolve(root);
    const normalizedTarget = path.resolve(target);
    return (
        normalizedTarget === normalizedRoot ||
        normalizedTarget.startsWith(normalizedRoot + path.sep)
    );
}
```

When `root` is the filesystem root (`/` on POSIX or `C:\` on Windows):
- `path.resolve('/')` returns `'/'`.
- `normalizedRoot + path.sep` produces `'/' + '/'` = `'//'`.
- `normalizedTarget.startsWith('//')` evaluates to `false` for every standard path (e.g. `'/home/user/project'`).
- On Windows with `C:\`, `normalizedRoot + path.sep` produces `'C:\\\'`.

When nanocoder is run inside container environments where `/` is the project root, or in a root directory of a drive, all file operations (`read_file`, `write_file`, `search_file_contents`) fail with:
`File path escapes project directory: ... -> ...`

## Environment

- **OS**: Linux (Docker / container root) / Windows
- **Node version**: v22.20.0
- **Provider**: Any
- **Model**: Any

## Steps to Reproduce

1. In Node REPL or a script:
   ```typescript
   import {isPathInside} from './source/utils/path-validation';
   console.log(isPathInside('/home/user', '/')); // prints false!
   ```
2. Start nanocoder with root directory as workspace: `nanocoder run "read_file package.json"`.
3. Path containment rejects any valid file under the root.

## Expected Behavior

`isPathInside('/home/user', '/')` should return `true`.

## Actual Behavior

Returns `false` because of the duplicated path separator (`//`).

## Logs/Screenshots

```text
Error: File path escapes project directory: package.json -> /package.json
```

## Additional Context

89799969:
## Problem

isPathInside (and the absolute-path branch of isValidFilePath) builds the containment prefix as 
ormalizedRoot + path.sep.

When the workspace root is a filesystem root (/ or C:\), that becomes // or C:\\. No real path starts with those sequences, so every descendant is rejected and file tools fail with `File path escapes project directory`. This hits containers that use / as the project root and Windows drive-root workspaces. See #1240.

## Fix

Reuse isPathInside from isValidFilePath and skip the extra separator when the resolved root already ends with one:

`	s
const prefix = normalizedRoot.endsWith(path.sep)
  ? normalizedRoot
  : normalizedRoot + path.sep;
return normalizedTarget.startsWith(prefix);
`

Sibling-prefix protection (/proj vs /proj-evil) is unchanged for normal directories.

## Tests

Added `isPathInside: filesystem root contains descendants without a doubled separator` in source/utils/path-validation.spec.ts.

`
pnpm run test:ava source/utils/path-validation.spec.ts
`

All isPathInside / isRealPathInside cases pass on Windows. Remaining suite failures on this machine are pre-existing POSIX-path fixtures (/home/user/project) unrelated to this change.

Fixes #1240

### DIFF
--- a/.changeset/allow-filesystem-root-workspaces.md
+++ b/.changeset/allow-filesystem-root-workspaces.md
@@ -1,0 +1,5 @@
+---
+"@nanocollective/nanocoder": patch
+---
+
+Allow projects located at a filesystem root, including `/` and Windows drive roots, to pass path-containment checks. Closes #1240.
--- a/source/utils/path-validation.spec.ts
+++ b/source/utils/path-validation.spec.ts
@@ -6,7 +6,7 @@ import {
 	writeFileSync,
 } from 'node:fs';
 import {tmpdir} from 'node:os';
-import {join} from 'node:path';
+import {join, parse as parsePath} from 'node:path';
 import test from 'ava';
 import {
 	isPathInside,
@@ -317,6 +317,16 @@ test('isPathInside: rejects siblings, ancestors, and shared prefixes', t => {
 test('isPathInside: resolves ../ traversal before comparing', t => {
 	t.false(isPathInside('/proj/../etc', '/proj'));
 	t.true(isPathInside('/proj/src/../lib', '/proj'));
+});
+
+test('isPathInside: filesystem root contains descendants without a doubled separator', t => {
+	// POSIX `/` and Windows `C:\` both already end with a separator. Appending
+	// path.sep would produce `//` or `C:\\`, which no real path starts with.
+	const root = parsePath(process.cwd()).root;
+	const child = join(root, 'home', 'user', 'project');
+	t.true(isPathInside(root, root));
+	t.true(isPathInside(child, root));
+	t.true(isValidFilePath(child, root));
 });
 
 test('isPathInside: is lexical, so an escaping symlink still passes', t => {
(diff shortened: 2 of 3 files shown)


will-lamerton:
Thanks @89799969 - the core fix is right, and I verified the new `isPathInside` test fails against `main` and passes here.

I pushed one follow-up commit to your branch: `search-file-contents.tsx` was rebuilding the containment prefix by hand as `root + path.sep`, so it carried the exact same doubled-separator bug. Since #1240 lists `search_file_contents` as one of the failing tools, the issue wasn't fully closed without it. It now delegates to `isPathInside`, with a test that pins the project root to the filesystem root.

/re-review

will-lamerton (approved):
Verified locally: both regression tests fail against main and pass here, tsc and biome clean. Merging.

PR #1253, opened 2026-09-10
fix(tools): hide dotfiles when list_directory defaults to .

### LINKED ISSUE #1237: [Bug] list_directory leaks all hidden files/dotfiles by default due to !dirPath.startsWith('.')
## Description

When calling the `list_directory` tool without specifying a `path` parameter, `dirPath` defaults to `'.'`. Inside `walkDirectory()`, dotfiles are meant to be skipped unless `showHiddenFiles` is `true`. However, the skip condition contains `!dirPath.startsWith('.')`. Because `'.'.startsWith('.')` is `true`, `!dirPath.startsWith('.')` evaluates to `false`, which bypasses the filter and leaks all hidden files and directories (`.env`, `.git`, `.vscode`, `.DS_Store`, etc.) into the tool output.

## Environment

- **Node version**: v22.20.0
- **Provider**: Any
- **Model**: Any

## Steps to Reproduce

1. Have hidden files in the project root directory (e.g. `.env`, `.config`, `.nanocoder-history`).
2. Run `list_directory` with default arguments: `{}` or `{ path: '.' }` or `{ path: './src' }`.
3. Check the returned directory listing.

## Expected Behavior

With `showHiddenFiles: false` (the default), hidden files and directories should be excluded from directory listings.

## Actual Behavior

Every dotfile and hidden directory in the directory is printed in the listing. Even in `source/tools/list-directory.spec.tsx`, line 556 had to work around this bug:
```typescript
// List a subdirectory explicitly (not the current dir) to hide dotfiles
```

## Logs/Screenshots

In `source/tools/list-directory.tsx`:
```typescript
const dirPath = args.path || '.';
...
// Skip hidden files unless showHiddenFiles is true
if (
    !showHiddenFiles &&
    item.name.startsWith('.') &&
    !dirPath.startsWith('.')
) {
    continue;
}
```

## Additional Context

- N/A

89799969:
## Problem

list_directory skips hidden entries with:

`	s
if (!showHiddenFiles && item.name.startsWith('.') && !dirPath.startsWith('.')) {
  continue;
}
`

The default dirPath is '.', and '.'.startsWith('.') is true, so the filter is disabled. Listing the project root (or ./) leaks `.env`, `.git`, `.vscode`, etc. even when `showHiddenFiles` is false. The existing default-hidden test only listed an explicit `subdir` and missed this. See #1237.

## Fix

Treat a path as a hidden-directory request only when a path segment actually starts with a non-. character (e.g. `.github`, `./.config`) — not when it is the project root (`.` / `./`):

`	s
const listingHiddenDir = /(^|[/\\])\.[^./\\]/.test(dirPath);
`

Explicit `showHiddenFiles: true` and listing a real hidden directory still work as before.

## Tests

- New: `list_directory hides dotfiles when listing the project root (.)` covering `undefined`, `.`, and `./`.
- Full list-directory.spec.tsx: **31 passed**.

`
pnpm run test:ava source/tools/list-directory.spec.tsx
`

Fixes #1237

### DIFF
--- a/.changeset/hide-default-list-directory-dotfiles.md
+++ b/.changeset/hide-default-list-directory-dotfiles.md
@@ -1,0 +1,5 @@
+---
+"@nanocollective/nanocoder": patch
+---
+
+Hide dotfiles when `list_directory` uses its default project-root path. Closes #1237.
--- a/source/tools/list-directory.spec.tsx
+++ b/source/tools/list-directory.spec.tsx
@@ -571,6 +571,34 @@ test.serial('list_directory hides dotfiles by default', async t => {
 	}
 });
 
+test.serial('list_directory hides dotfiles when listing the project root (.)', async t => {
+	t.timeout(10000);
+	const originalCwd = process.cwd();
+
+	try {
+		const testDir = join(process.cwd(), 'test-listdir-hidden-root-temp');
+		mkdirSync(testDir, {recursive: true});
+		writeFileSync(join(testDir, '.hidden'), 'secret');
+		writeFileSync(join(testDir, 'visible.ts'), 'content');
+		writeFileSync(join(testDir, '.env'), 'API_KEY=leak');
+
+		process.chdir(testDir);
+
+		for (const path of [undefined, '.', './']) {
+			const result = await listDirectoryTool.tool.execute!(
+				path === undefined ? {} : {path},
+				{toolCallId: 'test', messages: []},
+			);
+			t.false(result.includes('.hidden'), `path=${path} should hide .hidden`);
+			t.false(result.includes('.env'), `path=${path} should hide .env`);
+			t.true(result.includes('visible.ts'), `path=${path} should list visible.ts`);
+		}
+	} finally {
+		process.chdir(originalCwd);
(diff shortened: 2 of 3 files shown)


will-lamerton:
Thanks for this - the diagnosis is exactly right and the regression test is a good one (I verified it fails against `main` and passes with the fix).

I pushed one follow-up commit (04b41cb2) rather than block on a review round. The escape hatch was testing the raw `path` argument, and `isValidFilePath` accepts absolute paths inside the project root, so a project that itself lives under a hidden ancestor - `~/.local/share/proj`, a CI cache dir - would match the regex and leak every dotfile. That is the same class of leak this PR fixes for the default `.`.

The check now runs against `relative(root, resolvedPath)`, which is `''` at the project root, `.github` for a genuinely hidden directory, and drops hidden ancestors entirely. I also hoisted it out of the per-entry loop since it only depends on the requested path. Added a test covering the hidden-ancestor case.

32 tests passing in `list-directory.spec.tsx`. Merging once CI is green.

will-lamerton (approved):
Verified locally: the bug reproduces on main, the new tests fail without the fix, tsc and biome are clean, 32 tests pass. Follow-up commit tightens the hidden-dir check to the project-relative path.

PR #1255, opened 2026-09-10
fix(schedule): reference-count shared cron subscriptions

### LINKED ISSUE #1239: [Bug] ScheduleEventSource.unregister() stops shared cron jobs due to lack of reference counting
## Description

In `source/events/sources/schedule.ts`, `ScheduleEventSource` manages cron subscriptions. As documented in `register(expression: string)`:
*"Subsequent calls with the same expression are no-ops, so multiple subscriptions sharing a cron expression only create one underlying job. The router fans the resulting event out to all of them."*

However, `unregister(expression: string)` immediately stops the underlying job and removes it from `this.jobs`:
```typescript
unregister(expression: string): void {
    const job = this.jobs.get(expression);
    if (!job) return;
    job.stop();
    this.jobs.delete(expression);
}
```
If two skills or listeners register the same cron expression (e.g., `0 9 * * MON`), and one subscription is cancelled or unregistered, `job.stop()` is called. The remaining subscription is silently orphaned and will never receive scheduled ticks again.

## Environment

- **Node version**: v22.20.0
- **Provider**: Any
- **Model**: Any

## Steps to Reproduce

1. Register two subscriptions with the same cron expression:
   ```typescript
   source.register('0 9 * * MON'); // Sub 1
   source.register('0 9 * * MON'); // Sub 2
   ```
2. Unregister subscription 1:
   ```typescript
   source.unregister('0 9 * * MON');
   ```
3. Check `source.listRegistered()`: it returns `[]` and the cron job is stopped.
4. Trigger the scheduled tick: Subscription 2 never receives the event.

## Expected Behavior

Shared cron jobs should be reference-counted. A job should only be stopped and deleted when all subscriptions registered for that expression have been unregistered.

## Actual Behavior

The first unregister call stops and deletes the job, leaving all other active subscriptions for that schedule silent.

## Logs/Screenshots

In `source/events/sources/schedule.ts`:
```typescript
register(expression: string): void {
    if (this.jobs.has(expression)) return;
    const job = this.factory(expression, () => { ... });
    this.jobs.set(expression, job);
}

unregister(expression: string): void {
    const job = this.jobs.get(expression);
    if (!job) return;
    job.stop();
    this.jobs.delete(expression);
}
```

## Additional Context

Add a `refCounts = new Map<string, number>()` in `ScheduleEventSource`. Increment on `register()`, decrement on `unregister()`, and only invoke `job.stop()` when the count reaches zero.

89799969:
## Problem

`ScheduleEventSource` documents that multiple subscriptions may share one cron expression, and that one underlying job fans ticks out to all of them. `register()` correctly treated a duplicate expression as a no-op for job creation, but `unregister()` stopped and deleted the job on the first call.

If two skills registered `0 9 * * MON` and one later unregistered, the remaining subscription was silently orphaned and never received ticks again.

## Fix

Reference-count each expression:

- `register()` increments the count and creates the job only on the first subscription
- `unregister()` decrements and stops the job only when the last subscriber leaves
- `stop()` still tears everything down and clears the counts

## Tests

- Existing `source/events/sources/schedule.spec.ts` suite still passes
- New regression test `shared expression stays live until the last unregister` covers: one job for two registers, job stays listed after the first unregister, tick still dispatches to both subscribers, last unregister stops the job

```text
8 tests passed
```

`tsc --noEmit` passes.

Fixes #1239

### DIFF
--- a/.changeset/schedule-shared-cron-refcount.md
+++ b/.changeset/schedule-shared-cron-refcount.md
@@ -1,0 +1,5 @@
+---
+"@nanocollective/nanocoder": patch
+---
+
+Keep shared cron jobs alive until every subscriber unregisters. `ScheduleEventSource.unregister()` used to stop the underlying job on the first call even when other subscriptions still used the same expression. Closes #1239.
--- a/source/events/sources/schedule.spec.ts
+++ b/source/events/sources/schedule.spec.ts
@@ -110,6 +110,30 @@ test('unregister stops the job and removes from the map', t => {
 	t.deepEqual(stub.stops, ['0 9 * * MON']);
 });
 
+test('shared expression stays live until the last unregister', async t => {
+	const {router, events} = captureRouter();
+	const stub = stubFactory();
+	const source = new ScheduleEventSource(router, stub.factory);
+
+	router.subscribe(cronSub('a', '0 9 * * MON'));
+	router.subscribe(cronSub('b', '0 9 * * MON'));
+	source.register('0 9 * * MON');
+	source.register('0 9 * * MON');
+	t.is(stub.jobs.length, 1);
+
+	source.unregister('0 9 * * MON');
+	t.deepEqual(source.listRegistered(), ['0 9 * * MON']);
+	t.deepEqual(stub.stops, []);
+
+	stub.jobs[0]?.onTick();
+	await new Promise(r => setImmediate(r));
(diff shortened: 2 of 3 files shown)


will-lamerton:
Thanks for this PR @89799969  - feel free to add yourself as a contributor to our website via a PR which I will approve :)

PR #1256, opened 2026-09-10
fix(config): preserve MCP source through loadAppConfig unwrap

### LINKED ISSUE #1248: [Bug] `validateProjectConfigSecurity` never runs, hardcoded-credential scanner is dead code
## Description

`validateMCPConfigSecurity` is the app's only defense-in-depth signal for
"you committed a hardcoded credential into a project-level MCP config." It is
only invoked, via `validateProjectConfigSecurity`, on servers whose
`.source === 'project'`. But the object type that field lives on
(`MCPServerWithSource`) is stripped down to the plain `MCPServerConfig`
(which never had that field) before it reaches this call, so the filter is
always empty and the check never fires in the shipped app.

Root cause is an integration break between two refactors: `mcp-config-loader.ts`
tracks provenance on a wrapper type for its own merge logic, but the consumer
(`config/index.ts`) unwraps to the inner type before handing it to the
validator, silently discarding the field the validator's filter depends on.

The unit test for this validator (`validation.spec.ts`) passes only because
it hand-constructs an object with `.source` set directly on the inner type 
a shape that never occurs in the real data flow which is exactly why this
break went unnoticed.

## Location

- `source/config/mcp-config-loader.ts:47-63` (`mapServerConfig` never copies
  `.source` onto the returned `MCPServerConfig`)
- `source/config/index.ts:704-705` (`loadAllMCPConfigs().map(item => item.server)`
  strips the wrapper's `.source` field)
- `source/hooks/useAppInitialization.tsx:210` (`validateProjectConfigSecurity(config.mcpServers)`
  called on the already-stripped array)
- `source/config/validation.ts:58-60` (filters `server.source === 'project'`,
  always `undefined` on real objects)

## Environment

- **OS**: Any
- **Node version**: v22.22.3
- **Provider**: N/A
- **Model**: N/A

## Steps to Reproduce

1. Add an MCP server entry to a project's `agents.config.json` with a
   hardcoded-looking secret, e.g. `"env": {"API_KEY": "sk-abc123..."}`.
2. Start `nanocoder` in that project.
3. Observe that no warning is logged, even though `validateMCPConfigSecurity`
   is designed to detect exactly this pattern.

Alternatively, add a `console.log` (or breakpoint) inside
`validateMCPConfigSecurity` and confirm it is never reached for a real
project config, because the array passed to `validateProjectConfigSecurity`
contains objects with `source: undefined`.

## Expected Behavior

A project-level MCP server config containing an apparent hardcoded credential
should trigger `validateMCPConfigSecurity`'s warning.

## Actual Behavior

The warning never fires for any real config, project-level or otherwise,
because the `.source` field its filter depends on is stripped before the
validator ever sees the data.

## Additional Context

      `mapServerConfig` and the `.map(item => item.server)` unwrap are
      directly verifiable as never setting/preserving `.source`)

**Suggested fix**: preserve `.source` through the unwrap in `config/index.ts`
(don't discard the wrapper before validation), or have
`validateProjectConfigSecurity` accept the `MCPServerWithSource[]` array
directly instead of the stripped `MCPServerConfig[]`.

89799969:
## Problem

`validateProjectConfigSecurity` is the app's defense-in-depth check for hardcoded credentials in **project-level** MCP configs. It filters on `MCPServerConfig.source === 'project'`, but `loadAppConfig` unwrapped `MCPServerWithSource` to plain `MCPServerConfig` before validation:

```ts
const mcpServers = mcpServersWithSource.map(item => item.server);
```

The loader only tracks provenance on the wrapper, so every runtime object had `source: undefined`, the filter was always empty, and the scanner never ran. The unit test passed only because it hand-set `.source` on the inner type — a shape that never occurs in the real data flow.

## Fix

When unwrapping, copy the wrapper's `source` onto the runtime object (`MCPServerConfig.source` is already an optional field):

```ts
const mcpServers = mcpServersWithSource.map(item => ({
  ...item.server,
  source: item.source,
}));
```

Both `useAppInitialization` and `plain/initialize.ts` then see real provenance.

## Tests

- Existing `source/config/validation.spec.ts` suite still passes
- New regression test `loader unwrap keeps source so project configs reach the validator` models the real `MCPServerWithSource[]` shape and asserts:
  - the production-style unwrap keeps only `project-server` in the project filter
  - the old strip-wrapper path yields an empty filter (the bug)
  - `validateProjectConfigSecurity` accepts the preserved-source array

```text
5 tests passed
```

`tsc --noEmit` passes.

Fixes #1248

### DIFF
--- a/.changeset/mcp-project-config-security-source.md
+++ b/.changeset/mcp-project-config-security-source.md
@@ -1,0 +1,5 @@
+---
+"@nanocollective/nanocoder": patch
+---
+
+Wire project-level MCP security validation to real configs again. `loadAppConfig` dropped the loader wrapper's `source` field before `validateProjectConfigSecurity` filtered on it, so hardcoded-credential warnings never fired. Closes #1248.
--- a/source/config/index.ts
+++ b/source/config/index.ts
@@ -702,7 +702,12 @@ function loadAppConfig(): AppConfig {
 
 	// Load MCP servers from the new hierarchical configuration system
 	const mcpServersWithSource = loadAllMCPConfigs();
-	const mcpServers = mcpServersWithSource.map(item => item.server);
+	// Keep provenance on the runtime objects: validateProjectConfigSecurity
+	// filters on MCPServerConfig.source, which the loader only tracks on the wrapper.
+	const mcpServers = mcpServersWithSource.map(item => ({
+		...item.server,
+		source: item.source,
+	}));
 
 	// Load auto-compact configuration
 	const autoCompact = loadAutoCompactConfig();
--- a/source/config/validation.spec.ts
+++ b/source/config/validation.spec.ts
@@ -101,3 +101,46 @@ test('validateProjectConfigSecurity - only validates project-level configs', t =
 		validateProjectConfigSecurity(mcpServers);
 	});
 });
+
+test('loader unwrap keeps source so project configs reach the validator', t => {
+	// Mirrors MCPServerWithSource from mcp-config-loader: provenance lives on the wrapper.
+	const wrapped = [
+		{
+			server: {
+				name: 'project-server',
+				transport: 'stdio' as const,
+				command: 'npx',
+				env: {API_KEY: 'hardcoded-key'},
+			},
+			source: 'project' as const,
+		},
+		{
+			server: {
+				name: 'global-server',
+				transport: 'stdio' as const,
+				command: 'npx',
+				env: {API_KEY: 'hardcoded-key'},
+			},
+			source: 'global' as const,
+		},
+	];
+
+	// Production path: loadAppConfig must copy wrapper.source onto the runtime object.
+	const mcpServers = wrapped.map(item => ({
+		...item.server,
+		source: item.source,
+	}));
+
+	t.deepEqual(
+		mcpServers.filter(server => server.source === 'project').map(s => s.name),
+		['project-server'],
+	);
+
+	// Stripping the wrapper without copying source is the regression: the filter is empty.
+	const stripped = wrapped.map(item => item.server);
+	t.is(stripped.filter(server => server.source === 'project').length, 0);
+
+	t.notThrows(() => {
+		validateProjectConfigSecurity(mcpServers);
+	});
+});


will-lamerton (changes requested):
Thanks for this - the root cause analysis is right, the fix in `source/config/index.ts` is the correct minimal change, and both consumers (`useAppInitialization.tsx:210`, `plain/initialize.ts:163`) are covered by it. Changeset is correct too.

Two things I'd like sorted before merge.

### 1. Re-enabling the scanner emits false positives for correct usage

`loadProjectMCPConfig` runs `substituteEnvVars(mcpServers)` at `source/config/mcp-config-loader.ts:81`, before `mapServerConfig`. So by the time `validateMCPConfigSecurity` sees a server, `"$API_KEY"` has already been resolved to its literal value, and the `!value.startsWith('$')` check at `source/config/validation.ts:17` is always true.

I ran the real path with a temp project `.mcp.json`:

```
--- loaded env values after substitution ---
 good-citizen: source=project API_KEY="sk-live-secret-from-env"   # was "$MY_REAL_KEY"
 unset-var:    source=project API_KEY=""                          # was "$NOT_SET_ANYWHERE"
 actually-bad: source=project API_KEY="sk-hardcoded-literal"

--- security warnings captured ---
 * good-citizen
 * unset-var
 * actually-bad
total: 3
```

Only `actually-bad` is a genuine finding. The other two are users doing the right thing, and the warning tells them to "Consider using environment variable references (e.g. `$API_KEY`)", which is exactly what they wrote. Every project with an env key matching token/key/secret/password/auth would get a security warning on startup.

This is likely why the dead filter went unnoticed: the check has never run against real data. Fix is to validate the pre-substitution values, either by scanning inside `loadProjectMCPConfig` before `substituteEnvVars`, or by carrying the raw `env`/`headers` through for the validator.

### 2. The regression test can't catch the regression

The new test builds `mcpServers` by reimplementing the unwrap inline; it never calls anything from `source/config/index.ts`. I checked out this branch, reverted `source/config/index.ts` to main, and re-ran:

```
✔ loader unwrap keeps source so project configs reach the validator
5 tests passed
```

Green with the production fix entirely absent. `source/config/index.spec.ts` already has the harness for a real one (temp dir + `reloadAppConfig`): write a `.mcp.json` into a temp cwd, call `reloadAppConfig()`, assert `config.mcpServers[0].source === 'project'`. That also covers nc-review's first nit, since asserting the warning actually fires is what surfaces issue 1 above.

nc-review's second nit (`mapServerConfig` still drops `.source`) is fair but not blocking - `effective-config.ts:801` is the only other caller and it destructures `source` off the wrapper itself.

Also note the required `pr-checks` haven't run on this commit yet, so lint/types/tests/build are still unverified here. I'll get that approved.

89799969:
Pushed `f4a25ba` addressing both review items.

**1. False positives from env substitution**
`loadProjectMCPConfig` / `loadGlobalMCPConfig` / `loadEnvMCPConfigs` now map first, snapshot pre-substitution `env`/`headers` onto `rawEnv`/`rawHeaders`, then substitute for runtime. `collectMCPSecurityFindings` prefers the raw copies, so `$MY_REAL_KEY` stays silent and only a literal secret warns. `unset-var` (empty after substitution of an unknown `$VAR`) also no longer warns.

**2. Regression test that can fail**
Replaced the inline unwrap simulation with a serial end-to-end test that writes a temp `.mcp.json` (good-citizen / unset-var / actually-bad), calls `reloadAppConfig()`, and asserts:
- every runtime server keeps `source: 'project'` (fails if the unwrap in `source/config/index.ts` is reverted)
- runtime `env.API_KEY` is substituted, while `rawEnv.API_KEY` keeps `$MY_REAL_KEY`
- `collectMCPSecurityFindings` reports only `actually-bad`

Also exported `collectMCPSecurityFindings` so the scanner is testable without monkey-patching `logWarning`.

Verified: validation + loader specs pass; `tsc --noEmit` clean; Biome clean on the touched files.

89799969:
CI note on `2fd2104`:

- All **7721** unit tests that ran include the new validation/loader coverage.
- The only red check is the pre-existing flaky UI test `StatsDisplay changes range with arrow keys and closes on Escape` (`Timed out after 2000ms`), which is unrelated to this change (`source/commands/stats`).
- `Config Schema Freshness` is green again after regenerating `agents.config.schema.json`.
- `mcp-config-loader.ts` coverage rose 88.88% → 92.06% on the previous run; validation.ts remains 100%.

I cannot re-run the workflow without admin rights. Please re-run `Unit Tests & Coverage Analysis` when convenient, or let me know if you want me to take a look at that stats flake separately.

PR #1259, opened 2026-09-11
fix(security): gate daemon start on directory-trust disclaimer

addyCooks:
## Description
Closes a critical security vulnerability where `nanocoder daemon start` could execute untrusted code without user confirmation. The daemon fast path went straight to `runDaemonCli` → `startDaemon` → `bootSkillPipeline` without checking directory trust, loading every `.nanocoder/agents|commands|tools/*.md` and `skills/<name>/skill.yaml` unconditionally.

**Security Impact:**
1. **Unattended code execution**: The daemon loads skills in headless mode with NO confirmation prompts for `execute_bash`, `write_file`, `string_replace`, `diff_edit`, and MCP tools
2. **Attack vector**: A malicious repo with `.nanocoder/skills/evil/skill.yaml` containing a `file.changed` subscription could execute arbitrary commands when ANY file is touched
3. **Bypass of all safeguards**: Headless mode is strictly more permissive than auto-accept - even bash execution runs without prompts

**Fix:** `daemon start` now checks `preferences.trustedDirectories` before spawning the daemon process and refuses with exit code 1 if the directory isn't trusted. The trust-check logic was refactored into a shared `ensureDirectoryTrust()` helper in `@/config/preferences` (eliminating code duplication from plain/shell.ts). Users must either:
- Run `nanocoder` interactively once to accept the trust disclaimer, OR
- Pass `--trust-directory` flag (one-shot bypass), OR
- Set `NANOCODER_TRUST_DIRECTORY=1` (persists trust)

## Type of Change
- [x] Bug fix (security vulnerability)
- [ ] New feature
- [x] Breaking change (security-critical fix - previously unprotected daemon now requires trust)
- [ ] Documentation update

## Changeset
- [x] Added a changeset (`pnpm changeset`) describing this change for the changelog

### Automated Tests
- [x] New features include passing tests in `.spec.ts/tsx` files
- [x] All existing tests pass (`pnpm test:all` completes successfully)
- [x] Tests cover both success and error scenarios

### Manual Testing
- [ ] Tested with Ollama
- [ ] Tested with OpenRouter
- [ ] Tested with OpenAI-compatible API
- [x] Tested MCP integration (if applicable)

## Checklist
- [x] If this was for an open issue, I was assigned to it (Issue #1245)
- [x] Code follows project style guidelines (Biome formatting and linting passed)
- [x] Self-review completed
- [x] Documentation updated (if needed) - Updated CLI help text for `--trust-directory` flag
[...]

PR #1260, opened 2026-09-11
fix(daemon): require directory trust before booting

### LINKED ISSUE #1245: [Bug] `nanocoder daemon start` has no directory-trust check; headless mode permits unattended command execution
## Description

`nanocoder daemon start` is a dedicated CLI fast path (`cli.tsx`) that goes
straight to `runDaemonCli` → `startDaemon` → `bootSkillPipeline`, loading and
registering every `.nanocoder/agents|commands|tools/*.md` and
`.nanocoder/skills/<name>/skill.yaml` in the project root. `source/daemon/`
has zero references to `trust`/`Trust` anywhere this command never touches
`App.tsx` or `useDirectoryTrust`, the app's only trust-check implementation.

Once booted, `headless` mode (used for every daemon-triggered run) excludes
only `ask_user`/`agent`/`write_plan` from the tool list `execute_bash`,
`write_file`, `string_replace`, `diff_edit`, and MCP tools all remain
available and are explicitly coded to skip confirmation in `headless`
(`execute-bash.tsx`'s own comment: *"Even auto-accept still prompts for
bash... in headless mode"* — i.e. headless is strictly more permissive than
auto-accept for bash). `execute_bash`'s denylist only blocks a handful of
literal catastrophic patterns and is trivially bypassed. OS-level sandboxing
is opt-in and explicitly unsupported on Windows.

## Location

- `source/cli.tsx:48-73` (daemon fast path never imports `App.tsx`/`useDirectoryTrust`)
- `source/daemon/cli.ts`, `source/daemon/daemon.ts`, `source/daemon/entry.ts`
- `source/skills/bootstrap.ts` (`bootSkillPipeline`)
- `source/tools/tool-manager.ts:38-58` (`MODE_EXCLUDED_TOOLS.headless`)
- `source/tools/execute-bash.tsx:168-182` (`if (mode === 'headless') return false;`)
- `source/utils/tool-approval.ts:13-20`
- `source/services/bash-sandbox.ts:136-140` (unsupported on Windows)

## Environment

- **OS**: Any (Windows has no sandbox fallback at all)
- **Node version**: v22.22.3
- **Provider**: N/A
- **Model**: N/A

## Steps to Reproduce

1. Commit `.nanocoder/skills/evil/skill.yaml` to a repo with
   `subscribe: [{kind: file.changed, paths: ["**/*"]}]` and a subagent whose
   `systemPrompt` drives `execute_bash` with a payload.
2. Clone the repo and run `nanocoder daemon start` in it (a documented,
   encouraged command) — no trust prompt appears anywhere in this path.
3. Touch any file (including a routine, unrelated edit) `file.changed`
   fires, dispatches the subagent in `headless` mode, and `execute_bash` runs
   the payload with no confirmation.

## Expected Behavior

`daemon start` should refuse to arm subscriptions / load skill content from a
project directory that has not been explicitly trusted, matching the
protection the interactive TUI is supposed to provide (see also F-21/this
issue's sibling, which shows that protection is itself incomplete).

## Actual Behavior

The daemon boots and arms every subscription with no trust check of any kind,
then executes tool calls (including `execute_bash`) with no human
confirmation and no default sandbox.

## Additional Context

      grep across `source/daemon/` for `trust`/`Trust` returns zero matches)

**Suggested fix**: add a trust check to the daemon boot path, e.g. refuse to
start (or start in a strictly reduced/no-exec tool posture) unless the
project directory is already in `preferences.trustedDirectories`, requiring
at least one prior interactive `nanocoder` run (or an explicit
`--trust-directory` flag) before `daemon start` will arm any subscription.

yetuge:
Fixes #1245

## Root cause

`nanocoder daemon start` is a fast path in `source/cli.tsx` that branches off before the interactive app and goes straight to `runDaemonCli("start")` → `launchDaemon` → `entry.ts`. Nothing along that path — or anywhere under `source/daemon/` — consults `preferences.trustedDirectories`: the trust check only lives in the interactive TUI (`useDirectoryTrust`) and the plain shell (`source/plain/shell.ts`).

Once booted, the daemon dispatches subagents in headless mode, which skips tool confirmations for `execute_bash`, `write_file`, `string_replace`, `diff_edit` and MCP tools, and OS-level sandboxing is opt-in (unsupported on Windows). So a daemon started in a directory the user never trusted arms every `.nanocoder/` skill subscription for unattended execution.

## Fix

Add a non-interactive trust gate with the same rules the plain shell already enforces (`ensureDirectoryTrust` in `source/plain/shell.ts`), applied at both layers that can boot a daemon:

- **`start()`** (shared helper in the new `source/daemon/trust.ts`): refuses with a user-facing message unless the project root is already in `preferences.trustedDirectories`, the caller passes `--trust-directory`, or `NANOCODER_TRUST_DIRECTORY=1` is set. Both explicit mechanisms record standing trust (like the plain shell's env-var path), because the daemon boots detached and can boot again via autostart without the CLI — ephemeral trust would not survive the spawn.
- **`entry.ts`**: the boot itself re-checks the same rule before standing up the pipeline, so autostart boots (launchd/systemd, set up by `daemon install`) that never run the CLI are covered too. The CLI gate is still needed: it fails before spawning, with a message the user actually sees, whereas a boot-gate-only refusal would land in `daemon.log`.

`--trust-directory` is now honored for `daemon start` (previously documented as applying only to `nanocoder run`); for a daemon it records standing trust rather than "this run only", since the process it arms outlives the invoking shell. `daemon install` itself is deliberately left ungated — it only writes the autostart registration, and the boot gate now covers the path it arms.

## Tests

New regression tests in `source/daemon/cli.spec.ts` (run against unpatched `main` with only the tests applied, then with the fix):

- `start refuses to spawn when the project directory is not trusted` — asserts exit 1, a `/not trusted/` message, **no** launcher invocation and no trust writes. **Failed on `main`** (the daemon spawned regardless).
- `start marks the directory trusted and proceeds when NANOCODER_TRUST_DIRECTORY=1` — asserts standing trust is persisted. **Failed on `main`**.
- `start records standing trust when --trust-directory is passed` — **failed on `main`** (the flag was ignored).
- `start proceeds when the directory is already trusted` — guards the happy path; passes before and after.

Red→green: on `main` (`537ee6c`) the new tests report `3 tests failed`; with the fix, `pnpm exec ava source/daemon/cli.spec.ts` → `12 tests passed`. Also green: `source/daemon/lockfile.spec.ts` + `source/daemon/install.spec.ts` (30), `source/plain/shell.spec.ts` (22), `source/cli.spec.ts` (54). `pnpm test:types` and `biome check` are clean on the touched files.

Fixes #1245


yetuge:
Addressed in b27380a — thanks, this round was useful.

**`nit` · `correctness` (cli.tsx:68)** — the daemon fast path no longer looks at `args.slice(2)`. The flag is read through `hasTrustDirectoryFlag()` in `source/daemon/trust.ts`, shared with the boot gate and position-agnostic, so `nanocoder --trust-directory daemon start` no longer silently degrades to "untrusted". Specs cover flag-before-subcommand, flag-after-subcommand, absent, and a lookalike (`--no-trust-directory`).

**`important` · `completeness` (cli.tsx:165)** — the `--help` text now says the flag is valid with `run` (that run only) and with `daemon start` (which also records standing trust, so the detached boot and any autostart boot skip the gate), and a `nanocoder daemon start --trust-directory` example was added. The same incorrect claim lived in `docs/getting-started/index.md`, so that row was updated too.

**`important` · `tests` (entry.ts:56)** — the boot rule moved into `checkDaemonBootTrust()` in `source/daemon/trust.ts` and `entry.ts` now calls it. A spec cannot import `entry.ts` as it stands: it runs `main()` at module load, so importing it would chdir and boot a real daemon. The rule is now testable where it lives, with specs for untrusted → refused with the message naming the directory, already-trusted → passes, `NANOCODER_TRUST_DIRECTORY=1` → passes and persists, plus one asserting the boot never derives trust from the invoking argv (autostart boots carry none). The honest gap that remains: nothing asserts `entry.ts` and `cli.tsx` *call* these helpers — that needs either a spawn harness around `cli.tsx` or splitting the `entry.ts` bootstrap so `main` can be imported with stubbed deps. Happy to do either in this PR if you want it here rather than as a follow-up.

**`important` · `scope` (trust.ts:38)** — took the documentation branch rather than splitting into `isDirectoryTrusted` / `trustDirectory`. The asymmetry is now spelled out on `ensureDirectoryTrust` in `source/daemon/trust.ts`, with the reason (the daemon outlives the CLI process and can be re-booted by autostart, so run-scoped trust would drop the user back to "untrusted" on the next boot), and `plain/shell.ts` carries the reciprocal pointer. If you'd prefer the split so the name carries the side effect, it's a small change — say the word.

**`nit` · `changeset`** — added `.changeset/daemon-boot-trust-gate.md` describing the new gate, the `daemon start` semantics of `--trust-directory`, and the autostart coverage.

**`blocking` · `duplicate` (#1259)** — flagging this instead of acting on it, since it isn't mine to resolve. #1259 was opened first (10:19Z) and covers the same issue. The one capability this PR has that I can't see in #1259's file list is the second layer: `entry.ts` re-checks the rule inside the daemon process, which is the only path covering `daemon install`'s launchd/systemd autostart, because those boots never run the CLI and so have no argv to read a flag from. Options as I see them — (a) land one and port the boot layer into the other as a follow-up, (b) leave both open and let a maintainer pick, (c) I close this one so the queue stays clean. No preference between (a) and (b); happy to rebase onto whichever you'd rather take. For transparency, I opened this before spotting that #1245 was already assigned to @addyCooks — I'd rather raise the overlap than leave two competing PRs sitting in the queue.

`source/daemon/trust.spec.ts` is new (9 specs, green); the existing four gate specs in `source/daemon/cli.spec.ts` still pass. `pnpm test:types`, `biome check`, the daemon suite, `source/cli.spec.ts` (54) and `source/plain/shell.spec.ts` (22) are clean locally; the only local failures are the pre-existing Windows `ipc` socket specs, which fail identically on `main`.

yetuge:
Follow-up: merged `main` into the branch (`773d34bd`) — the PR had gone stale against the base once #1261 landed. The only conflict was the `--trust-directory` row of the CLI flag table in `docs/getting-started/index.md` (main rewrote the neighbouring `--alt-screen` rows), resolved by keeping this branch's trust wording and main's alt-screen wording. Nothing under `source/daemon/` changed in the merge; `tsc --noEmit` is clean and the trust / daemon-cli / plain-shell specs are green (43 passed) afterwards.

will-lamerton (changes requested):
The fix itself is sound, and the response to the previous round was thorough. I fetched the branch, merged current `main` in (clean), and ran `tsc --noEmit`, `biome check` on the touched files, and the trust / daemon-cli / plain-shell / cli specs: all clean, 100 tests pass. The two-layer design checks out, `runDaemonCli` is the only production caller of `launchDaemon`, the plist and systemd unit set only `NANOCODER_PROJECT_ROOT` so autostart really does depend on persisted trust, and the boot gate sits before `createLLMClient` and `startDaemon`.

Requesting changes on the following.

**Blocking, not yours to fix:** #1259 is still open and #1245 is still assigned to @addyCooks, so a maintainer needs to pick a lane before either lands. Separately, the required `pr-checks` workflow has never run on this PR (only `changeset-check` and `label` did), so the green ticks above mean less than they look. The branch is also 128 commits behind `main`.

**Changeset is factually wrong.** `.changeset/daemon-boot-trust-gate.md` says `NANOCODER_TRUST_DIRECTORY=1` "works for a single boot". It doesn't: the env-var path calls `markTrusted` too, so it persists standing trust exactly like the flag. The behaviour is right (it mirrors `plain/shell.ts`), only the text is wrong, and that text ships in the changelog.

**The `--trust-directory` split deserves one more look.** Same flag, opposite persistence semantics: ephemeral for `run`, a permanent global trust grant for `daemon start`. You documented it well, but documenting a footgun doesn't disarm it, and `nanocoder daemon start --trust-directory` in a throwaway or shared build directory permanently trusts that path. Worth considering: have `daemon start --trust-directory` propagate `NANOCODER_TRUST_DIRECTORY=1` into the spawned child's env rather than persisting. The boot gate passes without a permanent write, and autostart, the one case that genuinely needs persistence, is already behind the explicit `daemon install`.

**Docs still walk users into the refusal.** `docs/features/skills.md:173` and `docs/features/scheduler.md:74` both tell people to run `nanocoder daemon start` with no mention of the new gate. The flag table in `getting-started` was updated, but those are the two pages where people actually meet the daemon. Related: `daemon install` stays ungated by design, so installing in an untrusted directory succeeds and then every autostart boot fails into `daemon.log` where nobody looks. A warning at install time closes that cheaply.

**Silent `savePreferences` failure.** `savePreferences` swallows write errors and returns void, and `markTrusted` returns `true` regardless. On a failed write the CLI prints "Marked X as trusted. Daemon started" while the child's own gate refuses and `waitForLockfile` times out. Rare, and it fails safe, but the message is untrue.

**Test gap:** please take your own offer here rather than deferring it. Nothing asserts that `cli.tsx` and `entry.ts` actually call the helpers, so a regression deleting either call site passes the suite. For a security gate that's the regression most worth catching.

**Minor:** the gate runs before `readLiveLockfile`, so `daemon start` in an untrusted directory that already has a daemon running reports "not trusted" instead of "already running".

yetuge:
Closing this in favour of #1429. #1259 landed on 2026-09-15 and I built this PR on a premise that does not survive checking it against the merged tree, so keeping it open would just ask a maintainer to referee a duplicate.

**The boot gate here is redundant.** I claimed it covered launchd / systemd / Task Scheduler autostart "boots that never run the CLI". That is wrong, and I should have verified it before writing it three times in this PR's description, the changeset and the code comments. All three autostart mechanisms invoke `nanocoder daemon start` — `buildLaunchAgentPlist`, `buildSystemdUnit` and `buildScheduledTaskXml` all spell out `daemon start` in `source/daemon/install.ts` — and the only caller of `entry.js` anywhere in the tree is `launchDaemon` in `source/daemon/cli.ts`. `runDaemonCli` checks `ensureDirectoryTrust` before calling it, so the gate `#1259` merged sits on every path that can boot a daemon. My `entry.ts` gate was a second copy of a check that had already run.

The `why does entry.ts need its own gate` question resolved the same way: it does not, because nothing reaches it without passing the CLI check first.

**What is still worth fixing is the three things you listed as non-blocking**, and they are all still missing from `main`:

- `ensureDirectoryTrust` reported `persisted: true` on an attempted write, while `savePreferences` swallows write errors — so `daemon start` printed "Marked <dir> as trusted" for a directory the next boot would refuse.
- `docs/features/skills.md:173` and `docs/features/scheduler.md:74` still point users at `nanocoder daemon start` with no mention of the gate, and `daemon install` in an untrusted directory still fails later into `daemon.log`.
- `cli.tsx` now derives the flag as `args.includes('--trust-directory')`, which is correct for both flag positions after #1259's change, so the position-agnostic helper this PR added is no longer needed either.

#1429 carries those three, plus a `loadService` seam on `DaemonCliOptions` so the install path can be tested without registering a real service. It is a fresh branch off current `main` with no shared history, so it will not conflict.

Sorry for the noise on this one — the flag semantics and the position-agnostic parsing were real findings at the time, but the duplicate question you raised was the right one and the answer was to not land a second gate.

PR #1261, opened 2026-09-11
feat(tui): fullscreen UI overhaul - alt screen default, rounded chrome, wheel scrolling

will-lamerton:
Long-lived UI branch. Makes the fullscreen TUI the default and brings the surrounding chrome in line with it.

## What's in it

- **Fullscreen by default.** Interactive sessions run on the alternate screen buffer with a bottom-anchored, clipping transcript viewport, so a long turn can't push the prompt off-screen. `--no-alt-screen` / `alternateScreen: false` keeps inline mode with native scrollback.
- **Wheel scrolling.** SGR mouse reporting is on by default so the wheel scrolls the transcript (Shift+drag, or Option+drag in iTerm2, to select text). `--no-mouse` opts back into native selection. This also disables alternate scroll mode (DECSET 1007) on the opted-out path, which was translating wheel ticks into arrow keys and cycling prompt history.
- **Welcome redesign.** Gradient block wordmark, centered quick-action menu, branch/directory line, all on one center axis. The subtitle is now the repo description.
- **Rounded chrome everywhere.** Prompt, session selector, IDE selector, file explorer, plan review, question prompt and tool confirmation all share the same rounded titled box; transcript and mode indicator align to the prompt's left border.
- **`/explorer` no longer crashes.** It was killing the CLI with exit 1 (`useUIStateContext must be used within a UIStateProvider`) because the provider wrapped only `ChatInput`. Hoisted to the whole interactive tree, which also makes the explorer's file selection actually reach the prompt.
- **Smaller fixes.** The startup summary shows the workspace instead of a config path; mode-switch model toasts no longer stack.

## Review notes

- Changesets: the branch was carrying one changeset for 28 commits. Added six covering the rest.
- `bracketed-paste-and-selection-mode.md` (unreleased) documented a **Ctrl+P selection mode that this branch deletes**, and called inline mode the default. The text is corrected, but whether Ctrl+P should come back is worth a decision: swapping a keystroke toggle for a config preference is a downgrade for anyone who selects text often.
- `interactive-app.spec.tsx` had been red since `1f833023` on a stale placeholder assertion; fixed.
- Two test harnesses were hiding bugs rather than catching them. `renderWithTheme` supplied a `UIStateProvider` the real app never had, which is why the explorer crash survived a green suite, and `file-explorer/index.spec.tsx` renders hand-written stand-ins instead of the real component. Both now have tests against the real thing.

https://claude.ai/code/session_01NPtyuUUFiQMF5Qf3dqo6hW


coralsundy:
Apparently --no-alt-screen or set   "alternateScreen": false does not disable new TUI for me.

Edit: Apparently the alternative screen is just fullscreen mode, has nothing to do with the new TUI, trying to find a way to restore the old TUI.

PR #1276, opened 2026-09-12
fix(tui): stop paste-then-keystroke race from dropping composer content

### LINKED ISSUE #1273: [Bug] Pasting into the composer, then typing immediately, silently deletes the pasted content
## Description

Pasting a multi-line block into the composer and then typing even a single character right afterward, with no deliberate pause, causes the entire paste to silently disappear. No error, no warning; the composer just ends up containing only what you typed after the paste.

**Note:** This is specific to `feature/new-ui`'s composer state handling. Please target any fix at `feature/new-ui`, not `main`.

Parent: #907

## Environment

- Reproduced via `ink-testing-library` against the real `UserInput` component (`stdin.write()` driving actual paste + keystroke sequences), and is expected to reproduce in a live terminal session the same way.
- Provider/model: not relevant, this is purely composer/input state.

## Steps to Reproduce

1. Start Nanocoder on `feature/new-ui` (`pnpm run start`).
2. Copy a multi-line block of text/code to your clipboard.
3. Paste it into the composer.
4. Immediately continue typing, e.g. append a short sentence right after the paste, without pausing. (This is timing-sensitive; if it doesn't reproduce on the first try, repeat a couple of times, the window is a short race, not guaranteed on every attempt.)

## Expected Behavior

The pasted content stays in the composer as a `[Paste #N: X chars]` placeholder, with whatever you typed afterward appended to it.

## Actual Behavior

The pasted content vanishes entirely. Only the text typed after the paste remains in the composer.

Captured via automated timing sweep (gap between paste and next keystroke):

gap 0ms  -> "x"                    <- pasted content GONE
gap 8ms  -> "x"                    <- pasted content GONE
gap 16ms -> "PASTED1\nPASTED2x"    <- survives once state settles

The same race also swallows the first Enter pressed right after a paste (Enter appears to do nothing).

## Root Cause

`insertPaste` (`source/components/user-input.tsx:191-204`) commits through `pushToUndoStack`, which reads a stale `currentState` snapshot rather than the state React has actually committed. The very next `onChange` event from `TextInput` rebuilds the input value from that stale pre-paste snapshot, overwriting the placeholder that was just inserted. See `source/hooks/useInputState.ts:276-305`.

## Suggested Fix

Make `useInputState`'s update path functional (`setCurrentState(prev => ...)`) instead of computing the next state from a captured `currentState` variable, so a paste commit can't be clobbered by the very next keystroke's state update.

## Additional Context

- [x] I can reproduce this issue consistently (timing-dependent, not on every attempt)
- [x] This issue occurs on the latest `feature/new-ui`

addyCooks:
## Description

Fixes #1273. 

Pasting into the composer and then typing (or pressing Enter) immediately afterward, with no pause, could silently drop the pasted content, or make Enter no-op.

Root cause: `insertPaste` and `updateInput` in `useInputState.ts` both computed their next state from the `currentState` closed over at the last render. A bracketed paste and a keystroke landing immediately after it (before React re-renders) both read that same stale snapshot, so the keystroke's commit clobbered the paste that had just landed. The same staleness meant Enter fired right after a paste would read the pre-paste state and silently no-op.

Fix: track a `currentStateRef` that `applyState` updates synchronously alongside React state, and have `insertPaste`/`updateInput`/`undo`/`redo`/`pushToUndoStack` read from it instead of the closure. `updateInput` additionally recognizes the specific paste-then-type race (via a one-shot `lastPasteCommitRef` set by `insertPaste`) and replays the typed suffix onto the real latest value rather than trusting a stale base. `handleSubmit` now reads the same ref so Enter right after a paste submits the pasted content instead of no-op'ing.

## Recording

B02-FIXED-paste-survives

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Changeset

- [x] Added a changeset (`.changeset/quiet-pastes-linger.md`) describing this change for the changelog

### Automated Tests

- [x] All existing tests pass

### Manual Testing

- [ ] Tested with Ollama
- [ ] Tested with OpenRouter

## Checklist

- [x] Self-review completed

### DIFF
--- a/.changeset/quiet-pastes-linger.md
+++ b/.changeset/quiet-pastes-linger.md
@@ -1,0 +1,5 @@
+---
+"@nanocollective/nanocoder": patch
+---
+
+Fix a race in the composer where typing or pressing Enter immediately after a paste could silently drop the pasted content (or no-op the Enter), by reconciling stale input state against the latest committed value instead of the value captured at the last render.
--- a/source/components/user-input.spec.tsx
+++ b/source/components/user-input.spec.tsx
@@ -1211,3 +1211,55 @@ test.serial('UserInput ignores terminal pastes while disabled', async t => {
 	unmount();
 });
 
+test.serial(
+	'UserInput keeps a paste placeholder when typing lands immediately after it',
+	async t => {
+		const {stdin, lastFrame, unmount} = render(
+			<TestWrapper>
+				<UserInput forceFocus={true} />
+			</TestWrapper>,
+		);
+
+		await wait(50);
+		pasteEvents.emit('paste', 'line one\nline two\nline three');
+		// No await between the paste and the keystroke: this is the race the
+		// bug reports - typing immediately after a paste, before React has
+		// re-rendered with the placeholder, must not clobber it.
+		stdin.write('x');
+
+		await waitForFrame(lastFrame, /\[Paste #\d+: \d+ chars\]x/);
+
+		t.regex(lastFrame()!, /\[Paste #\d+: \d+ chars\]x/);
+		unmount();
+	},
+);
+
+test.serial(
+	'UserInput submits a paste when Enter lands immediately after it',
+	async t => {
+		let submittedDisplay: string | undefined;
+
+		const {stdin, lastFrame, unmount} = render(
+			<TestWrapper>
+				<UserInput
+					forceFocus={true}
+					onSubmit={(_message, display) => {
+						submittedDisplay = display;
+					}}
+				/>
+			</TestWrapper>,
+		);
+
+		await wait(50);
+		pasteEvents.emit('paste', 'line one\nline two\nline three');
+		// No await: Enter races the same re-render as above.
+		stdin.write('\r');
+
+		await waitForCondition(() => submittedDisplay !== undefined);
+
+		t.regex(submittedDisplay!, /\[Paste #\d+: \d+ chars\]/);
+		t.is(lastFrame()!.includes('[Paste #'), false);
+		unmount();
+	},
+);
+
--- a/source/components/user-input.tsx
+++ b/source/components/user-input.tsx
@@ -155,6 +155,7 @@ export default function UserInput({
 		resetInput,
 		deletePlaceholder: _deletePlaceholder,
 		currentState,
+		currentStateRef,
 		setInputState,
 		insertPaste,
 	} = inputState;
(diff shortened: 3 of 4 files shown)


will-lamerton (changes requested):
Thanks for this - the bug is real and still reproduces. I applied your two new tests to an unmodified `feature/new-ui` and both fail there, so this isn't redundant with what's already landed. Full `user-input.spec.tsx` is green on your branch, and the changeset is correct.

That said, I don't think it's mergeable as-is, mainly because the base moved under you.

**1. Conflicts with #1170, which landed the same idea.** `feature/new-ui` now has its own `currentStateRef`, plus `undoStackRef`/`redoStackRef` and a `MAX_UNDO_STACK` cap, and `updateInput` there already does `const currentState = currentStateRef.current`. There are 7 conflict hunks in `useInputState.ts`, all in that overlapping logic. Resolving any of them in this PR's favour silently reverts the stack cap and the ref-backed stacks.

**2. The replay heuristic can't survive that rebase.** It fires only when `currentState.displayValue !== latestState.displayValue`, where `currentState` is the render-time closure. On current base those two are the same value, so the recovery becomes dead code.

**3. Verified regression: backspace right after a paste.** Typing `abc`, pasting, then Backspace with no pause, deterministic over 3 runs:

```
current feature/new-ui : abc[Paste #1: 28 chars]   placeholder survives
this PR                : ab                        placeholder destroyed
```

The guard requires `newInput.startsWith(currentState.displayValue)`, which a deletion fails, so it falls through and commits the stale pre-paste `placeholderContent`. Base is protected here because it reads the ref unconditionally.

**4. The fix only survives one keystroke.** `lastPasteCommitRef` is one-shot, so two characters landing before the re-render:

```
current feature/new-ui : xy    paste dropped
this PR                : yx    paste dropped, and order scrambled
```

Typing a word right after a paste is the realistic case, and it still loses the paste.

**5. CI has never run** on `5aa53ce0` (fork PR pending workflow approval), so nothing here has been validated by the pipeline.

**Suggested direction:** rebase onto current `feature/new-ui`, keep #1170's ref-backed `pushToUndoStack` / stacks / cap intact, and land the parts that are genuinely still missing there: `insertPaste` reading `currentStateRef`, `resetInput` syncing the ref, and exporting `currentStateRef` so `handleSubmit` reads it (the Enter-after-paste half, which merges cleanly). For the remaining clobber, the root cause is that `TextInput` hands back a `newInput` built from the value it last *rendered*; that wants a generic reconcile - diff the incoming value against the stale rendered value and replay the delta onto the ref, for inserts *and* deletions - rather than a one-shot append-only special case. Please add coverage for multi-keystroke and backspace-after-paste.

PR #1285, opened 2026-09-13
fix(config): preserve MCP source field in loadAppConfig for security validation

### LINKED ISSUE #1248: [Bug] `validateProjectConfigSecurity` never runs, hardcoded-credential scanner is dead code

addyCooks:
## Description

Fixes #1248

The `validateProjectConfigSecurity` function was never executing because the `.source` field (which identifies whether an MCP server config is project-level, global, or environment-based) was being stripped during config loading in `loadAppConfig`.

**Root cause**: In `source/config/index.ts`, the function unwrapped `MCPServerWithSource[]` objects by mapping to just `item.server`, which discarded the wrapper's `.source` field that the security validator depended on. The validator filtered for `server.source === 'project'`, but this field was always `undefined` in production.

**The fix**: Preserve the `source` field when unwrapping by spreading both `item.server` and `source: item.source`.

The hardcoded-credential scanner now correctly receives the source field and will warn users about project-level MCP configs containing apparent secrets (hardcoded API keys, tokens, passwords, etc.).

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Changeset

Changeset file: `.changeset/fix-mcp-config-security-validation.md`

### Automated Tests

### Manual Testing

- [ ] Tested with Ollama
- [ ] Tested with OpenRouter
- [ ] Tested with OpenAI
- [x] Tested MCP integration

## Checklist

- [x] Self-review completed

### DIFF
--- a/.changeset/fix-mcp-config-security-validation.md
+++ b/.changeset/fix-mcp-config-security-validation.md
@@ -1,0 +1,7 @@
+---
+"@nanocollective/nanocoder": patch
+---
+
+fix(config): preserve MCP source field to enable project-level security validation
+
+The validateProjectConfigSecurity function now correctly validates project-level MCP configs for hardcoded credentials. Previously, the .source field was being stripped during config unwrapping, causing the security check to never run.
--- a/source/config/index.ts
+++ b/source/config/index.ts
@@ -702,7 +702,10 @@ function loadAppConfig(): AppConfig {
 
 	// Load MCP servers from the new hierarchical configuration system
 	const mcpServersWithSource = loadAllMCPConfigs();
-	const mcpServers = mcpServersWithSource.map(item => item.server);
+	const mcpServers = mcpServersWithSource.map(item => ({
+		...item.server,
+		source: item.source,
+	}));
 
 	// Load auto-compact configuration
 	const autoCompact = loadAutoCompactConfig();


will-lamerton:
Hey! Can you look at the merge conflicts and nc-review points please @addyCooks? :)

PR #1286, opened 2026-09-13
feat(voice): Add local-first realtime voice mode

RONAK-AI647:
## Summary

Adds a local-first voice mode for Nanocoder with push-to-talk, hands-free VAD, speech transcription, text-to-speech, and interruption support.

This PR combines the six voice-mode implementation phases and includes the fixes from will's review and this fixes issue #622 

## Scope clarification

The Bash executor change is required for voice-mode barge-in. Hands-free voice can interrupt an active response while a Bash tool is running, so the executor now escalates from SIGTERM to SIGKILL when a process ignores cancellation. This prevents interrupted voice sessions from leaving child processes running.

Hands-free voice should not be combined with `yolo` or automatic tool approval in noisy environments, because unintended speech or background noise may trigger agent actions. Push-to-talk or normal approval mode is recommended for safer use.

## Features

- `/voice` command for enabling and configuring voice mode
- Push-to-talk recording with `Ctrl+T`
- Hands-free voice mode using VAD
- Local Whisper STT support
- Local Piper TTS support
- Optional cloud STT/TTS for verified OpenAI providers
- Barge-in interruption during:
  - Response generation
  - Tool execution
  - Speech synthesis
  - Audio playback
- AbortSignal propagation through tool execution
- VAD worker cleanup and bounded audio buffering
- Voice status UI
- On-demand dependency installation
- SHA256 checksum verification for downloaded assets
- Architecture-aware dependency installation
- Bundled voice plugin for published builds

## Review fixes included

- Rebased onto the latest `main`
- Fixed voice dependency installer URLs and platform handling
- Added HTTPS-only downloads and checksum validation
- Added real installer URL-path test coverage
- Fixed optional voice plugin build and distribution
- Removed the unused realtime capability abstraction
- Added Ctrl+T enabled-state protection
- Replaced repeated preference disk reads with cached reactive preferences
- Fixed VAD shutdown races and microphone orphaning
- Separated recording and TTS temporary-file ownership
- Fixed VAD temporary WAV leaks
- Added sustained-speech buffer limits
- Filtered `[BLANK_AUDIO]` responses
- Restricted cloud audio support to verified OpenAI providers
- Wired `voiceName` into cloud TTS
- Fixed signal-killed process handling
- Restored affected documentation
- Added a changeset
- Included voice plugin files in Biome validation

## Validation

- Main TypeScript checks pass
- Voice plugin TypeScript checks pass
- Biome checks pass
- 36 focused voice tests pass

## Notes

Native provider-specific realtime streaming is intentionally not included in this PR. Cloud voice currently uses STT/TTS REST endpoints, while all providers retain the existing Nanocoder request pipeline.

Hands-free VAD combined with automatic tool approval may require an additional maintainer decision regarding ambient-noise safety.


will-lamerton:
Hey  @RONAK-AI647 - looking great! Let me know when you'd like me to review :)

RONAK-AI647:
@will-lamerton , There are some issues yet there , I am sorting them , this month was hectic , apologies for delay.

I would love to see a detailed review if you do , makes my debugging easier.

will-lamerton:
@RONAK-AI647 Thanks for the thorough work here - the architecture is right (optional plugin isolating native deps, local by default, cloud allowlisted and failing closed) and the security hardening is genuinely good: HTTPS-only with redirect checks, pinned SHA256s with unlink-on-mismatch, whisper built from a checksummed source tarball.

But CI is red on four checks, and I root-caused three of them to two defects. Requesting changes.

## Blocking

**1. `tsconfig.json` alias points at a build artifact.** `"@nanocollective/nanocoder-voice": ["./plugins/voice/dist/index.d.ts"]` only resolves after the plugin is built. `Type Checks` has no build step, so on a clean checkout:

```
source/commands/voice.tsx(93,6): error TS2307: Cannot find module '@nanocollective/nanocoder-voice'
source/hooks/useVoice.ts(84,4): error TS2307: Cannot find module '@nanocollective/nanocoder-voice'
```

Same cause kills `Config Schema Freshness` - `typescript-json-schema` aborts with "No output definition" because of those two errors. One fix clears both: point the alias at `./plugins/voice/src/index.ts`, or commit a `.d.ts` shim.

**2. Six `useVoice` tests read your real preferences file.** All six CI failures assert a non-idle state and get `'idle'`. Those specs don't pass a `voicePreference` prop, so `useVoice` falls back to `getVoicePreference()`, which reads `~/Library/Preferences/nanocoder/preferences.json`. Voice is enabled there on your machine from manual testing, so `startStopRecording` gets past `if (!pref.enabled) return`. On CI the default is `enabled: false` and every test bails on line one. Pass an explicit `voicePreference` in each spec, and set `NANOCODER_CONFIG_DIR` to a tmpdir in `test.before`.

**3. The Biome re-inclusion does not work.** Biome will not re-include a nested path under an excluded directory, so `"plugins/voice/src/**"` after `"!plugins"` is a no-op. Verified on the branch:

```
Checked 0 files in 3ms.
x No files were processed in the specified paths.
i These paths were provided but ignored:
  - plugins/voice/src/
```

That is why `Format Checks` and `Linting` are green while `plugins/voice/src/*.ts` uses non-repo style throughout (brace spacing in imports, >80 col lines, `try { ... } catch {}` one-liners). "Included voice plugin files in Biome validation" is listed as a completed review fix but is not in effect - worth re-verifying the rest of that list the same way. Narrow the exclusion (`"!plugins/vscode"`, `"!plugins/*/dist"`) and run `biome check --write` over the plugin.

**4. Release-pipeline hazards.** `plugins/voice/package.json` has no `"private": true` and `.changeset/config.json` only ignores `nanocoder-vscode`, so changesets treats the voice package as publishable. Separately, `optionalDependencies: {"...-voice": "workspace:*"}` is rewritten to a concrete version on publish and will fetch a package that does not exist on npm. It is also redundant - the build copies the plugin into `dist/voice/` and `defaultLoadPlugin` already falls back to `../voice/index.js`. A `devDependency` gives you the workspace link without shipping it in the published manifest.

The standalone `CodeQL` check also fails, but in 4s while `pr-checks / CodeQL Security Analysis` passes - looks like infra, worth confirming.

## Correctness

- **TTS may speak an intermediate agent turn.** The effect fires whenever `messages` changes and the last message is `role === 'assistant'`. `conversation-loop.tsx` calls `setMessages` at every turn boundary, including turns carrying tool calls. On a multi-step run the first "let me check that file" turn triggers synthesis, consumes `pendingTTSRef`, and the real answer is never spoken. Gate on conversation completion instead.
- **VAD clips speech onset.** `vad-worker.ts` only buffers frames after RMS crosses the threshold, so the leading consonant of every utterance is discarded. A small pre-roll ring buffer flushed on `speech_start` fixes it.
- **SIGTERM'd `sox` may leave an invalid WAV.** sox patches the RIFF data-length field on clean close; SIGTERM skips that. Worth confirming whisper accepts the file - `SIGINT` is the signal sox traps for graceful stop.
- **`killProcessTree`'s SIGKILL guard is wrong.** A signal-killed process has `exitCode === null` and `signalCode === 'SIGTERM'`, so the guard reads "alive" for something SIGTERM already killed. Only the timer cleanup saves it today; if that ordering shifts you would SIGKILL a possibly-recycled PID. Use a boolean set in the exit handler. Also, each call adds `once('close')` + `once('exit')`, which accumulates under repeated `cancel()` - exactly the barge-in pattern this PR enables.
- **`cloud-audio.ts` never removes its abort listeners** - add cleanup in `finally`.
- **Dependency specs write to the real config dir.** The URL-construction test `mkdirSync`s the actual `~/.config/nanocoder/voice-bin`. Also, the non-root sudo test assumes a non-root runner - fine on `ubuntu-latest`, fails in a root container.

## Quality

- `recordingAudioPromiseRef` (`useVoice.ts:154`) is written and nulled, never read. Dead.
- The utterance-finalize block in `vad-worker.ts` is copy-pasted three times - extract `finalizeUtterance()`.
- The silence-marker regex is duplicated in `stt.ts` (`filterSilenceMarkers`) and `useVoice.ts` (`isBlankAudio`).
- `VadEventMap` is declared but never applied as the generic (`extends EventEmitter<VadEventMap>`), which is why `useVoice` casts to an inline shape with a `noExplicitAny` suppression.
- `InstallDependenciesOptions.installRunner` is marked `@deprecated` in a file that is new in this PR - just delete it.
- `plugins/voice/.gitignore` says `plugins/voice/real-test.mjs`; inside that file the pattern is directory-relative, so it matches nothing. Should be `real-test.mjs`.
- `plugins/voice/package.json` declares `"test:ava": "ava"` with no ava config or devDependency in that package. (The specs do run - root ava globs `plugins/**/*.spec.ts` - the per-package script is just dead.)
- `vad-worker.ts` handles `proc.on('error')` but not `'close'`. If `rec` dies mid-session the worker goes silent with nothing surfaced.
- `interactive-app.tsx` calls `useTheme()` mid-body just above the return - move it up with the other hooks.
- `CLAUDE.md` is not updated for the voice plugin or the new `plugins/` build step.

## Security

Two things to weigh rather than defects:

- Cloud STT ships microphone audio to OpenAI. It is correctly gated behind an explicit `/voice stt cloud`, but neither the command output nor the docs say the audio leaves the machine. For a local-first product that deserves one explicit line at the point of opt-in.
- Hands-free VAD combined with `auto-accept`/`yolo` is a real ambient-noise hazard, as you note. My read is this should be enforced in code rather than documented - refuse hands-free while in `yolo`, or force `normal` approval for the duration. A docs warning will not stop a podcast from synthesizing a tool call.

Happy to re-review once the four CI checks are green.

RONAK-AI647:
@will-lamerton , I request another review , any internal blockings or errors which I am unable to see, once thats fixed , will work towards the devlopment and demo.

And why exactly is this one check failing, whats the exact unit test failing, could you help me with this.

PR #1298, opened 2026-09-14
fix(tools): cap long tool output and add /expand for a single result

### LINKED ISSUE #1289: [Bug] Tool output can only be expanded all-at-once, and some formatters have no line cap
## Description

The only way to see a tool result's full output is a single global toggle (Ctrl+O) that switches *every* tool result in the conversation between a compact one-liner and full formatter output at the same time. There's no way to expand just the one result you actually want. This is made worse by the fact that not every formatter caps its expanded output `write_file`'s expanded view renders the entire file with no line limit, unlike `execute_bash`, which is capped at a fixed number of lines with a "+N earlier lines" indicator.

Parent: #907

## Steps to Reproduce

1. Have a conversation with a mix of tool calls, including at least one `write_file` on a large file.
2. Press Ctrl+O to toggle compact/expanded display.
3. Note it affects every tool result at once, and that the `write_file` result in expanded mode has no line cap.

## Expected Behavior

Expanding a result should be a per-message action, and every formatter's expanded output should be capped at a reasonable length, the way `execute_bash`'s already is.

## Actual Behavior

Expand/collapse is global and binary; `write_file` in particular has no cap in expanded mode.

## Recording

Image

## Root Cause

`source/utils/tool-result-display.tsx` the `compact` parameter on `displayToolResult`, together with the `ALWAYS_EXPANDED_TOOLS` set, apply one boolean to every tool result uniformly. `source/components/bash-progress.tsx` caps `execute_bash` output at `BASH_OUTPUT_DISPLAY_LINES`; `source/tools/file-ops/write-file.tsx`'s `WriteFileFormatter` has no equivalent cap.

Note: #769 (closed, fixed by merged PR #786) added a related but different 20,000-character cap that one bounds only the *LLM-facing* text of oversized results before they enter the conversation, explicitly keeping "structured tool payloads intact." It doesn't touch this terminal display, which reads `args.content` directly.

## Suggested Fix

Replace the global toggle with a per-message expand affordance, and apply a display-level line cap to every formatter's expanded view, not just bash's.

## Additional Context

addyCooks:
Closes #1289.

## Description

Two problems from the issue:

1. **Some tools printed unlimited output.** `execute_bash` stopped at 20 lines, but `write_file`, `string_replace`, `diff_edit` and tools without a formatter (such as MCP tools) printed everything.
2. **Expanding was all-or-nothing.** Ctrl+O switched every tool result between compact and full view at once, so there was no way to open just one.

**Line cap.** A new shared `CappedLines` component (in `tool-message.tsx`) shows the first 20 lines, then `… (+N more lines · /expand n)`. It only renders the lines it shows, so a very large file costs nothing extra. It's used by the `write_file`, `string_replace` and `diff_edit` formatters and by `ToolMessage`'s plain-text output, which covers tools without a formatter. `BASH_OUTPUT_DISPLAY_LINES` becomes `TOOL_OUTPUT_DISPLAY_LINES` so every tool uses the same limit.

**Per-result expand.** Each tool result now gets a number when it's shown. `/expand` lists recent results, and `/expand <n>` prints that one result in full (for `execute_bash`, this includes the command's output). It adds new output rather than changing old output, so it works in inline mode, where printed history can't change, and for results merged into compact tallies like "Read 7 files". Ctrl+O and the "Expand tool results by default" setting work as before. `/clear` also clears the list of expandable results.

## Recording

X02-FIXED-tool-output-cap

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Changeset

- [x] Added a changeset describing this change for the changelog

### Automated Tests

New tests: `CappedLines` (capped, uncapped, short output, `/expand` hint), `ToolMessage` plain-text cap, `write_file` cap, the `/expand` hint on a displayed result, and `/expand` itself (empty list, listing, full output, unknown number, tools that are never capped).

`test:types`, lint/format, `knip`, `test:audit` and `pnpm run build` all pass. Every spec for the changed files passes. On my Windows machine the full `pnpm test:all` fails, but with the same failures as unmodified `main` (Windows path and symlink tests, plus timeouts). Semgrep flags three existing files that this PR doesn't touch.

### Manual Testing

- [ ] Tested with Ollama
- [ ] Tested with OpenRouter

This only changes how results are displayed, so it doesn't depend on the provider. I rendered the real terminal output for a 30-line `write_file`, a 30-line result from a tool with no formatter, the `/expand` list, and `/expand <n>` for both a plain result and an `execute_bash` result.

## Checklist

- [x] Self-review completed


akramcodez (dismissed):
LGTM

akramcodez:
Thanks for the PR @addyCooks

addyCooks:
Thanks @akramcodez !

PR #1314, opened 2026-09-14
fix: reassemble UTF-8 chars split across bash output chunks

### LINKED ISSUE #1305: [Bug] BashExecutor stdout/stderr buffer concatenation slices multi-byte UTF-8 sequences across chunk boundaries
## Description

In `source/services/bash-executor.ts`, `proc.stdout` and `proc.stderr` handle incoming data by converting raw buffer chunks to strings using `.toString()`:
```ts
const limitedChunk = data.subarray(0, remaining);
state.fullOutput += limitedChunk.toString();
```
When multi-byte UTF-8 sequences (such as box-drawing borders, Unicode symbols `✔`, `✖`, progress bar characters, or non-ASCII characters in compiler output) are split across chunk boundaries or hit `remaining` bytes, calling `.toString()` on partial bytes causes them to decode as `\uFFFD` (the Unicode replacement character).

## Environment

- **Node version**: v20+ / v22+
- **Provider**: Any
- **Model**: Any

## Steps to Reproduce

1. Execute a command with dense Unicode output or box characters:
   ```bash
   !python3 -c 'print("─" * 100000)'
   ```
2. Inspect the output captured in `state.fullOutput` or rendered in the tool result.
3. Observe replacement characters (``) appearing at chunk boundary seams.

## Expected Behavior

Stream chunks should be decoded using `StringDecoder` from `node:string_decoder` to preserve multi-byte characters across chunk boundaries.

## Actual Behavior

Split multi-byte characters are converted to `\uFFFD`.

## Additional Context

kishore280:
Fixes #1305.

BashExecutor decoded each stdout/stderr chunk with Buffer.toString(). A multi-byte UTF-8 char split across a chunk boundary (pipe split or the truncation cut) decoded as replacement characters.

Switched to a StringDecoder per stream, flushed on close/error, skipped when already truncated.

Ran: test:types, test:lint, test:format, test:ava bash-executor.spec.ts


will-lamerton:
Rebased onto `main` and pushed. Two commits landed after this branch was cut - `91d509bd` (per-stream output budgets) and `e77289a5` (#1228) - which moved the whole output-collection block out of `BashExecutor` into `makeStreamCollector` in `source/utils/stream-collector.ts`. The lines this PR patched no longer existed, hence the conflict.

The fix itself was right and the bug was still unfixed on main (`append(limitedChunk.toString())`), so I moved it into the collector rather than replaying the diff. What changed from your version:

- `makeStreamCollector` now owns the `StringDecoder` and returns `{collect, flush}` instead of a bare listener, so there is somewhere to hang the end-of-stream flush. `BashExecutor` calls `flushStreams()` from `close`/`error`, same two sites you had.
- Your `outputTruncated` guard was a single flag shared across both streams, so stdout hitting the cap suppressed stderr's flush too. Now that each collector owns its budget the guard is naturally per-stream.
- `source/custom-tools/handler.ts` is the second `makeStreamCollector` caller, so it picks the fix up as well. Changeset updated to mention it.
- Added `source/utils/stream-collector.spec.ts` with unit coverage at the new home (no `python3` dependency, runs in milliseconds). Your four integration tests came across unchanged and still pass.
- Wrote down why the truncation path drops the decoder's pending bytes instead of flushing them, and why a cancelled run never reaches the flush - the two nits from the bot review.

Thanks for the fix, and sorry for the rebase over your commits. Merging once checks are green.

will-lamerton (approved):
Rebased onto the `makeStreamCollector` refactor and verified locally: 75 focused tests pass across `stream-collector`, `bash-executor` and `custom-tools/handler`, plus clean types, lint, format and knip. Good catch on the original bug.

PR #1331, opened 2026-09-15
fix(commands): don't report /update failure when the command exits 0

### LINKED ISSUE #1304: [Bug] hasCommandFailed in /update reports failure on exit code 0 if stdout/stderr contains words like "failed"
## Description

In `source/commands/update.tsx`, `hasCommandFailed` determines whether a package manager command succeeded or failed. In Strategy 1 (checking exit code), when `exitCode === 0`, it does not return `false`. Instead, execution continues down to Strategy 2 (pattern matching), which tests regexes including `/\bfailed\b/i` and `/\bcannot\b/i`. If the package manager exits with code 0 but outputs benign summaries like `0 vulnerabilities (0 failed)`, `hasCommandFailed` returns `true` and `/update` incorrectly displays an error message stating the update failed.

## Environment

- **Node version**: v20+ / v22+
- **Provider**: Any
- **Model**: Any

## Steps to Reproduce

1. Run `/update`.
2. Package manager runs `npm update -g @nanocollective/nanocoder`.
3. If npm outputs `added 1 package, removed 2 packages, 0 failed` and exits with code 0.
4. `/update` displays: `Failed to update Nanocoder automatically`.

## Expected Behavior

If the command exits with code 0 (`EXIT_CODE: 0`), it should be recognized as a successful run unless critical stderr messages indicate otherwise.

## Actual Behavior

Strategy 1 does not return `false` on `exitCode === 0`, allowing substring regexes in Strategy 2 to match words like "failed" in benign counts and trigger false failure reports.

## Logs/Screenshots

In `source/commands/update.tsx`:
```ts
// Strategy 1: Check exit code
const exitCodeMatch = outputStr.match(/^EXIT_CODE:\s*(\d+)/m);
if (exitCodeMatch) {
    const exitCode = parseInt(exitCodeMatch[1], 10);
    if (exitCode !== 0) {
        return true;
    }
    // BUG: Missing return false! It falls through to Strategy 2 below!
}

// Strategy 2: Check for critical error patterns
const criticalErrors = [
    ...
    /\bfailed\b/i, // Matches "0 failed", "tests failed: 0", etc.
    ...
];
```

## Additional Context

niukanen1:
Fixes #1304.

`hasCommandFailed` fell through from the exit-code check into pattern matching, so an exit code of 0 plus a benign summary like `0 vulnerabilities (0 failed)` was reported as a failure and `/update` showed "Failed to update".

When the command exited 0, only unambiguous error indicators (`command not found`, `no such file or directory`, `permission denied`, `error:` lines, `fatal`) still count as failure; the words "failed"/"cannot" in a success summary no longer do. The existing behavior for non-zero exit codes and for output without an exit-code line is unchanged.

**Testing**

- `pnpm run test:ava source/commands/update.spec.tsx` — 29 tests pass (22 existing + 4 new regression tests: `0 failed` summary, `0 cannot` summary, `error:` at exit 0 still fails, precedence checks).


niukanen1:
Addressed in b601781:

- **changeset**: added `.changeset/fix-update-exit-zero.md` (patch, `@nanocollective/nanocoder`).
- **tests**: every pattern retained in the exit-0 branch (`command not found`, `no such file or directory`, `permission denied`, `fatal`) now has a positive test asserting it still triggers a failure at exit 0, alongside the existing `error:` case.
- **description nit**: the PR body said 4 new regression tests but the diff had 3; the exit-0 positive tests above now cover the retained-pattern contract the description pointed at.

All 33 tests in `source/commands/update.spec.tsx` pass locally (`pnpm run test:ava source/commands/update.spec.tsx`).

akramcodez:
Thanks for the PR @niukanen1

PR #1353, opened 2026-09-16
fix(tui): select the completion when Enter follows a command fragment

### LINKED ISSUE #1327: [Bug] Enter pressed the instant a / menu opens submits raw text instead of selecting
## Description

`showCompletions` and `selectedCompletionIndex` are read from a stale render closure. A fast typer who presses Enter right after typing a slash-command fragment sends the raw, unfinished text to the model as a chat message instead of running (or even attempting to select) a command.

Parent: #907

## Steps to Reproduce

1. Type a slash-command fragment that doesn't exactly match a command (e.g. `/xy`).
2. Press Enter immediately, with no pause, as soon as the last character is typed.

## Expected Behavior

Either the completion menu stays open for a real selection, or nothing submits until a valid command is chosen.

## Actual Behavior

The raw fragment gets sent to the model as an ordinary chat message.

## Recording

Image

## Root Cause

`source/components/user-input.tsx:869-897` gates Enter's "select the completion" behavior on `showCompletions` and `selectedCompletionIndex`, both read from a closure that can be stale relative to the latest keystroke.

## Suggested Fix

Gate Enter on the memoized `commandCompletions` value (computed at render time, never stale) instead of the closed-over state variables.

## Additional Context

addyCooks:
Closes #1327

## Description

Pressing Enter immediately after typing a slash-command fragment submitted the raw fragment instead of selecting a completion.

**Cause.** The completion menu is opened by an effect that runs a commit after the keystroke that changed `input`. Ink re-subscribes the `useInput` handler in the same passive-effect flush, so for a turn or two the live handler sees the new `input` but `showCompletions` still `false`. Enter in that window skipped the "select the completion" branch and fell through to `handleSubmit`. Instrumenting the handler confirmed it: Enter one or two event-loop turns after the fragment rendered saw `input="/test-h"`, `showCompletions=false`, one memoized match, and submitted `/test-h`.

**Why not gate purely on `commandCompletions`** (the issue's suggestion): right after a completion is selected, the completed command still has matches, so a memo-only gate would re-select on every Enter and never submit.

**Fix** (`source/components/user-input.tsx`):
- The Enter gate keeps using the menu state when it says the menu is open (so a menu re-opened with Tab still works), and only when that state says closed falls back to the memoized completions the effect is about to show.
- The fallback applies the same suppression the effect uses closed for this exact input, or recalled from history now shared as `isMenuSuppressedFor` so the two cannot drift.
- Selecting a completion now marks that input closed, like Escape already did, so Enter on a selected command submits it.

## Recording

B06-FIXED-enter-selects-completion

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Changeset

### Automated Tests

### Manual Testing

- [ ] Tested with Ollama
- [ ] Tested with OpenRouter

## Checklist

- [x] Self-review completed
- [x] Documentation updated (if needed) no documented behaviour changes


will-lamerton:
Hey @addyCooks - looking good but could you resolve the conflicts? :)

addyCooks:
Done, 
I've conflicts resolved : )

PR #874, opened 2026-08-13
feat: add first-class OrcaRouter provider template

Marc-oss-hub:
## Description

Adds a first-class [OrcaRouter](https://www.orcarouter.ai) provider to the `/setup-providers` wizard and the provider factory, mirroring how the existing Requesty/OpenRouter providers are wired. OrcaRouter exposes 150+ namespaced models (`openai/*`, `anthropic/*`, `deepseek/*`, `z-ai/*`, …) behind a single OpenAI-compatible endpoint (`https://api.orcarouter.ai/v1`) and one API key. It also runs gateway-level, zero-trust security for AI agents on the same endpoint — screening every prompt/response and governing every tool call on a default-deny basis, with no application code changes.

I'm an engineer on the OrcaRouter team.

- Wizard: the new `OrcaRouter` template (base URL `https://api.orcarouter.ai/v1`, default model `openai/gpt-5.5`) sits next to `Requesty`/`OpenRouter` in `PROVIDER_TEMPLATES`.
- Provider factory: an `orcarouter`-named provider now gets the same `HTTP-Referer`/`X-Title` app-attribution headers as OpenRouter/Requesty, via a new `isOrcaRouterProvider` helper.
- Docs: a dedicated [OrcaRouter provider page](https://github.com/Nano-Collective/nanocoder/blob/main/docs/configuration/providers/orcarouter.md) plus a row in the Cloud Providers index.
- Tests: mirror the existing Requesty spec cases — wizard template config/build, and factory header wiring.
- Changeset: `minor` bump for the new provider template.

## Type of Change

- [x] New feature
- [ ] Bug fix
- [ ] Breaking change
- [ ] Documentation update

## Changeset

- [x] Added a changeset (`.changeset/add-orcarouter-provider.md`) describing this change for the changelog

### Automated Tests

- [x] New features include passing tests in `.spec.ts` files (`provider-factory.spec.ts`, `provider-templates.spec.ts`)
- [x] All targeted tests pass (`ava` on both spec files)
- [x] `pnpm test:types`, `pnpm test:lint`, `pnpm test:format`, `pnpm test:knip` all pass locally

### Manual Testing

- [x] Live API test through Nanocoder's own code path (`AISDKClient.create` → `createProvider` → `@ai-sdk/openai-compatible`): a real `openai/gpt-5.5` request to `https://api.orcarouter.ai/v1` returned `ORCA_OK` (usage 4,393 in / 18 out tokens).
- [ ] Tested with Ollama
[...]

PR #1364, opened 2026-09-17
feat: add first-class Cheaper Inference provider template

aiapienthusiast:
## Description

Adds Cheaper Inference (`https://api.cheaperinference.com/v1`) as a first-class provider template in the setup wizard. It also adds a documentation page.

Cheaper Inference is an OpenAI-compatible gateway. One endpoint and one API key reach models from several upstream providers. It speaks the OpenAI Chat Completions API, so it needs no new SDK path. It flows through the existing generic `openai-compatible` provider. This change is therefore a wizard template, a docs page and tests.

The change follows #874 (`feat: add first-class OrcaRouter provider template`), which added OrcaRouter in the same way.

### Files changed

| File | Why |
| --- | --- |
| `source/wizards/templates/provider-templates.ts` | New `apiKeyTemplate` entry: id `cheaper-inference`, base URL pre-filled, default model `claude-sonnet-5`. |
| `source/wizards/templates/provider-templates.spec.ts` | Two tests that mirror the Requesty and OrcaRouter pair: base URL, default model, comma-separated model parsing, default provider name. |
| `docs/configuration/providers/cheaper-inference.md` | New provider page, listed under Cloud Providers (OpenAI-Compatible). |
| `docs/configuration/providers/index.md` | Docs nav entry. |
| `.changeset/add-cheaper-inference-provider.md` | Changeset (minor), because the change is user-facing. |

### Not included

#874 also added `source/ai-sdk-client/providers/orcarouter.ts` and a `provider-factory.ts` branch that attaches `HTTP-Referer` and `X-Title` app-attribution headers.

Those headers are the OpenRouter convention. No documentation shows that Cheaper Inference reads them, so this PR does not assert that it does. If it does, the predicate file and the factory branch are a small follow-up.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

## Changeset

### Automated Tests

`pnpm test:all` did not run to completion. These checks ran on `b2ef8f01d507b88c6b67fb83a52505ef39bb87f5`:

- `pnpm test:types` — pass
- `pnpm test:types:vscode` — pass
- `pnpm test:lint` — pass, clean
- `pnpm test:knip` — pass, hints only
- `pnpm test:ava source/wizards/templates/provider-templates.spec.ts` — 44 passed, with the two new cases
- `pnpm test:format` — one pre-existing error in `source/tools/file-ops/write-file.tsx` (import ordering). It is present on the base commit. No file in this PR is flagged.
- `pnpm test:ava` (whole suite) — **did not finish**. The sandbox filesystem filled up and the run was killed. Every test it reported passed, and it printed no failure, but there is no exit code. This is therefore not a pass.
- `pnpm test:audit` and `pnpm test:security` — not run

The two new tests cover template construction, which has no error path. The third box stays unticked.

### Manual Testing

- [ ] Tested with Ollama
- [ ] Tested with OpenRouter

Nothing ran through nanocoder itself, so no box is ticked.

A direct check with a real key confirmed the endpoint. `GET /v1/models` returns the catalogue. A plain chat completion and an SSE stream both answer. The base URL and the key header in this template are the working ones.

A reviewer should confirm the path through nanocoder: the wizard writes the config it should, and a chat completion succeeds from the app.

## Checklist

- [x] Self-review completed

No logging was added. The change is a declarative template entry and docs, with no new runtime code path to log from.


aiapienthusiast:
Both nits looked at.

`sidebar_order` is now 26, which is unused - 19 was shared with `orcarouter.md`, `groq.md` and `together.md`, so the page slots in after the existing entries instead of tying with three others.

On the tests: agreed that they mostly exercise `apiKeyTemplate(...)`, which is uniform across its consumers. I left them mirroring the OrcaRouter pair on purpose - if this provider ever needs to stop using that helper, the pair is what fails first, and a test that asserts the helper's own output would duplicate the helper's tests. Happy to add field-level assertions (no `apiKeyPrompt`, no `sdkProvider`, default model `claude-sonnet-5`) if you would rather have them here than in the template suite.

akramcodez:
Thanks for the PR @aiapienthusiast

PR #1374, opened 2026-09-17
fix(tui): re-layout on every terminal resize, not just clamped ones

### LINKED ISSUE #1328: [Bug] Resizing the terminal sometimes doesn't re-layout the UI at all
## Description

`actualWidth` is a raw, non-reactive read of `process.stdout.columns`. The only state that actually triggers a re-render is a separately-clamped `boxWidth`. A resize that doesn't move `boxWidth` produces zero re-renders every layout consumer (welcome screen, input box, status bar, session selector) can silently freeze at the old layout after a resize, sized for a terminal that no longer exists.

Parent: #907

## Steps to Reproduce

1. Start Nanocoder at a given terminal size.
2. Resize the terminal to a different size.
3. Observe the layout.

## Expected Behavior

The UI re-measures and redraws for the new terminal size.

## Actual Behavior

The UI can keep rendering content sized for the *old* width. Since the real terminal is now a different size, lines drawn for the old width get cut off mid-word by the terminal itself, and boxes end up visibly misaligned.

## Recording

Image

## Root Cause

`source/hooks/useTerminalWidth.tsx:7-8,112` `calculateBoxWidth = Math.max(Math.min(columns - 4, 200), 40)` is the only reactive state; `actualWidth = process.stdout.columns` is a raw render-time read with no subscription of its own.

## Suggested Fix

Subscribe directly to the raw column/row count instead of deriving reactivity from the already-clamped `boxWidth`.

## Additional Context

addyCooks:
Closes #1328

## Description

`useResponsiveTerminal` returned `actualWidth` as a raw render-time read of `process.stdout.columns`, with no subscription of its own. The only reactive state was the clamped `boxWidth` (`max(min(columns - 4, 200), 40)`), so a resize that left `boxWidth` unchanged produced **no re-render at all**:

- below 44 columns, `boxWidth` pins to 40
- above 204 columns, it pins to 200

Resize anywhere inside those bands and every consumer welcome screen, input box, status bar, session selector kept rendering for the old terminal, so lines got cut off mid-word and boxes sat misaligned.

**Fix:** the raw column count becomes the reactive state, and `boxWidth` derives from it. `useTerminalWidth` and `useResponsiveTerminal` now both read the same subscribed value, and nothing reads `process.stdout.columns` during render. The single shared `resize` listener is unchanged, so there is still exactly one stdout listener regardless of how many consumers mount. `boxWidth` had been computed in two places; it is now computed once.

## Recording

B07-FIXED-live-resize

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Changeset

### Automated Tests

### Manual Testing

- [ ] Tested with Ollama
- [ ] Tested with OpenRouter

## Checklist

- [x] Self-review completed


addyCooks:
The failing `Unit Tests & Coverage Analysis` check here isn't from this PR.

`StatsDisplay changes range with arrow keys and closes on Escape` (`stats.spec.tsx:121`) wedges on a pre-existing stale-closure bug in `StatsDisplay`: 
the arrow handler steps the range from its captured closure value, 
but Ink re-registers `useInput` handlers in a passive effect that runs *after* the frame is painted. 

The test polls the frame as its readiness signal, so its second `→` lands in that window, recomputes the tab it's already on, 
and React bails out.

Nothing on this branch touches `/stats` or input handling, 
it reproduces on `main` too, so it can block any PR. 

Fix is up separately in #1379; 
once that merges I'll bring `main` into this branch for a green run here.

PR #1375, opened 2026-09-17
fix(tui): bound the reasoning stream to a tail and skip it when collapsed

### LINKED ISSUE #1329: [Bug] The reasoning/thinking trace re-processes its entire history on every flush, even while collapsed
## Description

`StreamingReasoning` re-wraps the FULL reasoning string on every token flush. `StreamingMessage` already has a bounded-tail helper (`computeStreamingTail`) for exactly this cost; `StreamingReasoning` never adopted it. Since `reasoningExpanded` defaults to `false`, this cost is paid on every flush for a panel that's collapsed and invisible by default.

Parent: #907

## Recording

Not included this is a pure performance/CPU-cost finding with no visual symptom a screen recording could show; the UI looks identical whether the render is fast or slow. Evidence below is from a render-timing harness instead.

## Evidence

30 flushes over ~150KB reasoning, expanded:   2890 ms (~96 ms/flush)
30 flushes over ~150KB reasoning, collapsed:  2356 ms (~79 ms/flush)  <- renders nothing, still costs this
StreamingMessage baseline, same size:         1187 ms (~40 ms/flush)
Full-wrap cost per call at 640KB: 214.10 ms   vs. a bounded 3.8KB tail: 2.16 ms

## Expected Behavior

Reasoning-stream cost should stay bounded to a fixed tail, the same as `StreamingMessage`, and should be skipped entirely when the panel is collapsed.

## Actual Behavior

Full-history re-wrap happens on every flush, expanded or not.

## Root Cause

`source/components/streaming-reasoning.tsx:35` calls `wrapWithTrimmedContinuations(reasoning.trimEnd(), textWidth)` on the complete accumulated string every flush, with no bound and no early-exit when `expand === false`.

## Suggested Fix

Reuse `computeStreamingTail` in `streaming-reasoning.tsx`, and skip the wrap (and the `calculateTokens` call) entirely when `expand === false`.

## Additional Context

- [x] I can reproduce this issue consistently (measured via a render-timing harness against the real component)

addyCooks:
Closes #1329

## Description

`StreamingReasoning` called `wrapWithTrimmedContinuations(reasoning.trimEnd(), textWidth)` on the **entire** accumulated reasoning string on every token flush, with no bound and no early exit. `StreamingMessage` already solved this with `computeStreamingTail`; the reasoning component never adopted it. Worse, `reasoningExpanded` defaults to `false`, so the full re-wrap plus `calculateTokens` ran on every flush for a panel that renders nothing.

**Fix:**
- Reuse `computeStreamingTail` (already exported from `streaming-message.tsx`) so each flush is O(tail) rather than O(history).
- Skip the wrap and the token count entirely when `expand === false`.

No new helper and no new file the bounded-tail logic stays in one place.

**Measured**, 30 flushes over a ~150KB trace (same shape as the harness in the issue):

| | before | after |
|---|---|---|
| collapsed | 2757ms total 91.9ms/flush | 348ms 11.6ms/flush |
| expanded | 2977ms total 99.2ms/flush | 757ms 25.2ms/flush |

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Changeset

### Automated Tests

### Manual Testing

- [ ] Tested with Ollama
- [ ] Tested with OpenRouter

## Checklist

- [x] Self-review completed


will-lamerton:
Great work @addyCooks :)

addyCooks:
Thanks @will-lamerton : )

PR #1392, opened 2026-09-18
fix(tui): window the checkpoint list like the session list

### LINKED ISSUE #1372: [Bug] Checkpoint list shows all items with no scrolling, unlike every other selector
## Description

`CheckpointSelector` renders every checkpoint in an unbounded list. `SessionSelector`, one file over, already caps at `limit = Math.min(items.length, 10)` the same windowed-list pattern was simply never applied to checkpoints.

Parent: #907

## Steps to Reproduce

1. Accumulate more checkpoints than fit on one screen (a project with regular `/checkpoint create` usage gets there quickly).
2. Run `/checkpoint load`.

## Expected Behavior

A scrollable, windowed list matching how session history is already handled.

## Actual Behavior

All checkpoints render in one continuous unbounded list, extending well past the visible terminal height with no scroll indicator.

## Recording

Image

## Root Cause

`source/components/checkpoint-selector.tsx` has no equivalent of `SessionSelector`'s existing `limit` logic.

## Suggested Fix

Apply `SessionSelector`'s existing windowed-list pattern to `CheckpointSelector`.

## Additional Context

addyCooks:
Closes #1372

## Description

`CheckpointSelector` rendered every checkpoint through `StyledSelectInput` with no bound, so a project with regular `/checkpoint create` use pushed `/checkpoint load` well past the visible terminal, with no scroll affordance.

It now uses `FilterableSelectList` with `visibleCount={10}` the same component `SessionSelector` uses so the window scrolls with the highlight, typing filters the list, and the component's own height-aware clamp shrinks the window further on short terminals.

**Note on the issue text:** it describes `SessionSelector` capping via `limit = Math.min(items.length, 10)`. That is stale  the session list moved to the shared `FilterableSelectList`, so "the windowed-list pattern SessionSelector already uses" is that component today, and that is what this PR applies.

Two details beyond the swap:
- The component's own `useInput` is now scoped with `{isActive: awaitingBackupConfirmation || checkpoints.length === 0}`. Ink broadcasts `useInput`, so leaving it active while the list owns Escape would fire `onCancel` twice. `SessionSelector` guards its empty state the same way and for the same reason.
- Items carry `searchText: checkpoint.name` so the filter matches on the name rather than the counts and age suffix, mirroring the session list filtering on its title.

The local `CheckpointOption` type is gone in favour of the shared `ItemSelectorOption`.

## Recording 

B11-FIXED-checkpoint-windowed

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Changeset

### Automated Tests

### Manual Testing

- [ ] Tested with Ollama
- [ ] Tested with OpenRouter

## Checklist

- [x] Self-review completed


will-lamerton:
Nice work @addyCooks :)

addyCooks:
Thanks @will-lamerton : )

PR #1393, opened 2026-09-18
fix(tools): match find_files patterns written with a leading "./"

### LINKED ISSUE #1343: [Bug] find_files fails to match any files when pattern begins with ./
## Description

When models or users query `find_files` using standard relative path notation starting with `./` (such as `./src/**/*.ts`, `./package.json`, or `./source/*.tsx`), `find_files` returns `No files or directories found matching pattern`. This occurs because `matchesGlob` receives normalized relative file paths that do not start with `./` (e.g. `source/cli.tsx`), but `tokenizeExpandedPattern` treats `./` as literal tokens (`.` and `/`). The tokens never match the target path.

## Environment

- **Node version**: v22.14.0
- **Provider**: Any
- **Model**: Any

## Steps to Reproduce

1. Open nanocoder in any repository.
2. Have the agent or tool execute `find_files` with `pattern: "./package.json"` or `pattern: "./source/**/*.ts"`.
3. Observe the tool output.

## Expected Behavior

`find_files` should match `package.json` or source files when prefixed with `./`.

## Actual Behavior

Returns `No files or directories found matching pattern "./package.json"`.

## Logs/Screenshots

In `source/utils/file-search.ts`:
`normalizePathForMatch(pattern)` only does `pattern.replace(/\\/g, '/')`, preserving leading `./`. In `matchesGlob`:
```typescript
export function matchesGlob(filePath: string, pattern: string, matchBasename = false): boolean {
    // Leading './' is not stripped from pattern, so target "package.json" cannot match "./package.json"
}
```

## Additional Context

macjayz:
## Description

`find_files` returns no results for patterns written with a leading `./`, such as `./src/**/*.ts` or `./package.json`. Project entries are relative paths with no `./` prefix, so the pattern is compared against `src/index.ts` and never matches — there is no error, just an empty result, which reads to the model as "these files do not exist".

This strips a leading `./` from the pattern before tokenizing it, and computes the basename fallback from the stripped pattern so `./*.tsx` behaves exactly like `*.tsx`. A bare `./` is deliberately left alone: it would strip to an empty pattern, which tokenizes to zero tokens and trips the token cache's size calculation. `../` is untouched.

Closes #1343

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Changeset

### Automated Tests

Two tests in `source/utils/file-search.spec.ts`: one over `matchesGlob` covering `./` patterns, Windows `.\` separators, the untouched `../` case, and the bare `./` guard; one over `findMatchingPaths` against a real temp directory, asserting `./package.json` returns the same set as `package.json`. Both fail without the fix.

### Manual Testing

- [ ] Tested with Ollama
- [ ] Tested with OpenRouter

Not applicable — this is provider-independent path matching, covered by the tests above.

## Checklist

- [x] Self-review completed

## Note for reviewers

`search_file_contents` has the same bug in its `include` parameter, which is passed straight to ripgrep's `-g` flag (`rg -g './src/**/*.ts'` matches nothing while `rg -g 'src/**/*.ts'` does). I left it out to keep this PR scoped to #1343 — happy to take it in a follow-up if you would like it fixed.


will-lamerton:
Nice work @macjayz :)

PR #1396, opened 2026-09-19
fix(tui): count the status bar's indent in its width budget

### LINKED ISSUE #1380: [Bug] The status bar truncates mid-word at exactly 80 columns
## Description

`development-mode-indicator.tsx` runs a careful segment-drop-order budget against `actualWidth` but it's rendered inside a wrapper with `marginLeft={3}` in `user-input.tsx`, and the budget math never subtracts that margin. At exactly 80 columns, the status line silently overflows by those 3 unaccounted columns, cutting text off mid-word instead of cleanly dropping an optional segment.

Parent: #907

## Steps to Reproduce

1. Use a terminal that's exactly 80 columns wide (the most common default).
2. Cycle to auto-accept mode (Shift+Tab) and give the session a moderately long name (`/rename <long name>`).
3. Look at the status bar.

## Expected Behavior

Either the full text fits, or an optional segment is cleanly dropped per the component's own documented priority order.

## Actual Behavior

Text truncates mid-word: `auto-accept mode on` renders as `auto-accept mode o`, and a long session name loses its last character too.

## Recording

Image

## Root Cause

`source/components/development-mode-indicator.tsx`'s budget math computes against `actualWidth` directly; `source/components/user-input.tsx` renders it inside `<Box marginLeft={3}>`, a 3-column offset the budget never accounts for.

## Suggested Fix

Subtract the wrapper's `marginLeft` from the budget (`actualWidth - 3`), or pass the already-reduced width in as a prop.

## Additional Context

addyCooks:
Closes #1380

## Description

`development-mode-indicator.tsx` runs a careful segment-drop-order budget so the status row degrades gracefully: drop the Ctrl-t hint, then the saving indicator, then the editor suffix, then the shift hint, then ellipsise the session name. All of that measured against `actualWidth` the full terminal width while `user-input.tsx` renders the row inside `<Box marginLeft={3}>`. The budget was short by exactly those 3 columns, so the row overflowed and the terminal cut it mid-word instead of the component dropping a segment.

Measured at 80 columns, auto-accept mode, a 45-character session name:

| | rendered row |
|---|---|
| before | **82 columns** overflows, cut mid-word |
| after | 79 columns, session name ellipsised cleanly |

**Fix:** the indent comes in as an `indentColumns` prop and is subtracted once (`budgetWidth`), which the six budget sites then use.

**Why a prop rather than hardcoding `actualWidth - 3`:** the indicator has two render sites in `user-input.tsx`. The wrapped one passes `indentColumns={3}`; the disabled/busy state has no wrapper, so it keeps the full width via the prop's `0` default. Subtracting 3 unconditionally inside the component would have under-budgeted that second site by 3 columns.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Changeset

### Automated Tests

### Manual Testing

- [ ] Tested with Ollama
- [ ] Tested with OpenRouter

## Checklist

- [x] Self-review completed


will-lamerton:
Awesome work @addyCooks :)

addyCooks:
Thanks @will-lamerton !!

